PR #9, opened 2015-07-16
AffineBuilder simplifies definitions of common NNs

davidweichiang:
Realized I never made a pull request about this. This introduces a class, AffineBuilder, analogous to the RNN builders, that encapsulates the parameters of an affine transformation. This simplifies definitions quite a bit. See xor.cc, nlm.cc, and lstm.cc for examples.

Another option would be to call it LayerBuilder and add in an activation layer as well. That makes it easier to set the initialization range of the weight matrix properly.


neubig:
Sorry this sat around for so long. This is a nice pull request, and we really should get something like this in the most recent version of DyNet. I'll make an issue.

PR #11, opened 2015-08-25
Tooriginclab

kaishengyao:
My changes to cnn. 
1) added depth-gated lstm code
2) added code to free allocated memory, so that cnn won't have memory leaks.
3) declare long instead of unsigned in some places when calling model initialization. visual studio detects these and complained.
4) use 8192 instead of 512 in concatenate_cols. That number is too small for a case that I found. 
5) initialization with fixed seed. Austin did similar thing but I am using a --seed argument.


neubig:
Hello, unfortunately this pull request sat around for a while and now the master branch has advanced quite a bit. I'm going to close this for now as it probably wouldn't be realistic submitting it, but I think things like depth-gated LSTM would be nice to have, so if you're still interested it'd be great if you could adapt them to the more recent version and commit them, and make pull requests with smaller granularity.

PR #29, opened 2016-01-01
skip nodes that do not participate in the backward computation.

yoavg:
Hi,
This seem to work but can you please verify I didn't do something silly?


neubig:
Looks good to me!

PR #31, opened 2016-02-12
Score cdec/Moses N-best lists with rnnlm

mjdenkowski:
I wrote this for N-best re-scoring experiments.  Nearly everything is wrapped in an "if" block so it shouldn't affect other code.


mjdenkowski:
Looks like Travis doesn't have the boost regex library, which I was using to parse various flavors of N-best lists.  Are you guys trying to minimize number of boost components used?

PR #34, opened 2016-02-14
pycnn: error message on the wrong assertion

dhgarrette:
@yoavg


yoavg:
yup. thanks for noticing.

PR #35, opened 2016-03-17
Updating installation instructions for PyCNN

alvations:
I have added more generic installation instructions, I've remove the multiple nested directories to get to the src, previously it was `mkdir cnn; git clone ...cnn.git` and it creates nested `cnn/cnn/cnn` path to get to the src directory, now it's `cnn/cnn`. 

Possibly, the source directory should be renamed to `cnn/src` in the future but it would require proper refactoring which this PR doesn't cover.

Hope this will help improve the installation process of `PyCNN` and more people can start using this nice tool!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,6 @@
 # Installing the pyCNN module.
 
+
 First, get CNN:
 
 ```bash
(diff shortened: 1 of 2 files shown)


alvations:
The failure on Travis CI is because the most recent `eigen` source fails to compile.

yoavg:
Thanks!
I prefer to not have it rely on my copy of eigen.tgz though, can you make the main eigen version be the freshest version from bitbucket, and my eigen.tgz only mentioned as a commented-out fallback?
(alternatively, if you can figure out how to check out an older, working version of eigen using hg, this would also be good)

alvations:
This seems to be the last stable version of `eigen`: https://bitbucket.org/eigen/eigen/commits/47fa289dda2dc13e0eea70adfc8671e93627d466

So i've changed the installation instructions to use `-r` with `hg clone`

PR #38, opened 2016-04-22
Add missing header files, so 'include' directory has all needed files.

aria42:
Might be missing something, but if you `make install` this locally to use in another project, the install headers are missing these two files which prevent you from including `<cnn/cnn.h>`


neubig:
Thanks for the fix!

PR #40, opened 2016-05-11
Added `inputVector(vec)` to PyCNN (so a separate `set` is not required).

dhgarrette:
Previously, to input a literal vector, it was necessary to call `v = vecInput(dim)` followed by a seprate call to `v.set(vec)`.  Now this can be accomplished more smoothly by calling `v = inputVector(vec)` to create an expression represeting vector `vec`.


dhgarrette:
@yoavg

PR #41, opened 2016-05-13
Added error messages when empty lists are passed to PyCNN functions.

dhgarrette:
Errors were previously more cryptic.

@yoavg


yoavg:
Did you verify the speed implications of this?
How much time does it take to create, say, 10,000 esum operations with vs without the assertion?

dhgarrette:
Yeah, I modified an xor example to perform an explicit concatenate.  With 10 million training iterations, the run time on my laptop went from 461 seconds to 464.  

That said, I'm totally fine with rejecting the change since it's definitely not necessary and only marginally useful.

PR #48, opened 2016-06-05
Add libboost-regex-dev, without which Travis tests fail

danielhers:


danielhers:
Apparently they fail with it too, but due to C++ errors.

neubig:
Thanks for the fix!

PR #54, opened 2016-06-15
Added `inputMatrix(v, (d1,d2))` to PyCNN (so separate `set` not req'd)

dhgarrette:
@yoavg


dhgarrette:
Makes it possible to do something like:

    y = inputMatrix([1,2,3,4,5,6], (3,2))

Instead of the old way, which was:

    y = matInput(3,2)
    y.set([1,2,3,4,5,6])

PR #56, opened 2016-06-18
Commented out PyCNN RNNBuilder.__cinit__ b/c it's called by subtypes.

dhgarrette:
@yoavg 

RNNBuilder's constructor was previously declared as `__cinit_`, so it wasn't recognized by the compiler, meaning that RNNBuilder didn't have a constructor, which caused opaque crashes when code tried to instantiate an RNNBuilder directly.  Changing it to `__cinit__` corrected this problem by making it so that if code tried to instantiate an RNNBuilder, it would get a clear error message.  However, it appears that RNNBuilder's subtype constructors always call RNNBuilder's constructor (even without an explicit `super` call), which means that the "cannot instantiate directly" error message gets triggered when constructing any RNNBuilder subtype.

As a result, I'm commenting out `__cinit__` so that the subtypes will work.  This means that RNNBuilder still won't give a good error message if the user attempt to instantiate it directly, but maybe there's another way around that...

### DIFF
--- a/pycnn/pycnn.pyx
+++ b/pycnn/pycnn.pyx
@@ -741,16 +741,10 @@ cpdef Expression affine_transform(list exprs):
 # {{{ RNNS / Builders
 # TODO: unify these with inheritance
 
-cdef class RNNBuilder: # {{{
+cdef class _RNNBuilder: # {{{
     cdef CRNNBuilder *thisptr
     cdef RNNState _init_state
     cdef int cg_version 
-    #def __cinit__(self, unsigned layers, unsigned input_dim, unsigned hidden_dim, Model model):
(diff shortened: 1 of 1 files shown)


yoavg:
oh. so the typo was a fortunate accident after all.
ok, can you maybe change this pull request such that RNNBuilder is renamed to _RNNBuilder or something like that, and verify that it works?

dhgarrette:
Good idea.  I've now changed RNNBuilder to _RNNBuilder and verified that if you instantiate eg LSTMBuilder it works but if you try to instantiate RNNBuilder you get "NameError: name 'RNNBuilder' is not defined".

PR #62, opened 2016-06-23
Initial version of PyCNN network visualizer

dhgarrette:
Import pycnn_viz instead of pycnn and call PrintGraphviz().

Key features:
* See parameter names
* Optionally show the dimensions of each node
* Distinguish explicitly-provided dimensions ({{}}) from inferred
dimension labels ({})
* Dimensions are computed and checked, and any errors are highlighted in
the graph
* Verbose and compact modes
* Optionally see names of lookups instead of indices (like seeing the
actual word of a word embedding)
* Collapse RNN states into single boxes
* Optionally detect and collapse BiRNNs
* Optionally assign labels to individual expressions

Some example network graphs displaying a variety of features can be seen here:
pycnn_viz_example1
screen shot 2016-06-24 at 4 16 20 pm

@yoavg


yoavg:
Looking at the second graph, shouldn't the dim from pickneglogsoftmax be 1?

dhgarrette:
@yoavg Yep, good catch. Fixed.

yoavg:
before I do the merge, can can you try adding the following as the last lines of pycnn.pyx, and see if it works:
`
if '--pycnn-viz-mode' in sys.argv:
    print >> sys.stderr, "PyCNN loaded in visualization mode."
    from pycnn_viz import *
`
?

the effect should be that adding the commandline switch --pycnn-viz-mode will import your expressions instead of the pycnn ones -- it works for non-cython modules, but I don't know if it holds also for cython ones. If it works, I think it is worth including.

dhgarrette:
If you add those lines, it compiles fine, but when you import pycnn it gives the following, presumably since the classes with those names are already loaded:

    PyCNN loaded in visualization mode.
      File "bilstmtagger3.py", line 1, in <module>
        from pycnn import *
      File "build/bdist.macosx-10.11-intel/egg/pycnn.py", line 7, in <module>
      File "build/bdist.macosx-10.11-intel/egg/pycnn.py", line 6, in __bootstrap__
      File "pycnn.pyx", line 1168, in init pycnn (pycnn.cpp:38137)
    TypeError: Cannot overwrite C type AdamTrainer

I also tried putting the full contents of pycnn.pyx under the `else` side of the statement, but apparently that's not valid:

    Error compiling Cython file:
    ------------------------------------------------------------
    ...

      from pycnn cimport *
      cimport pycnn


      cdef init(random_seed=None):
          ^

Thoughts? 

Also, maybe it should just be called `--vis-mode` since the "pycnn" part is implied?

dhgarrette:
(Actually, `--pycnn-viz-mode` makes more sense and is far less likely to conflict with any option defined by the user.)

yoavg:
hmm. annoying. oh well.

I will have some time to play with this tomorrow or on Monday if you rather wait and not do it yourself.
But I believe the following should work:

Rename the pycnn.pxd, pycnn.pyx files to cpycnn.{pxd,pyx} (and update the makefile and setup.py accordingly).
Add a pycnn.py file with:
```
import sys
if '--pycnn-viz-mode' in sys.argv:
   from pycnn_viz import *
elif '--pycnn-gpu' in sys.argv:
   from gpycnn import *
else:
   from cpycnn import *
```

PR #64, opened 2016-07-07
Support for memory larger than 2000MB

su-chao:
If we need memory larger than 2000MB (mb=2000), the code "mb << 20" in devices.cc will get a wrong number and lead to a failed allocation, due to the limited range of "int" type. When we change the type of mb from "int" to "size_t", the problem is solved.


neubig:
Thanks for the pull request! This is actually already fixed in the v2 branch of cnn, which will be merged into the master shortly: https://github.com/clab/cnn/issues/52

PR #66, opened 2016-07-18
V2

duyvuleo:
Hi,

May I request for a simple example for implementing mini-batch SGD training?


neubig:
Sure, this already exists here:

https://github.com/clab/cnn/blob/v2/examples/xor-batch.cc
https://github.com/clab/cnn/blob/v2/examples/rnnlm-batch.cc

PR #69, opened 2016-08-12
Py3 compatibility for pycnn

hayounav:
Make pycnn compile and run under python 3 (hopefully without breaking
python 2 compatibility)


neubig:
Sorry it took so long to get this integrated. Python 3 support is something we should have, but there have been rather big changes since this pull request was made. If it'd be possible to get an updated pull request on the current master branch that'd be great.

PR #87, opened 2016-09-22
Changes to allow compiling in Windows

mattr1:
With the included changes, CNN compiles in Windows, on Microsoft Visual Studio 2015. I tried to make minimal changes to enable this. A few comments:
  1) I removed pthread from the list of link libraries. I didn't see it being used anywhere. Did I miss something?
  2) I marked any windows-specific portions with "#if _WINDOWS". CMake automatically adds this compilation symbol to the makefile on windows.
  3) I did not port the multi-process code, so mp.cc/h are not compiled, and the rnnlm-mp example does not compile properly (the user needs to uncheck this project from the list of projects to compile in MSVC). 
  4) The SharedMemory allocator is not currently supported.
  5) I had to add a getpid.h included in all the examples that call getpid(). They are only using getpid() to generate the parameters filename. If the purpose was just to generate a random number, we could do that instead and get rid of getpid.h, or leave it as-is if you find it useful to have the pid in the filename.
  6) I made a single change in test-cnn to remove a warning when compiling in 64-bit mode. Do you tend to compile as 64-bit, or is the library intended to be only 32-bit?

I believe points #3 and #4 could be fixed later by depending exclusively on the boost.interprocess library. I noted this in the "TODO", and also added instructions for compiling in Windows to the README.


mattr1:
One final note: I contacted Chris to ask if he's interested/willing to accept changes keeping CNN compatible with Windows, so he is expecting this pull request.

neubig:
First, this is excellent! Thanks a bunch for ensuring compatibility.

Just one thing: what about adding getpid.h to the main cnn/ directory? If this is going to need to be called by many programs that use cnn, it should be part of the main directory itself. Also, it wasn't clear to me exactly what dependency this file was resolving. If it's including a header for something that cnn depends on, I think it should be included in the appropriate files within the cnn/ directory, so the files in the example/ directory don't have to call it directly.

mattr1:
I'm happy to move getpid.h to the cnn/ directory, but it is not actually used by anything in the cnn directory. Instead, it is only used by programs in the examples directory (and rnnlm). It is used to support calling getpid(), which is only done in the example programs. In fact, the examples only use getpid() to come up with a unique filename for saving model parameters. I felt that this use of getpid() was particular to the way the examples are written, so probably most people using CNN won't actually need getpid.h.

neubig:
OK, if that's the case this looks like the right thing to do. Thanks a lot for doing this!

PR #102, opened 2016-10-14
Added MKL, and got CUDA working for Windows

mattr1:
1) Added MKL support. I've only tested this in Windows. I get 3x speedup on seq2seq. I tried to make a general find_mkl function that should also work fine in Linux. 
2) Got CUDA working for Windows. There wee a variety of compile errors that happen with the full Eigen package when nvcc is compiling, which were fixed by not including all of Eigen when nvcc is compiling. A compile problem in Boost required setting BOOST_NO_CXX11_ALLOCATOR when nvcc is compiling. I don't know if there is a performance hit for this.
3) I've changed the library generation in Windows back to static rather than shared. A shared DLL in Windows by default does not export any symbols, and there was no easy way to do so. The most straightforward would require prefixing every function in the source code with a macro that would expand to "__declspec(dllimport)" on Windows and nothing on Linux, but that seems like a big pain for what seems like little gain (making a DLL instead of a static library). 

These changes are compiled against the latest version of Eigen, and Boost 1.61. CUDA for Windows currently requires a few changes to Eigen, which I've delineated in the install.md.


neubig:
This is really useful! Thanks a lot for committing this.

neubig:
@mattr1: Upon further testing, it seems that the changes to nodes.cc were breaking the "hinge" function on Linux (but not on Mac). I've reverted these changes here: https://github.com/clab/dynet/commit/180f87f87e5b8df1756077b921022b72cee41712 and tests seem to be passing again. I'm sure there's a way to reconcile and get it working on both OSs, but for the time being we need to prioritize Linux, which is the one that's officially supported.

mattr1:
@neubig. Okay, sorry about that. Thanks for fixing it. I think the differences are all just how nvcc interacts with the various host compilers.

PR #103, opened 2016-10-14
First shot at documentation using doxygen + breathe + sphinx

pmichel31415:
Just set up Doxygen and sphinx for a stub documentation.

I included Graham's doc, reformatted in .rst

### Remains to be done : 

* Document training.h for optimizer
* Document expr.h for operations
* Document builders and add to doxygen project
* Document minibatch and multiprocessing and add to doxygen project
* Create a readthedocs account for clab (needs to be an admin) and link the project (see http://docs.readthedocs.io/en/latest/getting_started.html#import-your-docs)

### Requirements : 

* Doxygen (config file is in doc/doxygen)
* sphinx (config file in doc/source)
* breathe for doxygen -> sphinx


neubig:
Thanks! This is great, but I think we should probably avoid checking in so many auto-generated files. Could you initiate a new branch with only the files that you actually wrote yourself, and we'll find a separate place to store the auto-generated files?

pmichel31415:
Is that better now? The tests are still failing though.

pmichel31415:
This should now be up to date with master

PR #106, opened 2016-10-18
Trivial correction.

memeda:
1. Compiling Error at MSVC 14.0 (MSVS 2015 update3) for Error  C2397.
    Because a inexplicitly narrow cast from unsigned to int. G++ is ok while MSVS report error.
    I use `static_cast<ptfdiff_t>` to explicitly cast, but I don't know whether this is the best solution. (can the `DSizes<ptrdiff_t, 3>` to be `DSizes<unsigned, 3>` ?)
2. Using MKL using -DMKL=TRUE for CMake comandline or error reporting.


memeda:
Wow, It's so nice~~

PR #107, opened 2016-10-18
Fixing Hinge::backward to work with CUDA for both Linux and Windows 

mattr1:
The single-line version of the expression breaks when compiling on MSVC with CUDA (something about the interaction between nvcc and MSVC). Breaking it into these multiple lines works. I've tested on both Windows and Linux, with CUDA and non-CUDA, and they all work (running test-dynet). I believe they should be the same speed, but if there are any speed concerns, I can wrap the changes with a "#ifdef _WIN32". I was hoping to minimize the number of places that have windows-specific code.


neubig:
Thanks! I haven't had a chance to profile this, but I think we can prioritize maintaining windows support and code conciseness and use this as-is.

PR #109, opened 2016-10-19
added more param inits, init expressions for cfsm

dbc148:
I added initfromvector and file for parameter initializations, as well as implementing add_parameter(dim, parameterinit) as that appeared to be unimplemented.  I also added a initialize_expressions to the cfsm_builder, as you need to pre-initialize all expressions when using it with model rollback.

### DIFF
--- a/dynet/cfsm-builder.cc
+++ b/dynet/cfsm-builder.cc
@@ -233,6 +233,15 @@ void ClassFactoredSoftmaxBuilder::serialize(Archive& ar, const unsigned int) {
   ar & p_rc2ws;
   ar & p_rcwbiases;
 }
+
+
+void ClassFactoredSoftmaxBuilder::initialize_expressions() {
+  for (unsigned c = 0; c < p_rc2ws.size(); ++c) {
+    Expression temp = get_rc2w(c);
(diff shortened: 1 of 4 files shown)


dynet/cfsm-builder.cc
```diff
@@ -233,6 +233,15 @@ void ClassFactoredSoftmaxBuilder::serialize(Archive& ar, const unsigned int) {
   ar & p_rc2ws;
   ar & p_rcwbiases;
 }
+
+
+void ClassFactoredSoftmaxBuilder::initialize_expressions() {
```
neubig: Hmm, there's a bit of a disconnect here between "initialize_expressions" and the for loop, which simply creates and destroys temporary variables. Is there a cleaner and more intuitive way to do this? At the very least this should be commented.

dbc148: I could assume no parameters have been initialized yet, loop through them, and initialize all of them?  If I don't have that assumption, I basically am reimplementing get_rc2w and get_rc2wbias, or overwriting previously initialized expressions (which shouldn't be problematic?  not sure, though).

neubig:
Great, thanks!

Issue #108, opened 2016-10-19
DyNet init function is not easy to use


neubig:
Currently, the DyNet "init" function takes command line arguments as-is, and is not straightforward to access programmatically. It would be better to have a structure containing the options that  can be passed to the initializer (while maintaining the compatibility with the current interface).

odashi:
I saw that currently there are 2 codeblocks to parse commandline args (one is in `initialize()`, and another is in `initialize_gpu()` which is called from the former).
They attempt to consume same args like `--dynet-mem`, and this behavior looks kind of confusing.
I think it might be better to integrate both to single codeblock as well as separating commandline parsing and actual configurations.

odashi:
Oh thanks, new code look better!

neubig:
Resolved by https://github.com/clab/dynet/pull/117

Issue #86, opened 2016-09-21
--dynet options having to be at the beginning of the command line is confusing


neubig:
When running programs compiled with --dynet, the command line options having to be at the beginning of the command line options is confusing. It'd be better if we didn't have this restriction.

neubig:
Resolved by https://github.com/clab/dynet/pull/117

neubig:
We've had a report that this is apparently still not fixed in the Python bindings.

neubig:
This was an error with some versions of upstream python: https://bugs.python.org/issue23058

Upgrading to Python 2.7.10 will probably fix the problem.

PR #117, opened 2016-10-21
New init

pmichel31415:
This pull request addresses issues #86 and #108 by : 

- Providing a "DynetParams" `struct` holding all general parameters
- Adding functions to initialize given DynetParams (accessible programmatically).
- Keeping legacy interface
- Allowing for dynet params to be put anywhere in the command line

**HOWEVER** I couldn't compile it for CUDA on my computer so someone with a gpu and CUDA installed should probably check this before merging.

### DIFF
--- a/dynet/cuda.cc
+++ b/dynet/cuda.cc
@@ -4,6 +4,7 @@
 
 #include "dynet/dynet.h"
 #include "dynet/cuda.h"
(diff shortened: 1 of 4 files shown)


neubig:
I confirmed this works on CUDA, awesome work!

PR #128, opened 2016-10-26
Python works in Windows, including MKL and/or GPU support

mattr1:
Windows Python build works correctly with MKL, GPU, none, or both. Updated documentation also. MKL and GPU support complicated setup.py.in somewhat, but I tried to keep it as clean as possible. 

Tested on Windows and Ubuntu Linux, with normal, MKL, GPU, and MKL+GPU. rnnlm.py works in all 8 cases.


neubig:
Thanks a lot!

mattr1:
No problem. Thanks for integrating the changes quickly.

Matt

PR #149, opened 2016-11-07
Adding installation notes for OS X

jkkummerfeld:
Minor change to provide additional information in the documentation for OS X users.


neubig:
Thanks! Perhaps we should also accommodate MacPorts users as well?

jkkummerfeld:
I don't have MacPorts set up on my current machine (so I wasn't able to test this), but I checked online for the relevant package names and updated as appropriate.

neubig:
Great, thanks!

Issue #179, opened 2016-11-21
API for clipping gradients by L2 regularization ?


zhang-wen:
i want to normalize the L2-norm of the gradient of the cost function each time to be at most a predefined threshold of 1, when the norm was larger than the threshold

is there clip_gradients (L2 regularization) API for python ?

pmichel31415:
sgd = new SimpleSGDTrainer(&model);
    sgd->clip_threshold = 1;

Is that what you're looking for?

zhang-wen:
yes, exactly, but no API in python.
is it easy to implement the API in python ?

pmichel31415:
Shouldn't be too hard, I'll look into it

pmichel31415:
Now that I think of it there is a flag to enable/disable clipping altogether in c++, it would eb a good idea to access it from python.

One way of doing this would be to disable clipping whenever a negative value is passed to `set_clip_threshold` instead of creating new accessors/setters.

What would be better @yoavg ?

yoavg:
does it make sense for set_clipping_threshold(0) to disable clipping?

pmichel31415:
I think it makes sense if this is specified in the docs.

zhang-wen:
it works, thank you !

PR #181, opened 2016-11-21
Added support for gradient clipping disabling to python

pmichel31415:
New behaviour : in set_clip_threshold(thr) if thr<=0 then set thr to 0 and clipping_enabled to false.


pmichel31415:
cf #179

Issue #155, opened 2016-11-08
cosine similarity


bsharpataz:
first of all -- I'm very new to issue writing, so.... here goes...

I was trying to implement a fairly simple architecture in dynet and got stuck on doing a cosine similarity (I'll essentially be creating two composite vectors from embeddings and comparing them with a hinge ranking loss in the final architecture). I see the implementation of dot_product, and esum, etc, but without the L2 normalization, I can't use the dot_product as a cosine sim. 

I do see squared_norm() in expr.cc on github, but can't seem to use it/find its implementation (I can use other functions which are also in the same file, like squared_dist()) to make sure it's what I'm looking for.  And, if it is what I'm looking for, can you please point me to an example of how to use it?

thanks!

neubig:
Do you mean that just calling
    dot_product(x, y) / squared_norm(x) / squared_norm(y)
doesn't work?

Of course, we should have a node to calculate cosine similarity, so I'll make this an enhancement.

rasoolims:
It seems that the python version does not have the squared_norm and also the sqrt functions.

rasoolims:
I tried your suggestion ```dot_product(x, y) / squared_norm(x) / squared_norm(y)``` but it seems that dividing by expression does not work in Python 
https://github.com/clab/dynet/blob/45d60a47990d763d0945293f4c1833b71b7da8a7/python/_dynet.pyx#L1862

neubig:
This is correct that the `/` operator should be supported, but you can use `dy.cdiv()` to achieve the same effect. I'll close this issue and open a new one.

PR #183, opened 2016-11-21
Added a bunch of missing functions

pmichel31415:
I haven't implemented the dynet_viz version of squared_norm because I'm not too sure of what the last argument of GVExpr should be. If someone knows better please tell me.

cf #155


yoavg:
re the viz interface, @dhgarrette is the one to ask.

there are merge conflicts now, can you fix?

neubig:
I'll go ahead and merge this to prevent later merge conflicts, but let's remember that we're still missing the dynet_viz version of squared norm.

PR #184, opened 2016-11-22
Protobuffer alternative for serializing and deserializing models.

kentonl:
Protobuffer alternative for serializing and deserializing models.


neubig:
Thanks @kentonl, this is great! But looking at the "changed files" in the pull request it seems that there are a lot of files that include recent changes to master that aren't relevant to this particular change. Maybe this is because you're sending a pull request to clab/dynet:protobuf instead of clab/dynet:master? Could you try to re-do the request so that only the minimal changes show up and we can do a proper code review?

Also, it looks like compile on Travis is failing now. We'll need to make sure it passes before we integrate the change.

yoavg:
this has great potential!

@kentonl the way to resolve this messy commit is probably to start a new branch from the latest master, switch to that branch, and then 'git cherry-pick 2c15499' (if this is indeed your only relevant commit. otherwise you can also add more cherry-pick lines). Then push that branch.

but it would also be nice to get a few more words of description on what you actually did and what this enables. 
- does this completely deprecate our dependence on boost serialization?
- can we still save "complex" objects like LSTMBuilder? 
- does this allow loading models without using `add_parameters` in the same order that was used before saving?
- does the python save/load functionality still work (given that we remove the boost methods)?

kentonl:
Thanks for looking at this @neubig! I sent the pull request for the protobuf branch because this is just an experimental proof of concept requested by @ikekonglp. The code style/organization is inconsistent with the rest of DyNet, and there are some undocumented dependencies (apt-get libprotobuf or brew install protobuf).

To get a clean list of changes, could you update the protobuf branch to e56018dfe75b6df0b9ccbf85817ca9ef030023e3?

Also, the main benefits of serializing this way is that it doesn't depend on the boost version. It's also is faster and more compact. In the provided benchmark, I get the following results:

21 ms. to save model with protobuffer.
14 ms. to load model with protobuffer.
433 ms. to save model with boost.
756 ms. to load model with boost.

The boost model is 24MB and the protobuf model is 3MB.

kentonl:
Closing this for a cleaner pull request.

PR #185, opened 2016-11-22
ability to get expression's dimension

yoavg:


neubig:
Yes, this is something that we should have had long ago.

Issue #84, opened 2016-09-20
Better model saving/loading


neubig:
Currently, cnn models use model the boost serialization library to perform model saving/loading. This is easy, but only works if you use the same version of boost to save/load files. Ideally, we can create a version of model saving that works across boost versions, perhaps by creating a custom boost serializer that reads/writes the parameters appropriately in text format.

honnibal:
I found the Boost version stuff the trickiest part of installing the library.

Is it possible to remove the dependency on Boost altogether? Does Eigen need Boost deeply, or only for serialization?

If the Boost dependency could be dropped, the library could be packaged for PyPi, so that "pip install dynet" would work.

neubig:
Model saving/loading relies on boost, so until this issue is resolved, the
dependency to boost definitely cannot be removed. Serialization is the main
place where it's being used, but it looks like there is also code for
threaded computation and calculation of the digamma function that uses
boost and may not be trivial to remove.

Of course, reducing reliance on external libraries is generally a good
thing, and with c++11 we can largely live without boost, so it might be
something to think about.

neubig:
We are currently considering a YAML-based approach to improving model saving/loading in DyNet. Using a library such as this may be useful: https://github.com/jbeder/yaml-cpp

xunzhang:
I think YAML format is more readable and easy to maintain as an independent module. By using these popular formats we could also deal with different languages. But I suggest considering some official libraries together with their performance.

Since this work relates to lots of development and refactoring work, please try to give a proposal first(for example, use YAML format..) and then people from the community could contribute, in case no contributions are conflicted or in vain. Thanks.

neubig:
Thanks for the comment @xunzhang . I did post this in hope of keeping people informed. One thing that we've been thinking about a little bit is whether to rehaul the overall model saving/loading API, or to just do a simple fix of the current code. Here are details of the two options: 

--- Simple fix
Probably the shortest path to doing this would be to re-write something very similar to boost::archive::text_iarchive that instead is implemented directly in DyNet and outputs YAML.

--- More complicated fix
Something that might be nicer would be to re-haul the model format a little bit so that it outputs things in a more easy-to-read format where the model specification (structure of builders, etc.) is written out first, then the data comes later in the file. This means that people could open the file without being overwhelmed by the model parameters themselves.

With regards to comparing official YAML libraries, I think this is something that could be done, but I've heard good things about yaml-cpp. If anyone else has experience I'd love to hear.

[...]

PR #188, opened 2016-11-22
Protobuffer alternative for serializing and deserializing models.

kentonl:
### Model serialization with protobuffers
This is a proof of concept for model serialization using protobuffers. The main benefits of this method is that it doesn't depend on the boost version, it's faster, and it's more compact. The provided benchmark gets the following results:

```
21 ms. to save model with protobuffer.
14 ms. to load model with protobuffer.
433 ms. to save model with boost.
756 ms. to load model with boost.
```

The boost model is 24MB and the protobuf model is 3MB.

### Usage
Call the following functions after adding the parameters to the model.
```
SerializationUtil::Save(const std::string &model_file, const Model &model) ;
SerializationUtil::Load(const std::string &model_file, Model *model) ;
```

### Limitations
* The protobuffer code currently lives entirely in the `proto` directory, since it's not clear how to properly integrate it into DyNet.
* Additional dependency on the protobuf library.
  * `sudo apt-get libprotobuf`
  * `brew install protobuf`
  * currently break Travis.
* It only supports the serialization and deserialization of `dynet::Model`. 
  * hint: you can compose protobufs to support additional configurations and remove dependencies on boost serialization entirely.
* `add_parameters` and `add_lookup_parameters` must be called in exactly the same order.
* No support for python yet (not familiar with that side of DyNet).


proto/tensor.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message ParameterProto {
+  optional DimensionsProto dimensions = 1;
+  repeated TensorProto value = 2;
+}
+
+message ModelProto {
+  repeated ParameterProto parameter = 1;
+  repeated ParameterProto lookup_parameter = 2;
+}
```
kentonl: Good catch. I've included the serialization of L2WeightDecay in the latest commit.

neubig:
Hey @kentonl , just FYI, we are currently working on re-tooling the DyNet model saving/loading interface according to this: https://github.com/clab/dynet/issues/84

We aren't going to go with protobuf as the default option as it introduces an additional dependency on an external library, but we're thinking of having it as an option. Once we have the basic interface we'll probably go back and reference this pull request. If you're interested in helping at that point we'd love to have your help. (That's why I was keeping this open instead of closing it)

PR #190, opened 2016-11-22
Re-inputting the generated output

talbaumel:


neubig:
Thanks, looking good!

Issue #201, opened 2016-11-29
Inconsistent usage of pointers and references


mitchellstern:
As it stands, most of the builder classes in `dynet/` use `Model*` in their constructors, whereas most of the builder-like classes in `examples/` use `Model&`. On the other hand, `ComputationGraph`s are largely used as references throughout the code base, whereas there are a small number of pointers (some of which are unused, as in `dynet/treelstm.h`). There's even a mismatch in the current version of `c2w.h`, where (among other issues) a `ComputationGraph*` is being used where a `ComputationGraph&` is expected by `RNNBuilder`.

I'm not sure I can speak for the internal DyNet code, but for all of the builders and example models, is there any compelling reason why we shouldn't be using references to `Model`s and `ComputationGraph`s throughout? If not, I'd be happy to make a pull request to this effect for consistency.

(Possibly related to issue #90.)

neubig:
Thanks for taking a look and offering to help with this! It is indeed related to missing coding guidelines of https://github.com/clab/dynet/issues/90 .

I propose that we use the following as the coding standard for pointer vs. reference in DyNet, and welcome any pull requests to fix things:
http://stackoverflow.com/a/114351

pmichel31415:
@mitchellstern are you currently working on this? I can start working on it later this week so let me know what I can do

mitchellstern:
Thanks for offering to help! I started making some changes to this effect earlier this week -- let me get back to you here in the next day or two with an update and we can see whether it would make sense to parallelize.

mitchellstern:
@pmichel31415: I've made an initial pass at this in #207. I tried to be fairly comprehensive, but please feel free to supplement wherever needed in case I missed something.

neubig:
This has been fixed in: https://github.com/clab/dynet/commit/97867c5e29a0be7bf7a893c9cf78224b89fc9a15

Thanks @mitchellstern !

PR #207, opened 2016-12-11
Changed pointers to references where `nullptr` is invalid

mitchellstern:
This standardizes the code base according to @neubig's suggestion in #201. A few minor warnings are also fixed.

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -11,7 +11,7 @@ Create a model, and an SGD trainer to update its parameters.
 .. code:: cpp
 
     Model mod;
-    SimpleSGDTrainer sgd(&mod);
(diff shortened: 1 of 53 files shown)


neubig:
Thanks! This looks good. This will break external software toolkits that rely on DyNet's C++ API, so I think we should notify people on the dynet-users mailing list, and then pull in a day or so.

mitchellstern:
Sounds good, thanks! I think the major API change to note is that all builders (e.g. all RNN and softmax variants) and all trainers now take a `Model&` instead of a `Model*` during construction. In addition, a few utility functions (e.g. in `dict.h` and `pretrain.h`) were modified to take references instead of pointers where appropriate.

PR #210, opened 2016-12-13
Create python_tutorial.rst

nsaphra:
I think there's a large enough python user base that this is a necessary addition, but I haven't really expanded it. It's barely more than a stub, with a little info pulled from a change log.


yoavg:
Thanks!

PR #211, opened 2016-12-13
Update tutorial.rst

nsaphra:
Since I created a new python tutorial file.


pmichel31415:
Apparently with this https://nbsphinx.readthedocs.io/en/0.2.11/ ipython notebook can be embedded in sphinx documentation, this might be a good start.

 I'll take a look and see if it's possible to integrate with the readthedocs pipeline

PR #212, opened 2016-12-13
Add ipython notebooks to the docs

pmichel31415:
Notable changes : 

- Tutorials reorganized in docs, c++ and python are separated
- ipython notebooks from examples/tutorials are included in the python docs (if it doesn't break readthedocs   *fingers crossed*
- The ipython notebooks have been slightly modified to fit in the doc hierarchy. Most headers are downsized from 1 level in markdown

N.B. : this PR is redundant with #211

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -19,8 +19,9 @@ import subprocess
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#sys.path.insert(0, os.path.abspath('.'))
-
+# sys.path.insert(0, os.path.abspath('../../examples/tutorials'))
+if not os.path.islink('tutorials_notebooks'):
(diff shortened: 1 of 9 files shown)


yoavg:
Is there a way to see how the docs would look like after merging the PR?

pmichel31415:
For lack of a better medium, here are screen captures from my local version (don't mind the ugly default theme)

---
Global menu (in index.rst)

menu
---

Tutorial menu (tutorial.rst)

menu_tuto
---

Example of a notebook

notebook

yoavg:
Looks good, thanks.
Just to verify, the tutorials notebooks still remain at the previous location as well, right?

And one tiny addition before I merge: could you please add a link also to https://github.com/clab/dynet_tutorial_examples ?
we had it linked from the main page before but it got dropped at some point..

pmichel31415:
Yes, the solution for now is to create a symlink to the notebooks during the doc compilation. This is problematic for windows but it's the best I could do with sphinx' limitations.

Thank you for the link I'll add it, you can merge after my next push

PR #215, opened 2016-12-14
Update index.rst

nsaphra:


yoavg:
@nsaphra @pmichel31415 hi, I am trying to figure out the relation between #211  #212 #215 .
#212 definitely goes in, but are #211 and #215 compatible / needed?

pmichel31415:
I think #212 takes care of #211 and #215  by reorganizing the tutorials

nsaphra:
@yoavg As far as I can tell, without #215, the changes I made in #210 won't be visible in navigation. Correct me if I'm wrong, but currently the python tutorials don't show up in the index or on the sidebar navigation. 

I think #211 should be merged because adding the python tutorials as a separate file made the deleted python tutorial section redundant.

nsaphra:
I see that #212 reverts back to including python in the main tutorial rather than in its own file and makes #211 and #210 unnecessary. I will close them.

PR #217, opened 2016-12-14
Mnist example

pmichel31415:
This adds a the standard MNIST classification problem to the example pool (with benchmark on cpu)

The `mlp.h` file implements a simple feed forward network builder, it might even be useful to add it to the core dynet functionalities alongside the rnn builders, tell me what you think of it


neubig:
Looks good, thanks!

PR #222, opened 2016-12-22
Python 3 bugfix

mitchellstern:


yoavg:
Range creates a list in python 2 and is slower. Python 3 support is nice to have, but need to quantify the performance hit we get because of this.

mitchellstern:
Cython seems to be able to compile `xrange` even for Python 3, though I'm not sure how it's handled under the hood -- perhaps this commit can be omitted. The other issue is a genuine bug for Python 3 though; deserialization of `BiRNNBuilder`s isn't working correctly at the moment.

yoavg:
If cython handles the xrange for python 3, lets keep it as xrange. The list in the serialization is indeed important (and in a speed where speed doesn't matter at all) so lets fix only that. Can you change the pull request to include only the list fix?

mitchellstern:
Done! Now only the bugfix is included.

Issue #218, opened 2016-12-21
Simplify/robustify LogSoftmax backward calc


yv:
The gradient value for logsoftmax should be something like "grad - sum(grad)*softmax(x)" (which Theano and TensorFlow use).
```
z.t<0>().device(*dev.edevice) = fx.t<1>().binaryExpr(dEdf.t<1>(), FWeightedError()).sum();
Eigen::array<int, 1> bcast; bcast[0] = fx.d.rows();
dEdxi.t<1>().device(*dev.edevice) += fx.t<1>().exp() * -z.t<1>().broadcast(bcast) + dEdf.t<1>();
```
or, in cnn1
`float off_diag_sum = -(*fx).binaryExpr(*dEdf, FWeightedError()).sum();`
where FWeightedError calculates `expf(t) * d / expf(t)`, which is `d` in most cases and **nan** in cases where t is large and negative, (which means the nan will propagate and clip_gradients will abort the program).

Simplifying this to the moral equivalent of
`float off_diag_sum = -(*dEdf).sum();`
would remove the occurrence of nans in some cases.

PR #223, opened 2016-12-22
Simplify LogSoftmax (#218)

yv:
Patch for #218


neubig:
Wow, it's hard to believe that FWeightedError function even exists, as it's essentially a noop except when it breaks your whole program. Perhaps you could also delete it?

Also, could we either speed-test -ffinite-math-only or remove it from this pull request until we can?

PR #232, opened 2017-01-05
Documentation for the Dim object

pmichel31415:
For now I've put it in the minibatching part of the documentation. Not sure if this is the best place.

One remark : I'm not too sure of what the "truncate" function achieves so if someone knows... I left it as a TODO for now.

On a related note I think there should be a "Core Dynet" section in the docs with documentation for the computation graph and other core classes like parameters and such. I'll look into it but if someone with more experience in the internals wants to start, go ahead.


pmichel31415:
OK I removed it from the sphinx source for now. It can now be merged, at least the code is somewhat documented

PR #235, opened 2017-01-10
fix windows compiling error by modifying the unsupported initializer_list

Oneplus:
syntax like `Eigen::array<int, 1> x({1});` is not supported on VS15 (MSVC19.0.) and cause compile error on windows. change to `Eigen::array<int, 1> x = {1};`


neubig:
Thanks a lot!

PR #241, opened 2017-01-13
fixing cmake error with defined BOOST_ROOT

cherryc:
Previously, cmake would fail if BOOST_ROOT was defined but
Boost_INCLUDE_DIR was not defined, now it does not


neubig:
Thanks a bunch!

PR #243, opened 2017-01-14
Refactor of the attention example

emanjavacas:
This pull request attempts to (i) avoid code duplication and (ii) make more explicit the difference between decoding during training and testing.
After some thoughts on how to accomplish this, I settled on python support for co-routines using generators (and their `next` and `.send(arg)` methods). I think this makes the code more clear and readable in terms of intentions, and also helps encouraging python literacy on co-routines.

### DIFF
--- a/examples/python/attention.py
+++ b/examples/python/attention.py
@@ -80,29 +80,50 @@ def attend(input_vectors, state):
     return output_vectors
 
 
-def decode(dec_lstm, vectors, output):
-    output = [EOS] + list(output) + [EOS]
-    output = [char2int[c] for c in output]
+def make_decoder(input_sequence):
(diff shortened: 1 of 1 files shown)


yoavg:
This is really cool, but I am a bit worried that the `.send(arg)` mechanism is a bit too obscure for a "tutorial" code (which is not about teaching generators).  

Perhaps some dedicated, detailed comments explaining the architecture could help.

emanjavacas:
Yes, I was also worrying the same when I was coding it. Anyway, I think I will just make a blog post out of that producer-consumer example and I will try to add some inline comments as you suggest. So, I will close this and create a new pull request once I am done.

PR #249, opened 2017-01-16
Updating install docs to be consistent with build

jkkummerfeld:
Minor change to show the correct executable.

For new users it may be obvious, but I hit the problem that I still had 'examples/xor' from a previous build, and so when I tried to run it I got a segfault.


neubig:
Great, thanks!

PR #254, opened 2017-01-17
Fixing MKL build for OS X

jkkummerfeld:
On OSX the intel installer arranges the MKL directory slightly differently. These changes account for that (tested on OSX 10.12 Sierra) and I also added a note about where to find the library usually.


neubig:
Many thanks!

PR #256, opened 2017-01-18
Added Mac installation instructions and eigen

alvations:
Added explicit instructions for Mac and explicit installation instructions for eigen.

Hopefully the bolded phrases and explicit instructions can help users identify the correct installation instructions for their OS and reduce pains of installation.


neubig:
Awesome, thanks!

Issue #255, opened 2017-01-17
Setting dynet memory in jupyter notebook, using an API


napsternxg:
I couldn't find a way to set the memory of dynet in jupyter notebook. I have resorted to the following hack:
```
import sys

try:
    __file__
except:
    sys.argv = [sys.argv[0], '--dynet-mem', '10000']

"""
Outputs:
[dynet] random seed: 2618859252
[dynet] allocating memory: 10000MB
[dynet] memory allocation done.
"""
```

Is there a better way to set the memory while working from the notebook?

pmichel31415:
Hi,

I'm working on making this available like in the c++ code, expect a pull request by tomorrow if everything goes well.

RobRomijnders:
How did this end up

pmichel31415:
Hi I did a pull request regarding this ( #258 ) but it's still waiting for @yoavg 's approval.

If it's urgent you can clone directly my branch https://github.com/pmichel31415/dynet/tree/python_params (I just put it up to date with the master branch)

RobRomijnders:
Thanks, trying that now

RobRomijnders:
I'm getting AttributeError

`AttributeError: 'module' object has no attribute DynetParams`

pmichel31415:
Hmm weird it works for me. Have you tried running it from the build/python folder of this fork? If it works then it's probably because your  environment variables point to the master branch

RobRomijnders:
Thanks for the fast responses!
How would that work? Both 
`git clone https://github.com/pmichel31415/dynet/tree/python_params` and
`git clone https://github.com/pmichel31415/dynet/tree/python_params.git` 
result in errors for me.

So I ended up doing
`git clone https://github.com/pmichel31415/dynet.git`

pmichel31415:
Oh ok i thought it would clone the branch directly my bad.

Then

    git clone https://github.com/pmichel31415/dynet.git
    git checkout python_params

Or maybe

    git clone -b python_params https://github.com/pmichel31415/dynet.git

should do the trick

RobRomijnders:
Checking that now:)

RobRomijnders:
Works! Thank you!

pmichel31415:
Great. If you ever have the opportunity to test the gpu parameters, let me know if they work, I currently don't have access to a gpu.

RobRomijnders:
Sure, I'm happy to help too. How could I test that?

pmichel31415:
[...]

PR #258, opened 2017-01-18
Accessing and modifying dynet parameters from python

pmichel31415:
This is an attempt to address #255 by allowing the user to set the dynet options programmatically.

Here is the new behaviour, copy-pasted from the documentation : 

---

There are two ways to import the dynet module :

    import dynet

imports dynet and automatically initializes the global dynet parameters with the command line arguments (see the [documentation](http://dynet.readthedocs.io/en/latest/commandline.html)). The amount of memory allocated, GPU/CPU usage is fixed from there on.

    import _dynet
    # or
    import _gdynet # For GPU

Imports dynet for CPU (resp. GPU) and doesn't initialize the global parameters. These must be initialized manually before using dynet, using one of the following :

    # Same as import dynet as dy
    import _dynet as dy
    dy.init()

OR

    # Same as import dynet as dy
    import _dynet as dy
    # Declare a DynetParams object
    dyparams = dy.DynetParams()
    # Fetch the command line arguments (optional)
    dyparams.from_args()
    # Set some parameters manualy (see the command line arguments documentation)
    dyparams.set_mem(2048)
    dyparams.set_random_seed(666)
    dyparams.set_weight_decay(1e-7)
    dyparams.set_shared_parameters(False)
    dyparams.set_requested_gpus(1)
    dyparams.set_gpu_mask([0,1,1,0])
    # Initialize with the given parameters
    dyparams.init() # or init_from_params(dyparams)

---

The main change is that now importing `_dynet` or `_gdynet` DOES NOT initialize dynet (ie call `dynet::initialize()` so it has to be done manually.

This is the easiest I could do while preserving the behaviour of `import dynet`, however I don't use the python bindings a lot so this might not be the best design choice for actual users, let me know.

Also important  : NOT TESTED ON GPU


neubig:
Looks good to me, but I'll let @yoavg give the go-ahead on this.

neubig:
If @yoavg doesn't chime in on this soon I'd suggest just incorporating this. It looks good to me.

PR #259, opened 2017-01-19
Homebrew on macOS can install the dev version of Eigen

danieldk:
Thanks for DyNet. This is a small modification to the documentation to point out that on Homebrew/macOS you can install the development version of Eigen with Homebrew as well.

This has the benefit that Eigen is manageable through Homebrew (and it makes /usr/local less messy ;)).


neubig:
Thanks! This is good, but I think we should put the directions after the "blessed" version of Eigen and the warning about needing to be adventurous to use the latest version :)

PR #261, opened 2017-01-20
Yarin Gal dropout for vanillalstm

pmichel31415:
This is an implementation of [the yarin gal paper on dropout for LSTMs](http://papers.nips.cc/paper/6241-a-theoretically-grounded-application-of-dropout-in-recurrent-neural-networks).

---

Important points : 

- Inverted dropout (after checking the code of the paper [here](https://github.com/yaringal/BayesianRNN/blob/master/LM_code/main_new_dropout_SOTA_v3.lua) they do use inverted dropout.

- "tied" dropout, ie at each timestep the mask is applied to h_{t-1} and x_t only. According to the paper this is faster and in most cases on par if not better than applying a mask on each gate. The lua code also uses this.

- 2 different dropout rates : one for x and one for h. there is an overloaded `set_dropout` method taking two floats to handle that. The original `set_dropout` with one argument sets both dropout rates to the same value.

- For batched input we need to create different masks for each batch element. Technically this means we need to know the batch size in `start_new_sequence`. Instead of changing the interface which would require to change it for all `RNNBuilders` I just added a `set_dropout_masks(float bsize)` method which needs to be called after `start_new_sequence` if one wants differentiated masks. This is not an ideal behavior so I'm open to suggestions.

---

Experimental results : 

I trained two architectures on the PTB, using more or less the same code as the batched rnnlm example.

I used the same hyperparameters as yarin gal as much as possible.

The notable difference is that I'm training on sentences whereas they just feed the network contiguous sequences of fixed length (across sentences).

yarin_gal_comp

The legend reads like this : [number of layers]x[input dim]x[hidden dim] [LSTM dropout rate]x[embedding/last linear layer dropout rate]

In Gal's paper they attain a test ppl of around 82 with the 650x650 model. As you can see it still hasn't converged (one epoch takes 40-50 minutes on CPU with a batch size of 20 though, I'll probably have more results tonight)


neubig:
+1 +1 +1! This is great, but two questions:
1) Do you think this will work out-of-box for the coupled LSTM. If so, we should do that as well while we're at it. If there are some potential problems, then we can merge this for now and get back to the LSTM later. The main thing I'm worried about is if we're getting different flavors of dropout based on which builder we select.
2) Do the python bindings work with this?

pmichel31415:
Ah yes I just thought about this myself :)

 1. Should work modulo some adaptation. I can probably do this today
 2. The existing python bindings should still work, but I need to add some for the new methods.

I'll try to do this today or over the weekend

neubig:
Great, thanks!

pmichel31415:
OK I did the LSTMBuilder. 3 remarks : 

- For compatibility issues I had to do some ugly `if`s in the serialization method because the previous one didn't save important information.
- I use the same mask to drop c_t and c_t-1 for the output (resp input) gate
- Currently running language modelling benchmarks

neubig:
Awesome, sorry about the fragmentary requests but one more thing: could you add a quick note to the LSTMBuilder doc about which method we're using? Pending that, and the benchmarking sanity check I think we're good to go.

pmichel31415:
FYI here are the results for the coupled LSTM, using the same small architecture as above (2 x 256 x 256) and smaller dropout.

Seems much worse, although dropout still performs better.

I'm trying with smaller architectures. 

coupled_drop_nodrop

pmichel31415:
OK I added the documentation.

I've had troubles implementing python bindings for the subclass specific methods, if someone knows how to do it with the current state of things in the cython code...

In the meantime I think you can merge

neubig:
This looks great. The relatively bad performance of the LSTMBuilder is a bit worrisome given that this is what is used in most examples, so I'll do some testing on this. But this is orthogonal to this commit, so I'll go ahead and merge. Thanks a lot!

Issue #139, opened 2016-11-02
Throwing exceptions or aborting


odashi:
DyNet basically aborts the program when some fatal errors occurred (e.g. OOM).
However, this behavior is sometimes not desirable because abort() deprives users of opportunities to deal with these problems themselves.
Would you like to switch all abort() terminations to throwing an exception if possible?

If the library would like to keep the current behavior, it may be possible to add an Boolean parameter to DynetParams to switch the behavior in case of fatal errors (i.e., throwing an exception or aborting).

neubig:
Yes, I think it's a great idea to replace all places where we abort or assert with exceptions. I think it may be a good idea to create one macro that allows us to throw an exception with a message, and another macro that allows us to check a condition, then throw an exception with a message if the condition is false. Changing all the error handling in DyNet at one time would be a big task, but I welcome any pull requests in this direction!

odashi:
OK, I am just thinking about similar implementations to do that.
I also guess that this issue is actually related on the error-handling policy of DyNet, and it is better to be written in some documents for all application developers, e.g., what kind of exceptions would be thrown from the functions.

neubig:
I definitely agree. The need for coding conventions is a known issue (https://github.com/clab/dynet/issues/90), and we have a place in the doc reserved for this. Feel free to add to this doc if you'd like. I'll also try to fill it in soon.

pondpremt:
Not sure if this is the same problem, but I was playing with dynet in ipython shell, and mismatched input dimensions cause the shell to abort. Would be nice if it only just threw an error:

```
In [49]: a * W
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Mismatched input dimensions in MatrixMultiply: [{3} {2,3}]
[1]    2479 abort      ipython
```

neubig:
This problem has been noted in the coding style conventions: http://dynet.readthedocs.io/en/latest/code_style.html

This will need a concentrated effort to fix, and we welcome any contributions.

neubig:
Fixed by https://github.com/clab/dynet/commit/56e3b7aef59b9f576c4a4780733ca91614ed2520

PR #264, opened 2017-01-22
To make the DeviceMemPoolSizes constructor more robust

MalcolmSun:
Add `cerr<<The format of --dynet-mem is invalid` and `abort()` to avoid program encoutering some unexpected problems when the --dynet-mem's length  is not 1 or 3.

### DIFF
--- a/dynet/devices.cc
+++ b/dynet/devices.cc
@@ -7,6 +7,7 @@
 #include "dynet/cuda.h"
 #include "dynet/dynet.h"
 #include "dynet/expr.h"
+#include "dynet/except.h"
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


neubig:
Thanks! But we're actually in the process of changing to throwing errors instead of using abort (https://github.com/clab/dynet/issues/139). Would you mind changing the format?

MalcolmSun:
Ok, I will change the format, then create a new PR.

MalcolmSun:
In the new commit, I have changed `abort()` to `throw`. `invalid_argument_format` exception class extends `std::invalid_argument`.

MalcolmSun:
In the new  commit, I have changed the format and keep it consistent with code style.

neubig:
Great, thanks!

PR #267, opened 2017-01-24
Corrected path to examples as part of Python DyNet module installation

ratishsp:
Corrected path to examples as part of Python DyNet module installation.


neubig:
Thanks!

PR #269, opened 2017-01-25
DynetParams needs ctor to initialize the variables correctly

kecsap:


neubig:
Thanks! This is great (I wasn't aware of the crashes), but could you use standard C++ constructor style when initializing variables?

`DynetParams::DynetParams() : initialize_variables_here ... {`

PR #272, opened 2017-01-26
Removes regex header (boost regex is used instead)

stefan-it:
This PR removes the (standard) `regex` header in `train_rnnlm.cc`, because the regex header from Boost is used (and later included) to avoid conflicts.


neubig:
Thanks!

PR #275, opened 2017-01-29
Immediate computation (imperative style) (only C++ now)

cheezer:
Added immediate computation mode and INF/NaN checking to make it easy to debug. Two boolean variables in ComputationGraph are used to activate them. Immediate computation is performed if immediate_compute is set to True. Inf or Nan is checked if check_validity is set to True in immediate computation mode.

### DIFF
--- a/dynet/dynet.cc
+++ b/dynet/dynet.cc
@@ -85,8 +85,10 @@ void Node::backward(const std::vector<const Tensor*>& xs,
   }
 }
 
-ComputationGraph::ComputationGraph() :
+ComputationGraph::ComputationGraph(bool ic, bool cv) :
   ee(new SimpleExecutionEngine(*this)) {
+  immediate_compute = ic;
+  check_validity = cv;
   ++n_hgs;
   if (n_hgs > 1) {
     cerr << "Memory allocator assumes only a single ComputationGraph at a time.\n";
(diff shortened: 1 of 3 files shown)


neubig:
This looks great! But just a couple things:

1) I think perhaps this functionality is not major enough to add to the constructor. How about we keep the constructor as-is and add some setter functions instead.
2) After doing this, could you comment the functions in the standard way? If possible it'd also be nice to add a "debugging" section to the doc, but maybe this can be another commit.

cheezer:
Sure!

cheezer:
Just updated!

neubig:
Looks good, but one more thing: instead of using assert(), let's throw an exception. I know other places in the code are still using assert(), but we're trying to phase this out. After this is finished I think this is ready to incorporate. We'll do the python interface in a separate PR.

cheezer:
Just updated. When it throws an exception, it doesn't give the specific line that goes wrong. Other exceptions also doesn't print stacktrace explicitly. Is it better to print stacktrace?

When it detects the Inf in restricted log softmax, the error message is as follows:

[dynet] random seed: 2093637403
[dynet] allocating memory: 10MB
Running 92 test cases...
NaN or Inf detected
unknown location(0): fatal error in "restricted_log_softmax_gradient": std::runtime_error: NaN or Inf detected
/home/qizhe/tools/dynet/tests/test-nodes.cc(551): last checkpoint

*** 1 failure detected in test suite "DYNETBasicTest"

neubig:
Debuggers such as gdb or valgrind can be used to get a stack trace. I think this should be added to the "debugging" section of the doc. I'll merge this now, and add doc in a bit.

cheezer:
I can work on the doc this weekend if it is not urgent. Sorry I was busy with a homework yesterday..

neubig:
No problem! Doc would be greatly appreciated. I think you can make a new "debugging" section in the top level of the doc. I'll add some other stuff there as well after the initial commit.

Issue #274, opened 2017-01-29
boost::archive::archive_exception loading BiRNNBuilder


danielhers:
The following code fails:
```
import dynet as dy
              
m = dy.Model()
b = dy.BiRNNBuilder(2, 100, 200, m, dy.LSTMBuilder)
m.save("bilstm.model", [b])
m2 = dy.Model()
m2.load("bilstm.model")
```
Getting this error:
```
loading model 
done 44       
terminate called after throwing an instance of 'boost::archive::archive_exception'
  what():  input stream error
Abort
```
It's not clear from this error message what the problem is, but it seems to happen after initializing the loader ("done 44").

danielhers:
A version of DyNet from 6 days ago didn't have this problem, so apparently it was introduced since then.

pmichel31415:
OK I think I've located the problem, I introduced this bug with the dropout pull request.

As a hotfix I can disable the serialization of a bunch of new fields in LSTMBuilders. This will make it impossible to store different dropout rates but at least it will be backward compatible.

danielhers:
I don't mind updating my code, but at least make it clear what should be
changed and make the error message clear. It's not a problem with loading
an old model; it's right after saving it.

danielhers:
Thanks, fixed.

PR #277, opened 2017-01-30
Retro-compatible serialization

pmichel31415:
LSTMBuilder is returned to its previous version. This has 2 effects : 

- It is now impossible to save different dropout values for `x`, `h`, `c`. If anyone knows a clean, retrocompatible way to do this, let me know

- Since `input_dim` and `hid` are needed, I added a sanity check in `start_new_sequence` so that if these variables are not set, they are retrieved directly from the parameters (and fire a warning)

Addresses #274

### DIFF
--- a/dynet/lstm.cc
+++ b/dynet/lstm.cc
@@ -321,11 +321,18 @@ void LSTMBuilder::disable_dropout() {
 }
 
 template<class Archive>
-void LSTMBuilder::serialize(Archive& ar, const unsigned int) {
+void LSTMBuilder::serialize(Archive& ar, const unsigned int version ) {
   ar & boost::serialization::base_object<RNNBuilder>(*this);
   ar & params;
   ar & layers;
   ar & dropout_rate;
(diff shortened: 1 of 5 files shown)


neubig:
Boost serialization supports versioning:
http://www.boost.org/doc/libs/1_63_0/libs/serialization/doc/tutorial.html#versioning
Will this help?

Also, if possible it'd be nice to add a unit test with the issue code (https://github.com/clab/dynet/issues/274) to make sure that the bug has been fixed.

yoavg:
maybe this is a good time to reconsider a move from boost to something more
light-weight (and controllable) like in the protobuf PR?
https://github.com/clab/dynet/pull/188

pmichel31415:
@yoavg I agree that would be very convenient.

In the meantime, I added boost versioning for the LSTMBuilder as well as unit tests for the serialization of LSTM, VanillaLSTM and SimpleRNN.

PR #279, opened 2017-01-31
cpp implementation of attention.py in examples/python

ratishsp:
This is cpp implementation of attention.py example from examples/python.
I have tried to make an exact mapping of python code to the corresponding cpp code and have verified that the program converges at the same rate as the python example.
The trace of the last few lines of output:

0.0435848
it is working
0.0415076
it is working
0.0396145
it is working
0.0378819
it is working
0.0362891
it is working

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -4,7 +4,7 @@ if (NOT MSVC)
    set(MP_EXAMPLES rnnlm-mp xor-mp)
 endif()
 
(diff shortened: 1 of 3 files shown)


pmichel31415:
Wow thank you for this.

Just a few comments : 

- Could you switch to 2 spaces indent for consistency with the rest of the code base?
- Since this is meant for people new to DyNet/NLP, could you add comments. ideally : 
   - doxygen style comments in the .h file to document the functions
   - inline comments in the .cc file to explain what's being done

Cheers

ratishsp:
Sure. 
This commit fixes indentation to 2 spaces and adds comments.

neubig:
Great, thanks so much for this!

ratishsp:
Thanks to all of you for this excellent library!

PR #286, opened 2017-02-07
Documentation for model and parameters

pmichel31415:
I started documenting some of the core dynet functionalities.

For now I've done the Model, Parameter and ParameterInit classes.

I put all this in a separate section of the documentation called 'core functionalities', the idea being to add stuff like ComputationGraph, Tensor and all later because this is the kind of useful things that one needs to do more than just copy-pasting the example code.

I also took this opportunity to replace some `abort()` by domain_error exception in parameterinit


neubig:
This is excellent, thanks!

PR #287, opened 2017-02-08
Compgraph doc

pmichel31415:
Another bunch of documentation (ComputationGraph & nodes). The descriptions are mostly copy pasted from pre-existing comments in the dynet.h/nodes.cc files.


neubig:
Many many thanks!

Issue #247, opened 2017-01-14
CMake requires the full Eigen path in some environments


yoavartzi:
In some environments, the [installations script](http://dynet.readthedocs.io/en/latest/python.html) fails. CMake seems to require the full path for Eigen, and not the relative one. I couldn't figure out why it happens in some environments and not others. Maybe worth adding a note.

stepelu:
I confirm it's failing on latest OSX + brew.

neubig:
Haven't had time to try this, but maybe this could solve the problem? http://stackoverflow.com/questions/39027269/cmake-convert-relative-path-to-absolute-path-with-build-directory-as-current-d

Pull requests welcome!

PR #288, opened 2017-02-09
dynet-247. Fix cmake bug giving relative Eigen path

xunzhang:
Fix https://github.com/clab/dynet/issues/247.


neubig:
Great, thanks so much! I didn't think it would be this simple.

Issue #262, opened 2017-01-20
Glorot initialization not quite right for sigmoids and ReLUs


pmichel31415:
I think the Glorot initialization is not quite right for sigmoids, and in general for any function that does not satisfy

![codecogseqn 2](https://cloud.githubusercontent.com/assets/10391785/22168023/f4952236-df37-11e6-9fd9-781b2bf1bca3.gif)

near 0. which is true for the tanh but not for the sigmoid (![codecogseqn 3](https://cloud.githubusercontent.com/assets/10391785/22168056/2e838258-df38-11e6-8cf0-9b0bb011cd07.gif))

(See (2) through (7) in [the original paper](http://www.jmlr.org/proceedings/papers/v9/glorot10a/glorot10a.pdf?hc_location=ufi),)

Relu seems to be a special case as well (see [He et al. (2015)](http://www.cv-foundation.org/openaccess/content_iccv_2015/papers/He_Delving_Deep_into_ICCV_2015_paper.pdf)).

One solution would be to mimmick the interface in lasagne (see [here](http://lasagne.readthedocs.io/en/latest/modules/init.html#lasagne.init.Glorot))

neubig:
Nice catch! Yes, I think we should fix this. If I remember correctly a lot of other toolkits separate the Glorot and He methods, right? Eventually (maybe not necessarily in this issue?) we'll probably want to add HeNormal as well.

neubig:
Fixed by https://github.com/clab/dynet/pull/295

PR #295, opened 2017-02-10
Better initializers

pmichel31415:
This does mainly two things : 

- Add a `gain` parameter to the `ParameterInitGlorot` constructor and modifies the doc accordingly. Indeed, for Glorot to be correct, the actual scale should be : 

eq

Where f is the activation function.

Here are quantitative results with a 2 hidden layers MLP with hidden size 512 and sigmoid activations on MNIST (averaged over 20 experiments)

glorot

- Implemented the Saxe initializer (random orthogonal matrix, useful for vanilla RNNs). Added unit test for this one.

Also added python bindings for all of those (tested)

Addresses #262

### DIFF
--- a/dynet/model.cc
+++ b/dynet/model.cc
@@ -65,7 +65,7 @@ ParameterStorage::ParameterStorage(const Dim& d, float scale) : dim(d) {
   default_device->allocate_tensor(DeviceMempool::PS, values);
   default_device->allocate_tensor(DeviceMempool::PS, g);
   TensorTools::Zero(g);
-  if(scale == 0.0f) {
+  if (scale == 0.0f) {
     ParameterInitGlorot init;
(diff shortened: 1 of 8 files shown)


pmichel31415:
Hmm so it seems the test for saxe init fails on ios but not on ubuntu

I don't have a mac so if anyone can debug it for me it would be really nice. I commented the test.

neubig:
This looks like it's likely a numerical precision error:

/Users/neubig/work/dynet/tests/test-params.cc:59: error: in "params_test/init_saxe": check diff < epsilon has failed [5.07169728e-09 >= 9.99999972e-10]

Maybe it'd be fine to make the checking threshold a little more loose?

pmichel31415:
Thank you for taking a look, looks like that fixed it.

PR #301, opened 2017-02-13
Fixed CUDA linker on mac?

neubig:
This potentially fixes some problems when linking CUDA-related things on Mac OSX.


neubig:
Woops, but this breaks linking on Linux apparently. Will be merged pending confirmation that it works on both OSs.

Issue #231, opened 2017-01-05
Struct and class names don't appear in doc


pmichel31415:
So I've tried to correct this bug a few times to no avail

Basically the sphinx generated doc doesn't show the name of a class/struct as it should (see picture) : 

![screenshot from 2017-01-05 11-20-27](https://cloud.githubusercontent.com/assets/10391785/21687985/37f9535e-d339-11e6-9c90-0a8925c16d69.png)

if anyone with previous knowledge on doxygen/breathe/sphinx knows more any help would be welcome, I've kind of given up on this for now

pmichel31415:
Fixed by #302

PR #302, opened 2017-02-13
Fix struct names not appearing in doc

pmichel31415:
FINALLY!

Addresses #231 

See https://github.com/michaeljones/breathe/issues/312


pmichel31415:
Merging because already tested on the separate branch

Issue #300, opened 2017-02-13
GPU (backend cuda) build problem


iamyoungjo:
I am having problem on build with BACKEND=cuda. My system is OS X (10.11.6 El Capitan)
cmake works fine, once I do "make -j 4", it returns following error:
>Undefined symbols for architecture x86_64:
> ...
>ld: symbol(s) not found for architecture x86_64
>clang: error: linker command failed with exit code 1 (use -v to see invocation)

As I use GPU in tensorflow without any issues, I doubt that my CUDA setting has not done or something.

I searched similar issues here but I am the only person who's having this issue as it seems.
There's no issue on compilation if I don't put cuda backend.

If I missed a significant step here or if anyone who's familiar with this error, please help.
I wasted more than 6 hours because of this.

[make.log.zip](https://github.com/clab/dynet/files/771468/make.log.zip)

neubig:
I was able to replicate this, so I'll try to figure out the problem.

neubig:
This is potentially fixed by https://github.com/clab/dynet/pull/301

Could you try this out and see if this fixes the problems?

iamyoungjo:
I just tried, and it shows a different error:
> ld: library not found for -lcublas
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> make[2]: *** [dynet/libdynetcuda.dylib] Error 1
> make[1]: *** [dynet/CMakeFiles/dynetcuda.dir/all] Error 2
> make: *** [all] Error 2

[make.log.zip](https://github.com/clab/dynet/files/771669/make.log.zip)

iamyoungjo:
Any possibility that I am missing any library? Here'e my sample run for cuda setting validity check.
[cuda_test.log.zip](https://github.com/clab/dynet/files/771711/cuda_test.log.zip)

iamyoungjo:
@neubig any ideas? I am still stuck in there..

xunzhang:
@iamyoungjo According to your log, the reason why it is still failing is that the default linking args `-lcublas` could only search in library files from`/usr/lib` `/usr/local/lib`, while the default path for cuda is `/usr/local/cuda/lib`.

A work around should be expending search path using `DYLD_LIBRARY_PATH` under osx, like this:
`export CUDA_HOME="/usr/local/cuda"`
 `export DYLD_LIBRARY_PATH="/usr/local/cuda/lib"`
[...]

PR #303, opened 2017-02-14
Fix CUDA linker issue with -lcublas.

xunzhang:
This pull request solves [dynet-300](https://github.com/clab/dynet/issues/300). The root cause for dynet-300 was mentioned here at https://github.com/clab/dynet/issues/300#issuecomment-279583244.

Document for reviewing this pull request: https://cmake.org/cmake/help/latest/module/FindCUDA.html.


neubig:
Thanks! That was it.

PR #304, opened 2017-02-14
Speedup softmax/logsoftmax for GPU

mattr1:
Removing two extra memcopies (m and z) which required GPU synchronization. Also fix tabs->spaces in a couple places.

For my code, sped up GPU computation about 5-10%.


neubig:
Great! I hadn't realized the non-minibatched version still had this memory copy.

PR #306, opened 2017-02-14
Update log computation for MSVC to match the inline assembly version.

mattr1:
Also changed conditional-compile symbol to specifically check for MSVC (using MSC_VER), rather than WIN32, which only checks whether you are compiling for x86/x64 specifically.

### DIFF
--- a/dynet/cuda.h
+++ b/dynet/cuda.h
@@ -43,11 +43,14 @@ struct DynetParams;
 class Device;
 
 inline std::pair<int, int> SizeToBlockThreadPair(int n) {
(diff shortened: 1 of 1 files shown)


neubig:
Looks good to me. Thanks.

mattr1:
@neubig Thanks. Just so I know the convention you prefer: I did the change in a branch. Now that the branch is merged, I was planning on deleting the branch so that we don't fill up with feature branches. Is this how you prefer things be kept cleaned up?

neubig:
Yes, thanks.

Issue #92, opened 2016-10-05
Multi-device support


neubig:
Currently, cnn supports running computation on _either_ CPU or GPU, based on the setting of the -DHAVE_CUDA flag. It would be better if a single program could utilize both CPU and one or more GPUs within the same program. The framework for this is mostly there, minus a few remaining issues:

- [x] There are a number of places, mainly in the parameters and tensor operations that are still hard-coded based on the setting of HAVE_CUDA. These all need to be changed to be dynamic based on the device setting for the tensor.
- [x] Currently parameters are all declared on the default device. There needs to be an option to let parameters live on other devices.
- [x] We need to create a node that moves memory between devices.
- [x] Test code for multi-device support needs to be created.

iamyoungjo:
@neubig Hello. Is this enhancement still on-going? If not, I would like to jump on this as I have some time & purpose to work on this.

xunzhang:
Hi, I have an open pull request https://github.com/clab/dynet/pull/382 but I am currently working on other features so you can go ahead. If I have time, we can work on it together. @neubig what about merging the slight change in that pull request first? Thanks.

neubig:
Done. Also, I think this would be a good place to start for multi-device support: https://github.com/clab/dynet/issues/443

xunzhang:
Thanks a lot, I will work on it after finishing dynet-84 and will keep following on this as well.

neubig:
I think this is basically done, although it might have some rough edges. Thanks @xunzhang !

xunzhang:
Add reference for the related pull request https://github.com/clab/dynet/pull/704 .

PR #307, opened 2017-02-14
Fixes in parameters

pmichel31415:
-  orthogonal -> orthonormal
-  warning for cuda version of saxe init (unimplemented) <= this is untested on gpu but copy-pasted from similar statements (basically `#ifdef HAVE_CUDA throw exception #else .. #endif`
-  More docs for parameterstorage (for low level fiddling)

### DIFF
--- a/dynet/model.cc
+++ b/dynet/model.cc
@@ -215,11 +215,10 @@ void ParameterInitGlorot::initialize_params(Tensor & values) const {
 }
 
 void ParameterInitSaxe::initialize_params(Tensor & values) const {
-#ifdef HAVE_CUDA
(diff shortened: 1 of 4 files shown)


neubig:
Just FYI, in preparation for multi-device support, using HAVE_CUDA to figure out whether memory is living on the CPU or GPU is deprecated: https://github.com/clab/dynet/issues/92 (a rule of thumb, HAVE_CUDA should never have an "else", it should only be used to add GPU functionality, not remove CPU functionality).

That being said, I realize that `tensor.cc` is one of the files that still hasn't been fixed yet. Could we at least keep the deprecated switches to be only in `tensor.cc` so it will be easier to fix all at once later?

PR #312, opened 2017-02-15
Speedup sum on GPU by merging together sets of sums (less overhead calling CUDA kernels)

mattr1:
Unrolled sum when all of the items have the same batch dimension, so there are less CUDA calls (reduces GPU overhead). Also added special-casing for sum of 2,3, or 4 items (with same batch size). I think a sum of 2 items is probably very common, so it's worth special-casing. My experiments were 1-4% faster on GPU with no speed difference on CPU. Adds complexity to the function, but I felt it was worth the tradeoff for a little speed boost (and probably more significant speed boost for people doing a lot of summing).


neubig:
Looks good! We had something like this before, but it wasn't implemented for GPU. Will be nice to have it again.

PR #313, opened 2017-02-16
Added next-word-prediction, save and load model

vicmak:
Added an explicit function to get a softmax-ed distribution over the vocabulary, based on the prefix-context.
Added two functions for saving/loading a model to/from disk.

### DIFF
--- a/examples/python/rnnlm.py
+++ b/examples/python/rnnlm.py
@@ -20,6 +20,12 @@ class RNNLanguageModel:
         self.R = model.add_parameters((VOCAB_SIZE, HIDDEN_DIM))
         self.bias = model.add_parameters((VOCAB_SIZE))
 
+    def save2disk(self, filename):
+        model.save(filename, [self.builder, self.lookup, self.R, self.bias])
+
+    def load_from_disk(self, filename):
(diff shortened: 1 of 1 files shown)


neubig:
Thanks for this! Just a few quick requests:

* Following the DyNet coding conventions (http://dynet.readthedocs.io/en/latest/code_style.html) could you change all function names (including BuildLMGraph) to snake case, where all words are lowercase and separated by an underbar.
* The softmax calculation shouldn't be performed at every step, but only at the end of the loop, because we're not using it every time. This is important for efficiency, because softmax calculation is expensive. Let's move this calculation out of the inner loop. Also, the probabilities probably shouldn't be named `err`. What about something like `prob`?
* This is purely an aesthetic thing, but can we change `save2disk` to `save_to_disc`? "to" seems to go better with "from".

vicmak:
Changed according to review.
Thanks for the useful comments.

neubig:
Thanks! And sorry, I wasn't completely clear on the softmax. Lines 59-60 should also be outside of the loop, as they aren't used inside.

vicmak:
done

neubig:
Great, thanks!

PR #316, opened 2017-02-17
Removed unnecessary syncing primitives

farkhor:
The endpoint for a CUDA kernel is an implicit synchronization point for threads within the thread-block. Therefore, no '__syncthreads()' is needed.


neubig:
Great, thanks!

Issue #318, opened 2017-02-18
implementation of flip_gradient on dynet


iamyoungjo:
Hello.

This is more like proposal or suggestion than issue finding. I would like to implement flip_gradient, which is kind of a key feature for adversarial training. But dynet doesn't have it. I know this will never be a complicated tweak as it's just putting negative on backprop process only. But as I am lack of fundamental ideas of dynet structure, can you suggest which codes that I need to look into? I am trying to find the gradients in node unit.

I believe there could be many ways to make this work, but I need your suggestion.

Thanks,
YJ

neubig:
Sure, for this it'd probably be best to create a "flip_gradient" operation. A brief description of how to do so is here at the very end of the slides:
http://phontron.com/slides/emnlp2016-dynet-tutorial-part1.pdf

My suggestion would be to take a look at the code for the "nobackprop" operation and the unary "-" operation, and combine the ideas together. If you get this working we'd love to have a pull request!

iamyoungjo:
This tutorial is exactly what I was looking for. I will work on it and get back to you soon!

iamyoungjo:
Hi,

I implemented as you suggested. 

This is a snippet from nodes.cc
```c
// Flip gradient For adversarial training
template<class MyDevice>
void FlipGradient::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
  fx.d = xs[0]->d;
  fx.v = xs[0]->v;
}

template<class MyDevice>
void FlipGradient::backward_dev_impl(const MyDevice & dev,
                                   const vector<const Tensor*>& xs,
                                   const Tensor& fx,
                                   const Tensor& dEdf,
                                   unsigned i,
                                   Tensor& dEdxi) const {
  dEdxi.tvec().device(*dev.edevice) = dEdf.tvec().unaryExpr(scalar_negative_op<float>());
}
DYNET_NODE_INST_DEV_IMPL(FlipGradient)
```
And this is from simd_funtors.h. 
```c
namespace dynet {
template<typename Scalar> struct scalar_negative_op {
  EIGEN_EMPTY_STRUCT_CTOR(scalar_negative_op)
  DYNET_DEVICE_FUNC inline const Scalar operator() (const Scalar& x) const {
    return -x;
  }
  template <typename Packet>
  DYNET_DEVICE_FUNC inline Packet packetOp(const Packet& x) const {
    using namespace Eigen::internal;
    return pnegate(x);
  }
};
[...]

PR #321, opened 2017-02-19
adding flip_gradient operation for adversarial training

iamyoungjo:
This topic is discussed in #318 . For learnings in domain separated networks, you will need flip_gradient operation, GRL (Gradient Reversal Layer) in different term,  in many purpose. 

In order to implement it, I changed these files (incl. python bindings) as the advice from the thread #318 , but if there's any unnecessary or wrong changes, please give me a feedback. I checked with my networks and it worked as expected. 

Thanks,
YJ

### DIFF
--- a/dynet/expr.h
+++ b/dynet/expr.h
@@ -1092,7 +1092,7 @@ Expression nobackprop(const Expression& x);
  * \ingroup flowoperations
  * \brief Negative backprop 
  * \details This node has no effect on the forward pass, but takes negative on backprop process. 
- *          This is useful when you want to create a domain adaption network
+ *          This operation is widely used in adversarial networks.
  *          
  * \param x The input expression
  * 
--- a/dynet/expr.h
+++ b/dynet/expr.h
@@ -1090,6 +1090,18 @@ Expression nobackprop(const Expression& x);
 
 /**
  * \ingroup flowoperations
+ * \brief Negative backprop 
+ * \details This node has no effect on the forward pass, but takes negative on backprop process. 
+ *          This is useful when you want to create a domain adaption network
+ *          
+ * \param x The input expression
+ * 
+ * \return An output expression containing the same as input (only effects on backprop process)
+ */ 
+Expression flip_gradient(const Expression& x);  
+  
+/**
+ * \ingroup flowoperations
  * \brief Reshape to another size
  * \details This node reshapes a tensor to another size, without changing the
  *          underlying layout of the data. The layout of the data in DyNet is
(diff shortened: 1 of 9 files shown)


dynet/expr.h
```diff
@@ -1090,6 +1090,18 @@ Expression nobackprop(const Expression& x);
 
 /**
  * \ingroup flowoperations
+ * \brief Negative backprop 
+ * \details This node has no effect on the forward pass, but takes negative on backprop process. 
+ *          This is useful when you want to create a domain adaption network
```
neubig: I think "domain adaptation network" may be hard to understand here. What about saying something like "This operation is widely used in adversarial networks."

dynet/nodes.cc
```diff
@@ -1275,6 +1275,24 @@ void NoBackprop::backward_dev_impl(const MyDevice & dev,
 DYNET_NODE_INST_DEV_IMPL(NoBackprop)
 
 template<class MyDevice>
+void FlipGradient::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
+  fx.d = xs[0]->d;
```
neubig: There should be no need for this line, as dim_forward will be sure to set the dimensions of fx to the proper size.

dynet/nodes.cc
```diff
@@ -1275,6 +1275,24 @@ void NoBackprop::backward_dev_impl(const MyDevice & dev,
...
+
+template<class MyDevice>
+void FlipGradient::backward_dev_impl(const MyDevice & dev,
+                                   const vector<const Tensor*>& xs,
+                                   const Tensor& fx,
+                                   const Tensor& dEdf,
+                                   unsigned i,
+                                   Tensor& dEdxi) const {
+  // takes negative on backprop
+  dEdxi.tvec().device(*dev.edevice) = dEdf.tvec().unaryExpr(scalar_negative_op<float>());
```
neubig: There's one major issue here, and one way to potentially make things easier. The major issue is that `backward_dev_impl` must accumulate gradients, so instead of `=` we would need `+=`. Separately from this, we might be able to just use `-=` instead of using `unaryExpr`. In summary, how about changing this line to:

    dEdxi.tvec().device(*dev.edevice) -= dEdf.tvec()

dynet/simd-functors.h
```diff
@@ -71,6 +71,30 @@ struct functor_traits<dynet::const_minus_op<Scalar> > {
 } }
 
 namespace dynet {
+template<typename Scalar> struct scalar_negative_op {
```
neubig: If the above change to `nodes.cc` works, then this addition may not be necessary.

dynet/expr.h
iamyoungjo: That sounds more reasonable. I will update it as you suggested.

neubig:
Looks great, thanks!

PR #330, opened 2017-02-22
Turn on unit tests.

xunzhang:
Since CI only tests under CPU mode, all unit tests should pass. Under gpu environment, `test-nodes`, `test-params`, `test-trainers` should fail since some logics have been implemented yet.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,30 +1,43 @@
-find_package (Boost COMPONENTS system filesystem unit_test_framework REQUIRED)
-include_directories (${TEST_SOURCE_DIR}/src
-                     ${Boost_INCLUDE_DIRS}
-                     )
+#include(CTest)
+#enable_testing()
 
-add_definitions (-DBOOST_TEST_DYN_LINK)
+find_package(Boost COMPONENTS system filesystem unit_test_framework serialization REQUIRED)
+include_directories(${TEST_SOURCE_DIR}/src ${Boost_INCLUDE_DIRS})
+add_definitions(-DBOOST_TEST_DYN_LINK)
 
-# Sources:
-set(test_dynet_SRCS
-    test-nodes.cc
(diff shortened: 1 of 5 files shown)


neubig:
Thanks for this! I understand that you want to split up into multiple modules for use on Travis CI, but I often want to run all tests with a single executable on my own machine. Is there any way that we could make this so we satisfy both of these use cases?

xunzhang:
@neubig I am writing a pull request for solving part of https://github.com/clab/dynet/issues/84, collecting boost logic into `io-macros.h`. And I want to make sure there are no regressions failures after refactor. So I send this pull request first.

Good question, Let me think about it. Is that ok that we run with both cases and ignore that 3 failures under GPU mode currently?

xunzhang:
Or I can comment these three tests for unified test passing.

neubig:
@xunzhang I just realized that it's possible to run all tests by typing `make test`, so my original comment is resolved with no changes. (We might want to add something to the developer doc, just in case.) For your question, for now let's keep the tests failing just to maintain the status quo. One option would be to catch the exception and print a warning when running on GPU, but I don't think we need to do this now. I'm OK to merge this commit as-is if you agree.

xunzhang:
@neubig I add an `if-else` branch to make sure `make test` will pass in all modes. Please wait Travis turning green(there should be some latency for osx Travis machine) and then do merge. I will send another pull request handling dynet-84 this afternoon.

neubig:
@xunzhang Thanks, but turning off large segments of unit tests on GPU (particularly `test-nodes`) seems like a bad idea. I'd prefer that the tests that were failing continue to fail for the time being until we come up with a better solution like catching the errors that we know occur on GPU.

xunzhang:
@neubig Reverted back. Make sense, let's keep that in mind and add to my to-do list.

neubig:
Great, thanks a lot!

PR #331, opened 2017-02-22
dynet-84. Add macro interfaces for serialization, refactor code.

xunzhang:
This is the first step trying to resolve https://github.com/clab/dynet/issues/84: refactor serialization logic using Macro interfaces. In this case, we don't need to modify codes in a lot when importing other serialization tools and remove boost dependencies later.

I will write unit tests for demonstrating the interfaces' usage and for testing these interfaces in next pull request because it will take some time.


neubig:
Awesome, thanks!

PR #332, opened 2017-02-22
These device->host memory copies shouldn't be async.

mattr1:
I haven't been bitten by these lines, but I'm surprised no one else has had a problem with them. Unless I misunderstand the way cudamemcpy works, we should never do an asynchronous copy from device->host and then use the memory on the host immediately (without having done any synchronizing). I changed these to be synchronous calls to cudamemcpy since we use the memory immediately.


neubig:
Thanks, I think most were not async, but it looks like these might have been overlooked.

Issue #315, opened 2017-02-17
Immediate Computation needs Python Implementation


neubig:
Immediate computation (http://dynet.readthedocs.io/en/latest/debugging.html) is useful for debugging, but not implemented for Python. This can probably be added easily by adding optional arguments to the `renew_cg()` function.

neubig:
Fixed by #334

PR #334, opened 2017-02-24
Add immediate computation support for Python

Peilun-Li:
Add immediate computation support for Python as stated in #315 . I'm new to this project and hope the coding style is consistent and my understanding of the Python workflow is right. Please correct me if I'm wrong.


neubig:
Looks perfect, thanks a lot!

Issue #239, opened 2017-01-13
Weird dimension changing behavior


yoavartzi:
I noticed a weird behavior when adding two matrices. The dim() seems to be fine, as much as I understand how to read it, but the value() is a vector. The screenshot says it all. This is happening with the most recent version (just updated). 

![image](https://cloud.githubusercontent.com/assets/3289873/21941718/27d4a84c-d998-11e6-8f0d-0fbe57be503d.png)

yoavg:
Confirmed. I will look into it in the following days. Thanks!

yoavg:
This appears to be a C++ level thing. For some reason the ndims are set to 1 after the addition.

Code example:
```
  Expression x_ = parameter(cg,m.add_parameters({2,1}));
  cout << "x_.dim:" << x_.dim() << endl;
  cout << "(x_ + x_).dim:" << (x_ + x_).dim() << endl;
  Dim d = (x_ + x_).dim();
  cout << "(x_ + x_).nrows,ncols,ndims:" << d.rows() << " " << d.cols() << " " << d.ndims() << endl;
```

neubig:
The C++ version of the bug should be fixed by https://github.com/clab/dynet/commit/7f45b7e6b1dd85f510f7e6ce362dfd1e8942b52a

neubig:
Looks like the Python is still broken though. Any ideas @yoavg ?

pmichel31415:
I think I corrected this issue in #340 : 

Basically what `expr.dim()` did was returning (nb of elements,nb of rows, nb of cols, nb of batches) so here (2,2,1,1)

The new behavior in my pull request is returning (dim_1,dim_2,...,dim_n,batch_size) so this would correspond to (2,1,1) here.

Maybe it would be better not to include the batch size if it's equal to 1, what are your thoughts?

neubig:
I think it should probably return a tuple `((2,1), 1)`, which is probably what you did?

pmichel31415:
Aah that makes more sense, I'll do that.

neubig:
Fixed by #340?

PR #340, opened 2017-03-01
Add access to parameter gradient and direct numpy array + tests

pmichel31415:
Changelog : 

- `p.grad_as_array()` can now be called on a parameter or a lookupparameter to get the gradient computed by `backward`
- Add a `dy.inputTensor()` function : 

        Creates a tensor expression based on a numpy array or a list.
        The dimension is inferred from the shape of the input.
        if batched=True, the last dimension is used as abatch dimension

- Add rudimentary test files in `python_tests`

### DIFF
--- a/dynet/model.h
+++ b/dynet/model.h
@@ -107,7 +107,7 @@ struct ParameterStorage : public ParameterStorageBase {
 
   Dim dim; /**< Dimensions of the parameter tensor*/
   Tensor values;/**< Values of the parameter */
-  Tensor g;/*< Values of the gradient w.r.t. this parameter */
+  Tensor g;/**< Values of the gradient w.r.t. this parameter */
 
 private:
   ParameterStorage() {}
--- a/python/_dynet.pxd
+++ b/python/_dynet.pxd
@@ -34,6 +34,7 @@ cdef extern from "dynet/dim.h" namespace "dynet":
         int ndims()
         int rows()
         int cols()
+        unsigned operator[](unsigned i)
         void set(unsigned i, unsigned s)
         int size(unsigned i)
         CDim transpose()
@@ -58,6 +59,7 @@ cdef extern from "dynet/model.h" namespace "dynet":
         vector[CTensor] values
         vector[CTensor] grads
         CDim dim
+        CDim all_dim
 
     cdef cppclass CParameters "dynet::Parameter":
         CParameters()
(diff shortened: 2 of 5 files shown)


yoavg:
Didn't verify the details, but this looks good.

May I suggest changing the batch interface (or add both options?) so that a batch will be created from a list of ndarrays? So if an ndarrays is passed in, it's just a batch1 tensor, but if a list is passed, then it's a batch of the elements within it, each element shaped as the first?

pmichel31415:
OK I added a bunch of stuff, biggest changes are : 

- If the input of `inputTensor()` is a list of ndarrays, returns a batched expression acording to @yoavg 's suggestion

- Dimensions overhaul : for expressions, `e.dim()` now returns a tuple of all the dimensions + the batch dimension (previously returned (number of elements,number of rows, number of cols, batch dimension) ). This is for consistency with the c++ interface and also because it seemed weird to return the number of elements as part of the dimension. If you think this will break things please let me know and I will change it back.

- parameter.shape now returns the shape for any tensor of order > 2

- npvalue and value now support any shape

- tests for inputTensor were changed a little bit for tensors up to order 4 and assertions

neubig:
I'd like to have the option to specify the final dimension as the batch dimension as well, so would it be possible to have both as options? Specific reasons why we might want this is if we, for example, have a masking function input where we do

> mask = np.ones([10])
> mask[9] = 0
> mask_exp = dy.inputTensor(mask, batched=True)

I think this is probably easier and more efficient than creating a list of np objects. It's also probably more efficient if we're handling large batch sizes, as I guess it would require fewer memory copies.

neubig:
Ahh, and other changes look great.

pmichel31415:
I'm not entirely sure I get what you want. With the current version you can either : 

1. pass a list of np.ndarrays
2. pass a list of lists/tuples/anything that can be turned into a np.ndarray
3. pass a np.ndarray and optionally take the last dimension as the batch dimension 

From your comment it seems that you want option 3 but I didn't remove it, you can still use it like this (with batched=true)

neubig:
Oh, that's fine then!

pmichel31415:
OK I fixed the behaviour of the dim() method for expressions to solve #239

I think it's OK for me now

neubig:
LGTM!

PR #341, opened 2017-03-02
In-Memory File Read

vicmak:
Speed up disk IO while training

### DIFF
--- a/examples/python/util.py
+++ b/examples/python/util.py
@@ -1,5 +1,7 @@
 from collections import defaultdict
 from itertools import count
+import mmap
+
 class Vocab:
     def __init__(self, w2i=None):
         if w2i is None: w2i = defaultdict(count(0).next)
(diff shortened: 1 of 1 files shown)


neubig:
Thanks! This isn't used anywhere in the examples though. If we're going to add it, maybe it'd be nice to use it somewhere, or at least add a comment explaining it?

yoavg:
1) How much faster is this in practice?

2) From what I recall, the python mmap API differs in Windows vs Linux/OSX. Which one is this? should be documented.

vicmak:
1. When using large files ~ 2GB (on a VM) or more I think it worked a little faster, since the disk was the bottleneck. I did not measure it exactly.
2. @yoavg is right. This one is actually for windows. I will document the one for linux\OSx as well.

vicmak:
Even better - I set the linux API as the default.

neubig:
Looks good to me with the additional comment!

PR #344, opened 2017-03-02
Add sanity check in expressions to avoid memory violation

pmichel31415:
This fixes a bug in C++ where if you tried to access an expression value after the deletion of the underlying graph, you would get an error like `SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)` .

Highlights :

- This does more or less the same thing as python : keeps an id for each graph and checks whether the id of the graph with which the expression was created matches.
- Add a `name` field in expression for more explicit error messages
- Also added `except +` to forward functions in the computation graph in `_dynet.pyx`.

The last points means that now the following snippet :

    import dynet as dy
    x=dy.inputTensor(np.ones(10))
    y=dy.select_rows(x,[11])
    y.value()

raises a python `ValueError` instead of crashing python with `abort (core dumped)` and the c++ exception.

Now there is a concurrent sanity checking system for expressions in c++ and python, this could probably be fixed since python can now catch c++ exceptions so there's no need for double checking.

I'll leave that to a future pull request though.

### DIFF
--- a/dynet/devices.cc
+++ b/dynet/devices.cc
@@ -46,7 +46,7 @@ DeviceMempoolSizes::DeviceMempoolSizes(const std::string & descriptor) {
 Device::~Device() {}
 
 DeviceMempoolSizes Device::mark(ComputationGraph *cg) {
-  cg->incremental_forward({cg, (VariableIndex)(cg->nodes.size() - 1), "Mark expression"}); // needed so that we actually allocate the needed memory
+  cg->incremental_forward({cg, (VariableIndex)(cg->nodes.size() - 1)}); // needed so that we actually allocate the needed memory
   // for all existing nodes.
   return DeviceMempoolSizes(pools[0]->used, pools[1]->used, pools[2]->used);
 }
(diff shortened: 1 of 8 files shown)


neubig:
In general I approve, although there were a bunch of formatting changes included here so it was a bit difficult to check the actual changes. There's no significant speed hit due to doing this right? Even if there was it's probably still worth it, but just out of curiosity.

pmichel31415:
Ah yeah sorry the auto formatter is annoying.

So the only speed overhead added is that whenever you call `.value()` or `.dim()` on an expression it'll check the equality of two ints which is negligible. 

If the test fails it needs to create a stringstream for the error message but that should also be negligible.

Last but not least there is a memory overhead because each expression now stores a string.

neubig:
Hmm, I'd like to avoid adding a string to each expression if possible. These should be as light-weight as possible, hopefully with a fixed memory size. Would it be sufficient to call the `as_string()` function of the corresponding node?

pmichel31415:
OK I'll remove this altogether then.

you can't get any info on the node when raising this exception because when it occurs the computation graph (and hence the nodes) have been deleted so as_string is not an option.

I'll replace it with a generic message like in python

neubig:
Looks good!

Issue #147, opened 2016-11-07
Update specific parameters in SGD trainer


duyvuleo:
In SGD trainer, is it possible to update specific parameters (presumably their indices known in advance)??? 

For example:
vector<size_t> indices({1,2,5});
Trainer* trainer = SimpleSGDTrainer(&model);
trainer->update(indices);

Trainer only offers one update() function which will update all parameters once.

neubig:
This might be nice to have.

One way you can do this in the existing code is by doing the inverse: using the const_parameter and const_lookup expressions when inputting nodes to the computation graph will prevent them from being updated, so you can just call const_* for all but parameters 1,2,5. This will be a bit slower, but it should be sufficient for many cases.

duyvuleo:
Using const_parameter and const_lookup_parameter ones will be not comfortable in my situation.

Also, I tried to directly add the the following function into the SimpleSGDTrainer class. However, I always got the "Segmentation fault (core dumped)" error message in the update_rule() function.

void SimpleSGDTrainer::request_param_update(int p_index)
{  
  //gscale should be reused from update() step
  auto & p = model->parameters_list()[p_index];
  update_rule(scale, gscale, {&p->values, &p->g});
  p->clear();
}

Is there any way to fix this error?

Thanks!

neubig:
Fixed by #349. This will also be easier with the new ParameterCollection interface that we're currently developing in the new-readwrite branch.

PR #349, opened 2017-03-03
Make it possible to update a subset of parameters

neubig:
This is a first attempt at https://github.com/clab/dynet/issues/147

We still need to add unit tests, so this shouldn't be merged yet.


yoavg:
I know this is already merged, but perhaps the python interface can be changed to accept a single list of parameters (both lookup and non-lookup) and check the type internally?

neubig:
@yoavg Sure, I don't think anyone is using this yet, so we can go ahead and change.

Issue #352, opened 2017-03-05
Should have ability to read LookupParameters as matrix


neubig:
In some cases, we will want to use parameters like LookupParameters during training time, then as a simple matrix during test time (e.g. when implementing sampling-based approximations to a Softmax, such as NCE). This is not currently supported, but should be easy to add by allowing the `parameter()` expression to take LookupParameters as input as well.

yoavg:
+1 on having `parameter()` work also on LookupParameters.

Coming to think about it, what is the cost of allowing every 2dim parameter to have the LookupParameter functionality? Maybe we can just unify these two classes?

neubig:
I guess it would add an extra bool+vector to each parameter. The bool would be "whether this parameter has been used as a whole" and vector would be "if the parameter has been used as a lookup, which elements in this parameter have been used". The difficulty here though is that the intuitive layout of LookupParameters and Parameters is different because DyNet (Eigen) uses column-major order. I think it wouldn't be super-intuitive if we completely unified them.

yoavg:
Ok, so lets not.

PR #353, opened 2017-03-06
Code to allow expression for entirety of lookup params

neubig:
This code makes it possible to import the entirety of a lookup parameter matrix in the same way we do for standard parameters, addressing https://github.com/clab/dynet/issues/352

This is not yet tested completely, and not compiled on GPU, hence the WIP.


neubig:
Now tested to be working in C++ and Python, as well as on GPU. Should be basically ready to commit and will do so tomorrow if it looks good to others.

PR #357, opened 2017-03-06
Scala bindings for DyNet (via swig)

joelgrus:
We have created SWIG bindings so that we can use DyNet from Scala. They are pretty comprehensive, with lots of documentation and tests and examples, and we are actively using DyNet from Scala code.

Other than a few lines in the top level `CMakeLists`, all of our changes are under the new `swig` directory (and are hidden behind a flag which is OFF by default).

We wanted to contribute this back, as it seems like something that could be useful to a lot of people.

Incorporating this would require some sort of plan around keeping the bindings in sync with the root C++ code. Presumably that's already required for the Python bindings, so maybe it's not terribly hard.

Anyway, I know this isn't just a simple "LGTM" change, so let's discuss.


neubig:
LGTM!
Just kidding :) This is awesome though, I'll try to walk through this and see if we have any comments.

yoavg:
This is really cool! And am sure it will be useful for many. DL solutions for the JVMs are really lacking.

While I don't know Scala well at all, I am curious about two of the design decisions you took. Not saying they are wrong, they may very well be the best possible option, but I think it is still worth discussing.

1) The use of an explicit ComputationGraph object which is being passed around. This is indeed the way the C++ version does things, but the python version in contrast maintains it as a singleton which is being implicitly referred to when needed. You already depart from the C++ by maintaining a singleton (I think), so why not take it all the way? Not having to pass around the CG does make coding somewhat more pleasant.

2) The use of FloatVector, IntVector, ExpressionVector etc. What would be the tradeoffs in using these wrapper classes vs. using native scala lists?

joelgrus:
Regarding (1), one of our original design goals was "as close to the C++ API as possible". We were pretty far down that path when we realized that the C++ "let the computation graph go out of scope and get destructed" behavior didn't work in Scala. Initially we were going to force people to call `cg.delete()` manually, and then later we switched to the singleton approach.

On one hand, I agree that not having to pass the graph around would be nice; on the other hand, that would make the swig bindings more complex and would (in some sense) create a 3rd API. 

So the question would be: which is user-friendlier? Having to pass the cg around (but an API that looks like the C++ one, other than that one piece), or not having to pass it around (but having a different API). I'm not sure. Do other people feel strongly about one vs the other?

for (2), everything generated by SWIG is basically a Java wrapper around the C++, so the actual computations are being done on C++ `float[]` or `int[]`. Accordingly `FloatVector` and `IntVector` instances are in some sense just pointers to C++ arrays, so that it's basically costless for the native code to operate on them. 

There's no reason you couldn't make the API based on Java collections (I don't think SWIG knows anything about Scala collections), but I imagine you'd have to convert your `java.util.List` (or whatever) to a C++ vector any time you wanted DyNet to do something with it, so it would probably not be efficient. Also, I guess you'd have to let your C++ code modify your Java data structures, which I think ends up being kind of messy.

There are Scala helpers to construct e.g. a `FloatVector` from a `Seq[Float]`, so it's hopefully not too painful to use those.

yoavg:
Ok, I see. Being as close as possible to the C++ and having as minimal SWIG-wrapper as possible is definitely an ok criteria to go by.

jflanigan:
OMG this is so great!  Thank you!!

redpony:
I think having minimal SWIG is good, but idiomatic scala should be a
priority for an official Scala interface. Hiding implementation details and
emphasising the "functional programming" aspects of differentiable
algorithms seems like it might produce something that lets you use DyNet in
innovative ways, rather than just syntactic sugar that lets you use it in
the same old way from a new language.

that being said, this should be included in any case. Perhaps we should
create a contrib/ directory for things like this?

yoavg:
+1 on a contrib/ directory

jayantk:
FWIW I think these SWIG bindings are a step toward any more "Scala-y" interface. It's much easier to wrap the existing C++ functions with SWIG, then write the Scala-y interface directly in Scala, than it is to use SWIG to create that interface directly.

joelgrus:
yeah, I have actually spent the day hacking a first draft of a Scala-y interface, entirely on the scala side (I haven't touched the swig bindings at all), it will probably be working tomorrow, I'll share it when it's done.

joelgrus:
OK, I made a first attempt at a more scala-like interface. It's not complete (and not documented), but you can see it here:

https://github.com/allenai/dynet/tree/scaly/swig/src/main/scala/edu/cmu/dynet

At a high level, the changes are

* moved the SWIG bindings into `edu.cmu.dynet.internal` (but otherwise left them untouched)
* created a bunch of Scala wrapper classes, each of which wraps a package-private instance of the "internal" (SWIG-generated) Java class, and has methods that delegate to that instance. Basically, no one who uses this API should ever see/know about any of the SWIG-generated Java classes
* gave everything scala-like names
* simplified the `_Vector` wrappers and made them implement `IndexedSeq` so that they mostly "feel like" natural scala collections
* replaced the `dim()` helper function with factory methods in the `Dim` object
* put the singleton `ComputationGraph` in the `ComputationGraph` object, and moved all its methods to methods on that object
* accordingly, removed `ComputationGraph` from all function signatures
* versioned `Expression`s (as in the python code)
* instead of wrapping all the `ParameterInit` subclasses, I just made their "constructors" factory methods on the `ParameterInit` companion object
* all of the bare `Expression`-generating functions are now members of the `Expression` companion object

I updated all the examples and most of the tests. Probably just by looking at the examples you can get a good flavor of how it would work.

--

This version should be much more idiomatic Scala (at the expense of having two levels of wrapper code), but I don't know that it "emphasizes the functional programming aspects", can you say more about what you mean by that?

Anyway, please let me know what you think of this approach, or if you see any changes that would make it nicer to use.

redpony:
On Wed, Mar 8, 2017 at 7:23 PM, Joel Grus <notifications@github.com> wrote:

> OK, I made a first attempt at a more scala-like interface. It's not
> complete (and not documented), but you can see it here:
>
> https://github.com/allenai/dynet/tree/scaly/swig/src/
> main/scala/edu/cmu/dynet
>
> At a high level, the changes are
>
>    - moved the SWIG bindings into edu.cmu.dynet.internal (but otherwise
>    left them untouched)
>    - created a bunch of Scala wrapper classes, each of which wraps a
>    package-private instance of the "internal" (SWIG-generated) Java class, and
>    has methods that delegate to that instance. Basically, no one who uses this
>    API should ever see/know about any of the SWIG-generated Java classes
>    - gave everything scala-like names
>    - simplified the _Vector wrappers and made them implement IndexedSeq
>    so that they mostly "feel like" natural scala collections
>    - replaced the dim() helper function with factory methods in the Dim
>    object
>    - put the singleton ComputationGraph in the ComputationGraph object,
>    and moved all its methods to methods on that object
>    - accordingly, removed ComputationGraph from all function signatures
>    - versioned Expressions (as in the python code)
>    - instead of wrapping all the ParameterInit subclasses, I just made
>    their "constructors" factory methods on the ParameterInit companion
>    object
>    - all of the bare Expression-generating functions are now members of
>    the Expression companion object
>
> I updated all the examples and most of the tests. Probably just by looking
> at the examples you can get a good flavor of how it would work.
>
> --
>
> This version should be much more idiomatic Scala (at the expense of having
> two levels of wrapper code), but I don't know that it "emphasizes the
> functional programming aspects", can you say more about what you mean by
> that?
>
Here is one example of what I meant. You can think of RNNBuilders as
executing a recursive function on a sequence of input Expressions. Yes, we
wrote a C++ function that does this for some standard types of recurrent
networks, but if I had been writing DyNet from scratch in Scala (or Haskell
or whatever), the builder interface would have looked entirely different
than the one we have in C++ since Scala does this sort of thing in much
more beautiful ways than C++ ever could.

Here's the recursive definition of an RNN: the base case returns h_0 (a
learned expression, or 0); then recursively you apply tanh(W*[h_{t-1}; x_t]
+ b). Are you encoding a sequence and just want the final encoding? Well,
then just execute this as a foldLeft which creates the intermediate values
but doesn't return them to the user; only h_t is returned. Are you going to
attend to them later (e.g., in an attention based model)? Then use
scanLeft. Both of these take the same function which is the definition of
the RNN cell. It's a simple matter to change the definition of the RNN cell
from a simple RNN to a GRU. The code is factored beautifully, and there is
no new API with funny names to learn, just standard Scala functions like
fold and scan. What are the inputs to the RNN? Well, anything that
generates a list.

Why didn't I push for this in Python which could have also had a more
Pythonic interface? Well, Python can be a little slow and so we wanted a
fast interface, but Scala is quite fast, so I would be more than happy to
see the C++-accented builder interfaces disappear from Scala in favor of
native implementations. Of course, I realize that rewriting definitions of
RNNs and LSTMs is more work for you, and we've got nice and tested
definitions in the C++ code, so having both around, but I think a
functional language with idiomatic definitions of RNNs (and Recursive NNs,
etc etc) would be really cool. Most of the computational work, at the end
of the day, will be done by the C++ backend when it gets down to number
crunching, but you will have new and exciting ways to create differentiable
computations.

Does this make sense?

jayantk:
I don't think we have to make one monolithic change that implements the whole Scala API in one shot. The functionality we have in these bindings is useful as is, and is necessary for the more functional API that we're talking about. Let's not let the great be the enemy of the good.

That said, if we're just talking about RNN builders, we at AI2 can do that in this PR. Should be easy to do by copy-pasting the recurrence definitions from the C++ into scala with some appropriate wrapping.

redpony:
Yes, I agree-- I think the current interface is fine and isn't going to
cause any trouble (all the primitives you will need to write something more
interesting are in place and look good). I just wanted to make sure we
didn't create something that would bleed C++ness out into the calling
code...

neubig:
I'm interested in going ahead and merging this sooner rather than later to prevent the branches from diverging. I think it's important to make sure that we have some semblance of maintenance of the scala bindings if we incorporate them though. @joelgrus , would you mind if we pass issues on to you or mail occasionally when we do API updates that need to be incorporated?

joelgrus:
yes, contacting me for issues / updates sounds good. I would like to add one more README section on "how to add/update Scala bindings" and I would also like to make sure that the latest upstream changes don't break anything (they shouldn't but you never know). how about I do both on Monday morning and then we merge after that?

(Jayant and I are thinking about the "functional rnn" piece, but I don't think we'll have anything production-ready in the very short term, so best not to wait for that)

neubig:
@joelgrus Sounds good! Looking forward to it.

joelgrus:
ok, I pulled in the latest changes and everything still works, so it's ready for merging whenever you are.

neubig:
Great, thanks a bunch for doing this! I'll merge the pull request for now. In the future we should also probably mention this somewhere in the main doc as well. Tests via travis might also be nice so we at least know when we broke something in the bindings.

PR #358, opened 2017-03-06
Steps towards unifying error catching/reporting

neubig:
One of the most annoying things about DyNet at the moment is when you get a random and opaque C++ assert error. This pull request takes steps towards unifying error catching and reporting, and provides `DYNET_INVALID_ARG()` and `DYNET_RUNTIME_ERR()` macros that allow the easy throwing of formatted exceptions.

It is still in progress. When it is finished all throws and most asserts will be replaced by these macros.

### DIFF
--- a/doc/source/code_style.rst
+++ b/doc/source/code_style.rst
@@ -48,7 +48,16 @@ computation graph construction.
   or small enough (memory-wise) or in other ways cheap enough (time-wise) to
   copy.
 
-**Throwing Exceptions:** When the user does something illegal, throw an
-exception. "assert" should never be used for something that might be triggered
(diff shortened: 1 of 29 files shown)


neubig:
This pull request now gets rid of all asserts, replacing them with DYNET_ASSERT, which throws a catchable runtime error. It also replaces many of the direct throws of errors with DYNET_INVALID_ARG or DYNET_RUNTIME_ERR, which simplify formatting. It hasn't gotten rid of all direct throws yet, but this is lower priority (it will simplify code, but not change behavior), so I'll leave it to another pull request. I'll merge this later today if there are no objections.

PR #360, opened 2017-03-07
Fixed string encoding bug for python 3

### LINKED ISSUE #359: Error when trying to set memory amount in Python
Hi,

I'm trying to change the amount of memory that DyNet uses from within Python code, using the interface recently added (where you instantiate a `DyParams` object and then set parameters with it). I can create the `DyParams` object fine, but when I try to use it to set the memory, this happens:

```
>>> dyparams.set_mem(2048)
  File "_dynet.pyx", line 65, in _dynet.DynetParams.set_mem (_dynet.cpp:5070)
    cpdef set_mem(self, unsigned mem):
  File "_dynet.pyx", line 66, in _dynet.DynetParams.set_mem (_dynet.cpp:5011)
    self.cparams.mem_descriptor = str(mem)
  File "stringsource", line 15, in string.from_py.__pyx_convert_string_from_py_std__in_string (_dynet.cpp:53660)
TypeError: expected bytes, str found
```

All of the other parameter setters described in the documentation work fine, it seems to only be `set_mem` that has an issue. My suspicion is that this may be a Python 3 issue.

I'm using Python 3.5.1 on macOS 10.12.3 with a DyNet install built from source pulled from the GitHub repository today.

pmichel31415:
Fixes #359 

Tested for python 2.7 and 3.4

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -63,7 +63,7 @@ cdef class DynetParams:
         dynet.initialize(self.cparams)
 
     cpdef set_mem(self, unsigned mem):
-        self.cparams.mem_descriptor = str(mem)
+        self.cparams.mem_descriptor = str(mem).encode()
 
     cpdef set_random_seed(self, unsigned random_seed):
         self.cparams.random_seed = random_seed


pmichel31415:
OK I will merge this since it's a small change and it's been tested

PR #364, opened 2017-03-09
Dynamically increasing memory pools

neubig:
This pull request will allow the size of memory pools to dynamically increase as more memory is required. This means no more "dynet out of memory" errors, and no more manually adjusting the size of the forward, backward, and parameter memory pools to make sure your program doesn't die before training.

This will be ready to commit pending more complete stress tests on CPU and GPU, and hopefully a unit test or two, including of the revert functionality.


chaitanyamalaviya:
Running out of forward memory due to a batch of longer sentences is a recurrent problem that I have been facing while training models and this would be a great help. Hopefully this could be merged soon pending the unit tests.

neubig:
@yoavg OK, I tried to make two changes to this and am pretty much satisfied.

1. Tests are added for the basic case, and seem to be working.

2. I've disabled the combination of checkpointing + dynamic memory allocation. The reason for this is because the current implementation will result in the number of pools increasing every time checkpointing rolls back to a previous InternalMemoryPool, and the only good way I could think about fixing this is freeing the memory pools every time we roll back, which seems quite costly in the standard use-case for checkpointing (where we calculate, then throw out results repeatedly during search). I think that it's OK to have this off because a) checkpointing is an advanced enough option that people using it can probably also specify `--dynet-mem` correctly, and b) I have some plans about how to do memory optimization in a better way than checkpointing, which hopefully will be coming in the near future.

If this looks good, maybe I can do a few more stress tests. ( @chaitanyamalaviya , would you like to be a guinea pig? :) )

chaitanyamalaviya:
Sounds good! Let me test this and I'll get back to you.

yoavg:
I was worried mostly about the check pointing case, which I really just hacked up to make things compile, without much thought. If we can live without it, and there are basic tests in place, can merge it in I guess.

neubig:
@chaitanyamalaviya Is this working out for you? I think this is basically ready to merge.

anglil:
@chaitanyamalaviya how long of a sentence do you usually count as one that's too long? This same problem is also recurrent in my case and I'm trying to set a good threshold to throw out longer sentences.

chaitanyamalaviya:
I usually put a threshold of 80 words. You can also try a smaller batch size to get around this. And also make sure you are renewing the computation graph in every iteration.

anglil:
@chaitanyamalaviya Good to know -  thanks!

PR #365, opened 2017-03-09
Pass exceptions from operations to python

pmichel31415:
Exceptions launched when building the comptuation graph (e.g invalid_argument in Node::dim_forward) Are now passed to python so that they can be handled in python.

Examples : 

Before : ipython inerpreter crashes : 

screenshot from 2017-03-09 12-11-57

After : Get a ValueError, the interpreter doesn't crash : 

screenshot from 2017-03-09 12-11-10


neubig:
Excellent! LGTM.

Issue #260, opened 2017-01-19
Need Python Batching Example


neubig:
This should be added to the official DyNet examples:
https://github.com/neulab/dynet-benchmark/blob/master/dynet-py/rnnlm-batch.py

neubig:
(The main sticking point now is that the code in the above link is totally different from the `rnnlm.py` example in the examples directory now. They should be as similar as possible.)

neubig:
Fixed by #368

PR #368, opened 2017-03-10
Minibatch variant of rnnlm example

ratishsp:
Created minbatch example for rnnlm following the example at https://github.com/neulab/dynet-benchmark/blob/master/dynet-py/rnnlm-batch.py
This addresses issue #260

### DIFF
--- a/examples/python/rnnlm-batch.py
+++ b/examples/python/rnnlm-batch.py
@@ -1,0 +1,143 @@
+from dynet import *
+import time
+import random
+
+LAYERS = 2
+INPUT_DIM = 256 #50  #256
+HIDDEN_DIM = 256 # 50  #1024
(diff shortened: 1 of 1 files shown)


neubig:
Thanks a bunch for this @ratishsp ! It basically looks good to me and will merge a bit later if there are no other comments from the Python side.

vicmak:
what is the difference from [https://github.com/clab/dynet_tutorial_examples/blob/master/tutorial_rnnlm_minibatch.py](url) on dynet tutorial example?

neubig:
This example will be incorporated into DyNet itself, which I think is important, even if there's little difference from the tutorial examples that we did for EMNLP. Also, it looks like this is closer to the C++ example than the other ones we have, which is good for people who want to draw parallels.

Issue #362, opened 2017-03-08
Allow loading constant parameters in Python API


jeanm:
It would be nice to have the equivalent of `dynet::expr::const_parameter()` in the Python interface, perhaps via something like `dynet.parameter(something, update=False)` to match `dynet.lookup()`.

pmichel31415:
Solved by #371

PR #371, opened 2017-03-11
Allow const parameters in python

pmichel31415:
Tested on python 2.7

I will try to write actual tests tomorrow

Addresses #362

### DIFF
--- a/python/_dynet.pxd
+++ b/python/_dynet.pxd
@@ -219,6 +219,8 @@ cdef extern from "dynet/expr.h" namespace "dynet::expr":
     CExpression c_input "dynet::expr::input" (CComputationGraph& g, CDim& d, vector[float]* pdata) except +
     CExpression c_parameter "dynet::expr::parameter" (CComputationGraph& g, CParameters p) except + #
     CExpression c_parameter "dynet::expr::parameter" (CComputationGraph& g, CLookupParameters p) except + #
+    CExpression c_const_parameter "dynet::expr::const_parameter" (CComputationGraph& g, CParameters p) except + #
+    CExpression c_const_parameter "dynet::expr::const_parameter" (CComputationGraph& g, CLookupParameters p) except + #
     #CExpression c_lookup "dynet::expr::lookup" (CComputationGraph& g, CLookupParameters* p, unsigned index) except +   #
     CExpression c_lookup "dynet::expr::lookup" (CComputationGraph& g, CLookupParameters p, unsigned* pindex) except + #
     CExpression c_lookup "dynet::expr::lookup" (CComputationGraph& g, CLookupParameters p, vector[unsigned]* pindices) except + #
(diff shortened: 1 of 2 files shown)


yoavg:
This looks a bit buggy in the sense that you cannot add the same parameter to a computation graph once as fixed and once as non-fixed (because, importantly, expressions for the same parameter are reused), although the API suggests that you can.

I suggest adding a separate const_expr() method, with a diff check than version (or with no check at all), and have the dy.parameter() method call the correct one based on the flag.

pmichel31415:
Oh good point.

Another alternative would be to add a separate `const_version` and `const_expr` field and treat them separately in the `expr` method, how would that compare in terms of interface/performance?

yoavg:
This is also an option. I would check the speed overhead of `expr(bool)` vs two different methods (`expr()`, `const_expr()`) though. Or at least do `cython -a` and see that the `expr()` code remains white.

pmichel31415:
Hi @yoavg I modified the expr function accordingly.

I tried running cython -a but I still haven't found how to make it work and I'm not an expert dynet's build pipleine, can you point me to the command to run ? I tried : 

    cython -a pybridge.h _dynet.pxd _dynet.pyx dynet.py

And variants thereof to no avail.

Thanks!

yoavg:
` cython -a --cplus _dynet.pyx` from within the build/python folder should do the trick (will produce a `_dynet.html` file)

pmichel31415:
Ah thanks the `--cplus` flag did the trick.

So FYI here are the results : 

Old expr (current master) : 
old_expr

New expr (current branch) : 
new_expr

neubig:
Is this ready to merge? It looked good according to my brief perusal, and would be nice to merge before other updates to parameter objects (e.g. in the WIP in https://github.com/clab/dynet/pull/384)

jcyk:
Hope this can be merged soon. Really need it.

neubig:
@pmichel31415 I was taking a look at this one more time and it seems that there is a disconnect between parameters and doc (parameters are "update", doc is "const"), and also that the defaults are different ("update=False" once, "update=True" twice). Mind taking one more look?

pmichel31415:
OK this is fixed thanks. cython analysis looks OK to you I think this is ready to be merged

neubig:
It looks like the `update=False` and `update=True` inconsistency is still there. Is this by design, or just a typo?

pmichel31415:
Sorry about that this should be fine now

neubig:
OK, looks good!

yoavg:
sorry, another question on this: I did not follow the code closely at all, but am curious about the semantics of const here --- does it mean "not affected by backprop (but is affected by regularization)" or "does not change at all no matter what"? whichever is the answer, I think it would be good to be explicit about this in the docs (also for the "lookup(e,i, update=False)" case)

PR #372, opened 2017-03-11
Add python to travis

pmichel31415:
Only 2.7 for now.

This is done via conda, see https://conda.io/docs/travis.html#the-travis-yml-file for reference

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,24 @@ matrix:
   include:
     - os: linux
       compiler: gcc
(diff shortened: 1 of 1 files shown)


pmichel31415:
OK I added python 3.

This is a bit ugly (lots of copy-pasting in .travis.yml) but I don't think there's a better solution if we're using matrix.include

As far as I'm concerned this is it for this pull request, the next step is to add systematic tests for python but that's for another time.

neubig:
Awesome! This is great for now.

PR #375, opened 2017-03-13
configurable error handling

Oneplus:
Add a new compile time flag `DYNET_DEBUG_LEVEL`. When
* `DYNET_DEBUG_LEVEL=0`, no error assertion is adopted.
* `DYNET_DEBUG_LEVEL=1`, activate runtime checking on the computation graph but disable development assertion
* `DYNET_DEBUG_LEVEL=2`, do all the assertion.
This flag is 1 by default.

NOTE:
Currently, the boundary between graph checking and development assertion is a little bit vague. So this PR didn't change the implementation of different checking but only wrap the code like
```
if (wrong_cond)
  DYNET_INVALID_ARG(msg)
```
with a `DYNET_INVALID_ARG_CHECK(wrong_cond, msg)` macro.
BTW, error checking is a little bit messy. I've seen the code of checking whether a file is correctly opened, raises an invalid_argument exception. Semantically, a runtime_error exception may be more suitable. But this kind of improvements are not involved in this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,11 @@ endif()
 # echo "export BOOST_ROOT=/cygdrive/d/tools/boost_1_58_0/boost_1_58_0" >> ~/.bashrc
 # then run source ~/.bashrc to have those environment variable effective immediately
 
+if (NOT DEFINED DYNET_DEBUG_LEVEL)
+  set(DYNET_DEBUG_LEVEL 1)
+endif()
+add_definitions(-DDYNET_DEBUG_LEVEL=${DYNET_DEBUG_LEVEL})
+
 if(BACKEND)
   message("-- BACKEND: ${BACKEND}")
 else()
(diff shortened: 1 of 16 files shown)


neubig:
Thank you! This is awesome. Just one concern though: currently DYNET_INVALID_ARG_CHECK will throw if the condition is true, while DYNET_ASSERT will throw if the condition is false. This seems to be pretty confusing. Perhaps we need to change DYNET_ASSERT to DYNET_INTERNAL_CHECK and have both check for true values, or change DYNET_INVALID_ARG_CHECK to throw when the condition is false. What do you think?

Oneplus:
Hi @neubig , what about `DYNET_INVALID_ARG_CHECK` throw when the condition is false?

neubig:
@Oneplus Yeah, I think that's fine.

Oneplus:
@neubig It's done. If any further suggestion, please advise.

neubig:
Looks good! With the new semantics of INVALID_ARG_CHECK, I think it's probably better to rename this ARG_CHECK, so you check if the arguments are correct, and if they are not throw an exception. I'll go ahead and do this myself after merging though.

PR #380, opened 2017-03-14
expression functions with varargs

joelgrus:
Added versions of all the `Expression.f(ev: ExpressionVector)` functions that take `Expression*` args. 

This requires creating potentially ephemeral `ExpressionVector` instances. as we discovered before, this can lead to problems when the garbage collector runs and deletes the underlying `std::vector`. I couldn't actually produce a segfault this way, but I think it's possible. 

Accordingly, I gave each `Expression` instance a `Seq[AnyRef]` of things that shouldn't be garbage collected before it is, and I modified all of the `Expression`-generating functions to set them.

let me know if you don't like this approach, but I think it's the right defensive thing to do.

(I also added some `debugString` methods, since outside of this package -- and in particular in PNP -- you can't see the internal state of DyNet objects, which makes it hard to figure out why something is not working right. I'll probably keep adding these as needed.)


joelgrus:
oops, this was supposed to be in my own fork :(

PR #382, opened 2017-03-16
[WIP] dynet-92. Multi-device support.

xunzhang:
- [x] Refactor HAVE_CUDA usage for multi-device
- [ ] ~~Currently, parameters are all declared on the default device. There needs to be an option to let parameters live on other devices~~
- [ ] ~~We need to create a node that moves memory between devices~~
- [ ] ~~Test code for multi-device support needs to be created~~

### DIFF
--- a/dynet/tensor.cc
+++ b/dynet/tensor.cc
@@ -18,62 +18,64 @@ using namespace std;
 namespace dynet {
 
 ostream& operator<<(ostream& os, const Tensor& t) {
-#if HAVE_CUDA
-  if (t.device->type == DeviceType::GPU) {
-    vector<real> vt = as_vector(t);
-    Eigen::Map<Eigen::MatrixXf> m(&vt[0], t.d.rows(), t.d.cols());
-    os << m;
-  } else {
-#endif
+  if (t.device->type == DeviceType::CPU) {
     os << (*t);
+  } else {
 #if HAVE_CUDA
-  }
+    if (t.device->type == DeviceType::GPU) {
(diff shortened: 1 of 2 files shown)


neubig:
Is the change in f9ac2e2 necessary? I think the logic is exactly the same, but the current code avoids the "if" statement if CUDA is not enabled.

xunzhang:
@neubig Yes, the logic is the same and the previous implementation avoids the `if` statement in CPU mode. The reason why I did this first is that it's easier to add more devices and more readable I think. For example, 
```
void foo(const Tensor &t) {
  if (t.device->type == DeviceType::CPU) {
    // cpu
  } else {
#if HAVE_CUDA
  if (t.device->type == DeviceType::GPU) {
    // gpu
  }
#endif
#if HAVE_ANDROID
  if (t.device->type == DeviceType::ANDROID) {
    // android
  }
#endif
  }
} // foo
```

If you don't think it is necessary, or you don't like `if`s here, I can definitely make further changes based on the old implementation.

neubig:
Ahh, fair enough. I think it's fine to continue as-is then!

rasoolims:
When searching for Android, I came across this issue. Just curious; is dynet installable on Android or ios?

xunzhang:
@rasoolims, thanks! We are currently working on supporting both GPU and CPU which is the first step of actual multi-device support and after that Android and other devices are in the plan. Hope dynet users could run dynet on Android and other devices in the near future.

PR #383, opened 2017-03-17
Wrong path in TL;DR section

peteruhrig:
Was:
``cmake .. -DEIGEN3_INCLUDE_DIR=../eigen -DPYTHON=`which python` ``
At least on my machine, it had to be
``cmake .. -DEIGEN3_INCLUDE_DIR=../../eigen -DPYTHON=`which python` ``


neubig:
Thank you!

PR #384, opened 2017-03-17
Implement for new ParameterCollection interface

xunzhang:
- [x] Implement get_fullname and get_namespace interfaces, add related tests.
- [x] Add interface for supporting m.add_parameters({10}, "a") and tests.
- [x] Add input parameter and add_subcollection verify.
- [x] Handle parameter copy logic: currently code treat copy as sharing objects with original ones while we might disable copy in the future.
- [x] Parameter retrieval interfaces implementation. By model, by name, etc.
- [x] Builder classes implementation and tests.

### DIFF
--- a/dynet/cfsm-builder.h
+++ b/dynet/cfsm-builder.h
@@ -38,7 +38,12 @@ public:
   expr::Expression neg_log_softmax(const expr::Expression& rep, unsigned wordidx);
   unsigned sample(const expr::Expression& rep);
   expr::Expression full_log_distribution(const expr::Expression& rep);
-
+  std::vector<ParameterStorage*> get_parameters() {
+    std::vector<ParameterStorage*> rl;
+    rl.push_back(&p_w.get_storage());
+    rl.push_back(&p_b.get_storage());
+    return rl;
+  }
 private:
   StandardSoftmaxBuilder();
   Parameter p_w;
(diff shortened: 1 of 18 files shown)


xunzhang:
Will file remaining sub-tasks in following pull request.

- Parameter/lookupParameters level Serializer implementation, load/save functions.
- Python interface wrapper and tests.
- Update tutorial

dynet/cfsm-builder.h
```diff
@@ -38,7 +38,12 @@ class StandardSoftmaxBuilder : public SoftmaxBuilder {
   expr::Expression neg_log_softmax(const expr::Expression& rep, unsigned wordidx);
   unsigned sample(const expr::Expression& rep);
   expr::Expression full_log_distribution(const expr::Expression& rep);
-
+  std::vector<ParameterStorage*> get_parameters() {
```
neubig: In this and the other classes in the commit, wouldn't be easier to save `local_model` as a member of the class, then when `get_parameters()` is called, call `local_model.get_parameters()`? The ability to do this simply is one of the major advantages of the `ParameterCollection` concept I think.

dynet/model.cc
```diff
@@ -85,6 +83,11 @@ DYNET_SERIALIZE_COMMIT(ParameterStorage,
 DYNET_SERIALIZE_IMPL(ParameterStorage)
 #endif
 
+bool validParameter(const std::string & s) {
```
neubig: This should be in snake case.

dynet/cfsm-builder.h
xunzhang: Agree!

dynet/model.cc
xunzhang: fixed, thanks

neubig:
Thanks! Looks good.

Issue #73, opened 2016-09-08
Question: how to get averaged parameters?


rasoolims:
Hi,

Thanks for the great library.

I was wondering if this library support averaging? I mean, maintaining an averaged parameter that is going to be used in decoding. 

Thanks

neubig:
Hi @rasoolims . I think this feature might be something nice to have, as there have been some reports that this helps stabilize accuracy in a manner similar to ensembling. This could conceivably be done by adding capability to create copies of models, or add models together (with some weighting factor). However, this isn't implemented at the moment. If someone is interested in implementing it we'd be happy to accept a pull request.

honnibal:
I've found averaging really useful. I've not run nearly as many neural network experiments as you guys, but what I've seen so far is that averaging removes the need for the really expensive part of the training — squeezing out that final 0.5% of accuracy.

It can be implemented very easily by modifying the optimizer. You just add another shadow parameter. I'll send a pull request if I have time, but in case I don't here are some quick notes if the next person needs this and is looking at this issue.

* The file to change is training.cc
* Look at the way the Adam, Adadelta etc solvers track shadow parameters. You need to add one just like that.
* You can either track an exponential moving average, or just sum the weights and divide at the end. 
* The Google guys have some magic numbers in their SyntaxNet research for this that seem to work very well. I found the description in their paper hard to follow and had to look it up in Tensorflow. It works out like this (apologies for the Cython):

```python

cdef void update_averages(float* ema,
        const float* weights, int nr_weight, float nr_update) nogil:
    decay = (1.0 + nr_update) / (10.0 + nr_update)
    if decay > 0.9999:
        decay = 0.9999
    for i in range(nr_weight):
        ema[i] -= (1-decay) * (ema[i] - weights[i])
```

neubig:
We would certainly really appreciate a pull request for this, where averaging is an option that we pass to each optimizer.

PR #385, opened 2017-03-18
add support for average parameter

Peilun-Li:
Add support for average parameter as in #73 .
Approach: 
1. (model.h and model.cc) add a "maintain_average" parameter (with default value false) for "add_parameters" functions, and "get_average_parameters" functions to get a parameter's corresponding average parameter.
2. (training.h and training.cc) add a "emadecay" parameter (with default value 0.9999) for all trainers. When trainer.update() is called, for each updated parameter, update its corresponding average parameter if exists.
3. the average parameter maintains a "Exponential Moving Average", i.e., parameter = parameter * decay + new_value * (1 - decay). 

Example Usage:
C++:
`Parameter p_W, p_W_avg;`
`p_W = m.add_parameters({HIDDEN_SIZE, 2}, 0.0, "", true);`
`p_W_avg = p_W.get_average_parameter();`
`float ema_decay = 0.99;`
`SimpleSGDTrainer sgd(m, 0.1, 0.0, ema_decay);`
Python:
`sgd = SimpleSGDTrainer(m, emadecay=0.9)`
`pa = m.add_parameters(1, maintain_average=True)`
`pa_avg = pa.get_average_parameter()`
`lookup = m.add_lookup_parameters((VOCAB_SIZE, INPUT_DIM), maintain_average=True)`
`lookup_avg = lookup.get_average_lookup_parameter()`

Also have tested briefly on both C++ and Python, and both CPU and GPU.


neubig:
First, thanks so much for this! This is great. I do have a few questions about design decisions though:
1. Wouldn't it be easier to just include a pointer to a Tensor for the averaged parameters within the ParameterStorage and LookupParameterStorage objects? The pointer can be null if the parameters are not to be averaged, and non-null if we want to maintain the average.
2. This it'd be nice if we could remove the duplicate calls to `update_average()` and just have everything within one function call.

Peilun-Li:
While my thought and understanding is that it would be convenient for users to get average parameters as real "Parameter" objects, so that they can rebuild computation graphs easily using average parameters just like original parameters (which is also TensorFlow's approach). In this case it would be convenient to store average parameters as separate ParameterStorage since the "Parameter" struct is defined as a index of model's ParameterStorage list. Another advantage of this design is that the average parameters can benefit from existed and future functions of Parameter and ParameterStorage.

And for the second part, I develop `update_average()` related functions following the design of `update_params()` and `update_rule() `of trainers. Are there more concise ways to do the average update considering different update rules for Parameter/LookupParameter and different CPU/GPU devices? Thanks!

neubig:
Thanks for the reply, and again for adding the functionality! The design decisions make sense now, but I haven't had time to do a detailed review yet. I'll try to get to it ASAP.

neubig:
Hi @Peilun-Li , I took another look at this. I think the current implementation is good, but it bugs me a little bit that both normal and averaged parameters are included in `params` and `all_params`. I understand that in the current implementation it is necessary to do this because `Parameter` objects are represented as an index into `vector<ParameterStorage*>`, so unless the average parameters are included in this vector they can't be referenced. Fortunately though, there is a new  branch here that removes this restriction, referencing the `ParameterStorage*` directly: https://github.com/clab/dynet/tree/new-readwrite/dynet

Would you mind taking a look at this new branch, and see if there's a way that we could implement this pull request in a way so that we don't need to keep the averaged parameters in the same vector as the normal parameters? Thanks a lot in advance!

Peilun-Li:
Hi @neubig, I make some changes of `Parameter::get()`allowing Parameter to be an index of `vector<ParameterStorage*>` different from that saves normal parameters, judging by a bool value `is_average_param`. Then it's convenient to save average parameters in separate vectors and users can still get them easily with the same api. Would these changes satisfy the requirement? Thanks!

rasoolims:
@Peilun-Li How about RNN builders? Is n't is better to have a general averaging option for the model instead? I think when one applies averaging, it covers all updatable parameters; so having redundant ```maintain_average=True``` for each parameter is annoying.

neubig:
@rasoolims This will be easier to support once we create the new read/write interface, which will make it easier to access parameters in builders. For now I think we can stick with the current interface that @Peilun-Li had made.

rasoolims:
So how can I set the model to average ```BiRNNBuilder``` with the current setting?

Peilun-Li:
Hi @neubig . Sure I'll have a relook on this issue.

Peilun-Li:
Hi @neubig , I've adapted the codes to 2.0 with some changes: I store averaged parameter as a separate ParameterStorage in the original parameter's object so that the logic may be more clear. Thanks!

rasoolims:
Is it still not part of the main branch?

neubig:
Sorry this took me forever to do the review on this! I just got back to taking a look, and I think it's much cleaner now. However, I think we can make it even cleaner if instead of passing `maintain_average` parameters, and `emadecay` to the trainers, we just pass the exponential moving average decay when we add parameters, and if this is a non-negative value (or more accurately, value between 0 and 1), we save this in the Parameters object. Then we only need to change `Trainer`, instead of adding something to each of the individual trainers.

Also, one thing I'd like to confirm: I realize that this commit adds a bit of complexity. If people are interested in using parameter averaging then I think this is worth adding, but if this is a minority feature that few people would want then maybe we should consider whether it's worth the added complexity (sorry about just mentioning it now, but I think it's worth thinking about).

rasoolims:
I agree that this adds complexity but my manual averaging experiments (via maintaining averaged numpy matrices) shows significant improvements in two different tasks. Maybe it is worth adding a guideline page on how to manually manipulate experissions and save/load them to/from numpy arrays.

neubig:
I'm going to close this, although adding information about how to do parameter averaging to the doc would be good. Sorry about the hassle @Peilun-Li, we appreciate the contribution even though in the end we didn't get to merge it.

PR #395, opened 2017-03-22
Fixed transpose dimension type

### LINKED ISSUE #394: transpose in Python with visualization enabled returns incorrect dimension
The following snippet:

```
import dynet
v = dynet.vecInput(1)
v.set([0])
t = dynet.transpose(v)
s = dynet.softmax(t)
```

works fine, but crashes if passed the `--dynet-viz` flag with this error:

```
  File "(...)/break_viz_transpose.py", line 7, in <module>
    s = dynet.softmax(t)
  File "(...)/lib/python3.6/site-packages/dyNET-0.0.0-py3.6-macosx-10.12-x86_64.egg/dynet_viz.py", line 311, in softmax
    def softmax(x): return GVExpr('softmax', [x], copy_dim(x))
  File "(...)/lib/python3.6/site-packages/dyNET-0.0.0-py3.6-macosx-10.12-x86_64.egg/dynet_viz.py", line 84, in copy_dim
    if a.dim.isvalid():
AttributeError: 'list' object has no attribute 'isvalid'
```

This seems to be because of line 316 in `dynet_viz.py`, which returns a list for dimensions instead of a SimpleConcreteDim object:

```
def transpose(x): return GVExpr('transpose', [x], x.dim[::-1] if x.dim.isvalid() else InvalidDim)
```

MattX:
Fixes #394

### DIFF
--- a/python/dynet_viz.py
+++ b/python/dynet_viz.py
@@ -313,7 +313,7 @@ def softsign(x): return GVExpr('softsign', [x], copy_dim(x))
 def pow(x, y): return GVExpr('pow', [x,y], ensure_same_dim(x,y))
 def bmin(x, y): return GVExpr('bmin', [x,y], ensure_same_dim(x,y))
 def bmax(x, y): return GVExpr('bmax', [x,y], ensure_same_dim(x,y))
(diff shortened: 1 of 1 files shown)


neubig:
Thanks a lot!

PR #397, opened 2017-03-23
rename `concat_batch_elems` to `concat_to_batch`

jcyk:
Concatenate list of expressions to a single batched expression.


neubig:
Thanks!

Issue #393, opened 2017-03-22
clip the values in a parameter


jcyk:
It's useful to keep model parameters in a small fixed range, say \[-0.01, 0.01\], like WassersteinGAN.

Is there a way to do that (in place)?

if no, I think it may be good to create some API like [`numpy.clip`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.clip.html) (i.e., `dynet.clip(parameter, left, right)`)

After some research on the source code, I found there may be various ways to do that. Not sure to make it a member function of `Parameter` or `Tensor` or `TensorTools` or `ParameterStorage` ....

yoavg:
Hmm... I worry `dynet.clip(...)` will be confused with the other dynet operations that work on expressions and produce expressions, and can be backproped through. I think something like `param.clip_inplace(left, right)` might be better in terms of the external API.

neubig:
I think this sounds like something that would be useful to have, and agree with Yoav on the API.

PR #398, opened 2017-03-23
allow clip parameters into a fixed range

jcyk:
1. For python, add function `clip_inplace` to `Parameters`

2 For C++, add function `clip` to `ParameterStorage`


jcyk:
To add #393.
Clip the values in the parameter to a fixed range [left, right] (in place)

jcyk:
@yoavg  ah, I didn't realize the design.
Of course, let the C++ API exposed at `Parameter` will be better.  
See the latest commit.

pmichel31415:
Very cool, with this WGAN are a step closer

dynet/tensor.cc
```diff
@@ -136,6 +136,23 @@ void TensorTools::CopyElements(const Tensor& v, const Tensor& v_src) {
 #endif
 }
 
+void TensorTools::Clip(Tensor& d, float left, float right){
+#if HAVE_CUDA
```
neubig: I don't think this is correct on CUDA. It looks like this is declaring a new array of floats completely unrelated to the content of `d`, then clipping it and over-writing `d`.

I think the best way to do this is correctly is creating a functor called `FClip` that performs clipping, then replace this whole function (both CPU and GPU) with:

    d.device(*d.edevice) = d.unaryExpr(FClip(left, right));

Take a look at `functors.h` for example of functors.

yoavg:
Btw, how does this interact with regularization, or with gradient tracking for the various training algorithms, if at all?

jcyk:
@neubig Sorry. I just borrowed the code and idea from other functions of `TensorTools`, The incorrectness is caused by my typo and easy to fix. However, you are right about the efficiency. Then I am thinking it may be time to review other code in `TensorTools`, e.g., `TensorTools::Identity` etc.

@yoavg Now, it just clips the parameter values, No any effect to what you mentioned, I think.

I made a new commit following @neubig 's suggestion.

neubig:
Yes, this will be incompatible with weight decay (FYI: weight decay is not L2 regularization unless you are using SGD, and weight decay is no longer on by default). There are two ways to fix this:

1. *Quick fix:* document this, and note that clipping will not be correct if weight decay is not turned on.
2. *Real fix:* make the `clip()` function read the weight decay parameter and adjust the range appropriately.

jcyk:
ah, I just realized the very special design of weight decay in Dynet!
I think it is not easy to do the real fix as the weight decay parameter is currently saved in other `Trainer` instances ( not even accessible from `Model`). That will make the API complicated.
a possible easier solution for users (who turn on weight decay) is to call `Trainer::rescale_and_reset_weight_decay` before `clip`?  (and for other problems it may raise, to reset related tracks after `clip`? I think the clip operation is kind of brute-force, it is hard to tell what is "correct", may just document it and let the user to decide)

Anyway, I just found the current commit won't pass the GPU-version compiling.
```
Linking CXX executable train_rnnlm
../dynet/libgdynet.so: undefined reference to `Eigen::internal::TensorExecutor<Eigen::TensorAssignOp<Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long>, 0, Eigen::MakePointer>, Eigen::TensorCwiseUnaryOp<dynet::FClip, Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long>, 0, Eigen::MakePointer> const> const> const, Eigen::GpuDevice, false>::run(Eigen::TensorAssignOp.<Eigen::TensorMap<Eigen::.Tensor</float, dynet1, 0, /libgdynet.solong>, 0:, Eigen undefined: reference to:MakePointer> `, EigenEigen::TensorCwiseUnaryOp<dynet::::FClipinternal:, Eigen:TensorExecutor::TensorMap<Eigen<Eigen::Tensor:<float:, 1, 0TensorAssignOp<,Eigen::TensorMap long>,<Eigen 0::Tensor<, Eigenfloat,:: 1, 0,MakePointer>  longconst>>,  const> const&0, , EigenEigen::MakePointer::GpuDevice>, const&)' Eigen:
:TensorCwiseUnaryOp<dynet::FClip, Eigen::TensorMap<Eigen:collect2: error: ld returned 1 exit status
:Tensor<float, 1, 0, long>, 0, Eigen::MakePointer> const> const> const, Eigen::GpuDevice, false>::run(Eigen::TensorAssignOp<Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long>, 0, Eigen::MakePointer>, Eigen::TensorCwiseUnaryOp<dynet::FClip, Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long>, 0, Eigen::MakePointer> const> const> const&, Eigen::GpuDevice const&)'
collect2: error: ld returned 1 exit status
```
 Any idea to fix that?

neubig:
Hmm, I believe that this is due to the fact that `tensor.cc` is not compiled using `nvcc`. There is some tricky stuff that we're doing to get things to compile properly on both CPU and GPU. Take a look at the `gpu-*.cc` files and `CMakeList.txt` and you'll see that there's a trick that we're using to compile some files with both gcc (for CPU) and nvcc (for GPU). I can't give detailed guidance now, but if you can work through this and fix the bug that's great. If not, I'll try to take a look when I have a bit of time.

jcyk:
The new commit should support both GPU and CPU.  In addition, it attempts to make a real fix to the incompatibility with weight decay.

neubig:
Looks great, thanks!

PR #401, opened 2017-03-24
remove boost split function in device.cc

sunalbert:
remove boost split function in device.cc to get rid of boost dependence.

### DIFF
--- a/dynet/devices.cc
+++ b/dynet/devices.cc
@@ -1,6 +1,5 @@
 #include "dynet/devices.h"
 
-#include <boost/algorithm/string.hpp>
 #include <iostream>
 #include <unsupported/Eigen/CXX11/Tensor>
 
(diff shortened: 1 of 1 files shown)


sunalbert:
As mentioned in the #84 to remove dependency on boost, I modify the code in device.cc and remove the boost split function. What do you think about it? I find boost is also used in other components,such as unit test. Is it necessary to replace it with other unit test framework? @neubig

neubig:
Thanks for this! I appreciate the contribution, but it might be better to avoid making the code more verbose like this. One option is that currently in https://github.com/clab/dynet/pull/384 we are adding some string manipulation functions. Once these are done, we can also add "split" to these functions, which will help remove this dependency.

sunalbert:
Ok, your idea is more elegent.

PR #406, opened 2017-03-27
Proposed example for rnnlm-batch-nce

cherryc:
I have place an example of an NCE RNNLM in examples/cpp/rnnlm-batch-nce.

I screwed up with git (tried to get fancy with rebases and paid for it), making the commit list wonky. There are 16 unnecessary commits there. However, the change list is perfectly sane (3 new files and changing the makefile), so hopefully this is salvageable. If not, let me know, and I can repair things in my fork and submit another pull request.

Opted to extend the benchmark RNN as I was more familiar with it, and it has a lot of nice timing code built in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,11 @@ endif()
 # echo "export BOOST_ROOT=/cygdrive/d/tools/boost_1_58_0/boost_1_58_0" >> ~/.bashrc
 # then run source ~/.bashrc to have those environment variable effective immediately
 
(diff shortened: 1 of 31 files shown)


neubig:
Awesome, thanks! I tested this out and everything seems to be working. Great speedups when training on CPU :)

cherryc:
I'm glad to help out, as the changes you made to support this helped us a lot. I'm also glad to hear everything is working. Thanks for looking at this so quickly!

Issue #404, opened 2017-03-25
Numeric issue with binary_log_loss


begab:
Hi,

When using the binary_los_loss function from the python interface, there seems be some numeric issues arising in case the network output is 1 and the expected output is 0 (the same thing does not happen in the reverse direction, i.e. 0 as prediction and 1 as expected output).
Here is an MWE, which produces for me the above:

```
In [2]: binary_log_loss(scalarInput(0), scalarInput(1)).value()
Out[2]: 87.3365478515625
In [3]: binary_log_loss(scalarInput(1), scalarInput(0)).value()
Out[3]: 1.1754943508222875e-38
```

The same is not experienced though if I pass arguments to the function in vector format.
```
In [4]: binary_log_loss(inputVector([1,0]), inputVector([0,1])).value()
Out[4]: 87.3365478515625
In [5]: binary_log_loss(inputVector([0,1]), inputVector([1,0])).value()
Out[5]: 87.3365478515625
```

Thanks in advance,
Gabor

neubig:
Resolved by https://github.com/clab/dynet/pull/412

PR #412, opened 2017-03-29
Binary logloss fix

pmichel31415:
Adresses #404 

Essentially now log(0) consistently returns 87. ...  which is log(DYNET_DEVICE_MIN)

Added unit testing


neubig:
LGTM. Thanks!

Issue #122, opened 2016-10-23
Layer Normalization


zhang-wen:
hello, guys, do implementations of run or lstm in  dynet use Layer Normalization Technique which can speed run Training (make earlier convergence) , if not, i suggest that you guys try it, thanks.

neubig:
You're referring to this, right?: https://arxiv.org/abs/1607.06450

This is not implemented, but it'd be nice to have. If you want to implement it @zhang-wen , we'd certainly appreciate a pull request.

zhang-wen:
yes, it is that paper.
need to implement by C++ ?

neubig:
Yes, C++ would be better. If that is not possible a prototype in Python might also be sufficient to allow us to translate it into c++.

zhang-wen:
https://github.com/ryankiros/layer-norm/blob/master/layers.py 
actually this contains an implementation on theano and torch, which is probably implemented by Jamie Ryan Kiros.
you can see the ln() function, but i think if you add Layer Normalization in your network framwork, it still need some coding, because the Layer Normalization is applied on the weights and bias of lstm, gru or run.

danielhers:
Is *batch* normalization implemented?

neubig:
No, neither of these are implemented, but we'd be happy to have a pull request!

neubig:
Layer norm is implemented by https://github.com/clab/dynet/pull/482

PR #417, opened 2017-03-30
[WIP]add cpu version batch normalization

sunalbert:
As mentioned in #122，dynet does not support batch normalization and layer normalization operation. I am trying to add batch normalization(https://arxiv.org/pdf/1502.03167.pdf), which is useful in GAN and CNN. The task has some sub-tasks:
- [ ] batch statistics operation: to get mean and variance of batch data. This operation also needs compute the moving_average and moving_variance which are used in test mode.I find there is a similar PR #385, maybe I can implement batch statistics with it.
- [x] batch normalization operation: based on the mean and variance provided by batch statistics, normalize the batch data.
- [ ] necessary comments and documents for batch statistics and batch normalization.
- [ ] unit test and some detailed debugging


sunalbert:
I find a Average operation in dynet, but it is not applicable for batch normalization, because batch normalization is a channel-wise operation and needs compute moving average. I'd like to add a new batch_mean operation rather than modify Average operation directly so that we can avoid compatibility problems. What do you think about it? any better idea? @neubig

pmichel31415:
Hi, just chiming in to share an idea I had when I though of implementing batchnorm in dynet a while back : 

would it be easier to implement it as a separate class (much like RNNBuilder or SoftmaxBuilder, something like NormalizationLayer...)? So you can store the statistics as attributes and you don't need to change the expression interface to accommodate a train/test switch.

Also this could be extended to implement BatchReNorm, LayerNorm, etc...

With this the only actual nodes that need to be implemented are the averaging, broadcasted multiplication/division, etc..

sunalbert:
Hi @pmichel31415, thanks for your reply. In my original idea, I should implement average operation, variance operation, batchnorm operation. They can be called as following:
`mean = mean(input, train_or_test)`
`variance = variance(input, train_or_test)`
`out = bacthnorm(mean, variance)`
If we implement batch norm as a  separate class, user can call batch normalization operation in one line code:
`out =  batchnorm(input)`
Right?

PR #420, opened 2017-04-02
fix issue419

Oneplus:
On MSVC, `DYNET_PP_NARG_` macro is not expanded correctly, which leads to errors in saving/loading models on MSVC. More detailedly, `DYNET_PP_NARG_` always expands to 1 on MSVC, thus, only the first member is serialized.

solution: use `BOOST_PP_VARIADIC_SIZE` instead of `DYNET_PP_NARG_`. (should be more portable using boost but I didn't change the original code.)

`({})` is also not supported on MSVC (http://stackoverflow.com/questions/5291532/does-msvc-have-analog-of-gccs), so change the implementation of `DYNET_VERSION_SERIALIZE_DEFINE`.

### DIFF
--- a/dynet/io-macros.h
+++ b/dynet/io-macros.h
@@ -32,8 +32,14 @@
   template void MyClass::save<boost::archive::binary_oarchive>(boost::archive::binary_oarchive &ar, const unsigned int) const; \
   template void MyClass::load<boost::archive::binary_iarchive>(boost::archive::binary_iarchive &ar, const unsigned int);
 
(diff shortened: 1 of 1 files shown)


xunzhang:
BTW, could I know why on MSVC, `DYNET_PP_NARG_` only expands to 1. What do you mean here? Thanks.

Oneplus:
@xunzhang An example for that is

```
std::cout << DYNET_PP_NARG(a) << std::endl; // MSVC result: 1
std::cout << DYNET_PP_NARG(a, b) << std::endl; // MSVC result: 1
std::cout << DYNET_PP_NARG(a, b, c) << std::endl; // MSVC result: 1
```

It seems this thread http://stackoverflow.com/questions/5530505/why-does-this-variadic-argument-count-macro-fail-with-vc explains the difference between gcc and MSVC.

Issue #230, opened 2017-01-04
Better support for saving/loading some, but not all parameters


neubig:
Currently it is not easy to save/load only some of the parameters in the model. It would be nice to have an interface that made this easier.

This is partially related to: https://github.com/clab/dynet/issues/84

neubig:
Fixed by https://github.com/clab/dynet/pull/637

PR #422, opened 2017-04-03
Implement native serialization and deserialization logic for ParameterCollection and support partial save/load.

xunzhang:
This pull request means to implement native save/load serialization and deserialization logic without Boost. https://github.com/clab/dynet/issues/230 https://github.com/clab/dynet/issues/84

There are still some commits need to be finished in this pull request.
- [x] Checking existing model file to avoid confliction.
- [x] Add more save and load interfaces.
- [x] Refactor and add more tests.

Moreover, there are still some ideas need to be discussed.
1. Do we need to support save/load a single `Parameter` and `LookupParameter`? If yes, although current mechanism can also apply while some existing code(exclude pull request) need to be refactored a lot.
2. Current implementation doesn't support Builder load which is supported in current version.
3. Current format is text based one considering readability, it is kind of a mimic version of JSON/YAML, but this will take more space and might lose some precision.

### DIFF
--- a/dynet/dim.cc
+++ b/dynet/dim.cc
@@ -7,7 +7,7 @@ using namespace std;
 namespace dynet {
 
 ostream& operator<<(ostream& os, const Dim& d) {
-  os << d.nd << '-' << '{';
+  os << '{';
   for (unsigned i = 0; i < d.nd; ++i) {
     if (i) os << ',';
     os << d.d[i];
@@ -26,18 +26,25 @@ ostream& operator<<(ostream& os, const vector<Dim>& ds) {
 istream& operator>>(istream& is, Dim& d) {
   char place_holder;
   int nd = 0;
-  is >> nd >> place_holder;
   is >> place_holder;
-  d.resize(nd);
-  for (unsigned i = 0; i < nd; ++i) {
-    if (i) is >> place_holder;
+  d.resize(DYNET_MAX_TENSOR_DIM);
+  bool batch_flag = false;
+  unsigned i = 0;
+  for (; i < DYNET_MAX_TENSOR_DIM + 1; ++i) {
+    if (i) {
+      is >> place_holder;
+      if (place_holder == 'X') {
+        batch_flag = true;
+        break;
+      } else if (place_holder == '}') {
+        break;
+      }
+    }
     is >> d.d[i];
   }
-  is >> place_holder;
-  if (place_holder == 'X') {
+  d.resize(i);
+  if (batch_flag) {
     is >> d.bd >> place_holder;
-  } else {
-    is >> place_holder;
   }
   return is;
 }
--- a/dynet/dim.cc
+++ b/dynet/dim.cc
@@ -7,7 +7,7 @@ using namespace std;
 namespace dynet {
 
 ostream& operator<<(ostream& os, const Dim& d) {
-  os << '{';
+  os << d.nd << '-' << '{';
   for (unsigned i = 0; i < d.nd; ++i) {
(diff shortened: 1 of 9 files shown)


xunzhang:
To explain why we don't need some interfaces like:
`save(vector<ParameterCollection> & models, string)`: there is no case using vector<ParameterCollection>. If they are independent, user can call `save(ParameterCollection, string)` one by one. If they are dependent, they should be only expanded into one `ParameterCollection`.

`save(vector<ParameterCollection> & models, vector<string>)`: user could do it one by one otherwise, they need to create a vector which is the same work calling save one by one.

Similar reason with `load(vector<ParameterCollection> & models, ...)` interfaces.

dynet/dim.cc
```diff
@@ -7,7 +7,7 @@ using namespace std;
 namespace dynet {
 
 ostream& operator<<(ostream& os, const Dim& d) {
-  os << '{';
+  os << d.nd << '-' << '{';
```
neubig: I think this may be confusing when people try to print out `Dim` for debugging purposes, etc. Could we leave `<<` the same way as before and modify `>>` to automatically infer `nd` as necessary?

dynet/io.cc
```diff
@@ -0,0 +1,246 @@
...
+  os << key_str << ':' << offset << '\n';
+  os.close();
+  // write model into model file
+  this->serialize(model, key, is_append);
+}
+
+void Pack::save(const ParameterCollection & model,
+                const std::vector<std::string> & filter_lst,
+                const std::string & key, bool is_append) {
+  // TODO
```
neubig: throw a not implemented error

dynet/io.cc
```diff
@@ -0,0 +1,246 @@
...
+    std::istringstream iss(line);
+    iss >> d;
+
+    // add param into input model
+    Parameter param = model.add_parameters(d);
+    param.get_storage().name = name;
+
+    // read param.get_storage().values
+    std::vector<float> params_lst;
+    auto deserialize_tensor_lambda = [&] () {
```
neubig: Maybe it's better to just define this and the following one as a private (inline?) member of Pack?

dynet/dim.cc
xunzhang: Good advice, let me think about it.

dynet/io.cc
xunzhang: Sure.

xunzhang:
@neubig Done

xunzhang:
For this, there are still some work remaining to be finished and enhanced. I will list them below for future pull requests.

- Python binding of new save/load interface.
- Support Parameter, LookupParameter save and load.
- Remove boost serialization code part.
- Support high dimension for ParameterCollection.

pmichel31415:
Hi quick question : will it still be possible to save parameters in python until the bindings are updated?

yoavg:
I guess not, but they will likely be updated quickly (that's my plan at
least). Having said that, I rather have a few days delay then to rush out a
premature API.

xunzhang:
@pmichel31415 Hi, the order of doing python binding and support Parameters in my last comment depends on the engineering effort I think. Both of them have high priority in our next release(v1.1).

pmichel31415:
OK let me know if you want help for the python bindings, I'd rather have them in this update

yoavg:
Me too. I think we shouldn't update the cpp serialization w/o the Python.

neubig:
@yoavg @pmichel31415 Note that this is on the new-readwrite branch, not master. Master won't be updated until everything is finished.

PR #423, opened 2017-04-04
Allow choosing dimension to pick for Python

Peilun-Li:
It seems we can choose a specific dimension to `pick` in C++ but not Python. This PR fixes this problem.


neubig:
Thanks!

PR #427, opened 2017-04-05
Implement RMSProp and refactor python trainers bindings

pmichel31415:
- implements RMSProp and adds unit tests & python bindings

- changes the way trainers are implemented in cython to leverage inheritance and remove copy-pasted code

- Adds documentation to python trainers


pmichel31415:
Just fixed a bug were the learning rate wasn't taken into account, now it works.

I've got slighty better results than Adam with this on one of my RNN tasks FWIW.

neubig:
Thanks!

PR #428, opened 2017-04-06
Fix BOOST_ROOT conditions at CMakeLists.

odashi:
Both the environment variable ($ENV{BOOST_ROOT}) and the -D option ($BOOST_ROOT) should be performed similarly, as well as the name "BOOSTROOT".


neubig:
Thank you!

PR #433, opened 2017-04-07
Change Cython version checking to match the way Cython is used

mattr1:
In the python build, Cython is used through importing the library, rather than calling the cython executable. This change prevents a problem when the PYTHON environment variable passed in to CMake differs from the default python and cython found in the system path. Also, the code is much simpler this way.


neubig:
@mattr1 Thanks! But this seems to be failing CI tests. Could you take a look?

mattr1:
@neubig Fixed. Forgot to make it python 3 compatible by adding parenthesis with print().

neubig:
Thanks!

Issue #169, opened 2016-11-17
add python to the documentation


yoavg:
Currently, the documentation is about the c++ modules.

Would be nice to be explicit about this, and also add links to some of the python resources.
While these are not yet very nice or exhaustive, we do have some useful resources:

https://github.com/clab/dynet/blob/master/python/CHANGES.md
https://github.com/clab/dynet/tree/master/pyexamples
https://github.com/clab/dynet/tree/master/pyexamples/tutorials

and of course the EMNLP tutorial (with examples and slides):
https://github.com/clab/dynet_tutorial_examples

These should be added to the main documentation page.
I'd do it myself, but I am not sure on how the documentation system is set up.
@pmichel31415 ?

pmichel31415:
So there are detailed guidelines in the documentation, under the "contributing" section [here](http://dynet.readthedocs.io/en/latest/doc_style.html)

TL;DR

If you want to write a tutorial you can directly modify the documentation source in [doc/source](https://github.com/clab/dynet/tree/master/doc/source) (careful, this uses rst markup, but you can convert from markdown using pandoc).

If you want to write a reference manual, sphinx can retrieve the documentation from python docstrings directly, although I am not certain it is straightforward with cython ([this stackoverflow question](http://stackoverflow.com/questions/10057476/how-do-i-use-sphinx-with-cython) might help).

For ipython notebooks I don't know if they can be embedded directly, but you can probably link them from the docs.

pmichel31415:
OK I came back to this because this is the biggest barrier of entry right now IMO.

Here's the state of things : 

1. Generating the doc automatically with sphinx (like what we do for c++) requires to compile the cython files (sphinx doesn't analyse the source code apparently, don't ask me why)
2. This means if we want to automate things completely on readthedocs like c++ we need to recompile the whole of dynet on readthedocs before generating the docs (impractical)

I see 2 solutions to this :

1. Cython should be documented with standard python docstrings (this allows at least to call help in the python/ipython interpreter)
2. A(n ugly) workaround for sphinx would be to have a manually compiled `_dynet.so` file on the repo and try to fetch the docs directly from this.

I can start working directly on 1. however I need your opinion on 2. @yoavg @neubig

neubig:
One thing I was thinking about was that it might be worth totally assisting the doc for Python and C++. in other words, the would be totally different readthedocs sites. Would this make things easier?

pmichel31415:
[...]

PR #434, opened 2017-04-08
Python documentation in readthedocs

pmichel31415:
This pull requests addresses #169 

This is kind of a dirty trick because of the limitations of readthedocs but basically what it does is create a 'pythonized' copy of `_dynet.pyx` with empty functions that can be interpreted by sphinx (see [doc/source/doc_util.py](https://github.com/clab/dynet/compare/python_doc?expand=1#diff-5020e1e9d2db8a550c522a52223e9fae) for the script).

This also change the docstring format in `_dynet.pyx` in the more readable [napoleon](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/) format (hence the big diff in _dynet.pyx)

The documentation has been restructured slightly so that the c++ and python reference are well separated and at similar levels..

You can see the result at this url : http://dynet.readthedocs.io/en/python_doc/index.html

This ugly trick will be fixed once dynet becomes available via pip install. Until then, I believe it is a decent enough solution.


yoavg:
Overall this looks very good! thanks! this is really great to have.

Two comments:
* Maybe we should list the python docs above the c++ ones? I think most users will be looking for these.
(but this is really not a major point, can also leave as is)
* More important, things starting with `_` were supposed to be private / internal. I can sort-of see why we would allow people to use `import _dynet`, but I really rather the other cases be hidden from the docs. In particular, `dynet._inputExpression`, `dynet._vecInputExpression` etc are not supposed to be used directly. People should use `dynet.vecInput` etc instead. Is there a way to hide the privates from the docs?

pmichel31415:
OK @yoavg I addressed your 2 points.

One remark though, removing _vecInputExpression for instance hides the `.set()` method which might be useful for people who want to use a static graph.

pmichel31415:
Also for now all functions/class need to be manually listed in `python_ref.rst` because there is no way to make custom groups. The solution would be to have e.g. all operations in a separate file, all RNNBuilders, all Optimizers, etc,..

yoavg:
Looks good to me. 

Re the `set()` functionality, it should ideally be documented on the functions that create and return the input expressions that have a set method.

Re splitting into different files, I was not sure how to do that nicely in cython, while still allowing for `from dynet import *`. But I am not against such a change in any way.

neubig:
Yeah, looks good to me to. Go ahead and merge if you're satisfied @pmichel31415

pmichel31415:
Yes I'll merge now and leave the other issues for further pull requests

PR #435, opened 2017-04-10
More save/load interface implementation.

xunzhang:
Support populating a `Parameter`, `LookupParameter`within or without a ParameterCollection.

- [x] Code clean
- [x] Keep parameters and lookup_parameters' order
- [x] Support high dimension.


neubig:
I browsed this quickly and basically LGTM. (I'll probably do a more complete look when the whole implementation is done). Is this ready to be merged?

xunzhang:
@neubig It's ready to merge I think, otherwise it will be open too many days here. The next pull request will be:
- Remove boost dependencies and remove builder save/load in old code
- Replace boost multi-thread impl with std library and add a boost switch flag to install.
- Python interface for save/load, document update.

neubig:
Awesome, thanks!

PR #436, opened 2017-04-11
rename RmsPropTrainer to RMSPropTrainer in SWIG bindings

joelgrus:
also, a couple of other small changes to the Scala wrappers:

* add an overloaded `(exprs: Expression*)` version of every function that takes an `ExpressionVector` parameter

* change Scala `Expression` instances to contain references to any other Expressions (or Vectors) they depend on, to prevent segfaults that you get from premature GC.


neubig:
Great, thanks!

PR #446, opened 2017-04-14
Fixed Python 3 encoding issue in save_all / load_all

MattX:
Just moved the call to encode() from `save`/`load` to `save_all`/`load_all` to make the API more consistent. Tested with python 2.7 and 3.6.


neubig:
Thanks!

Issue #410, opened 2017-03-28
generate categorical samples


pcyin:
Hi. It would be great if we could have an operation like `categorical_sample(P, size=3)` to sample from a batched tensor `P` of shape `(K, batch_size)`, where each row in `P` is a categorical distribution of size `K`. And `size` is the number of samples to draw from each row.

neubig:
This has been fixed to some extent.

PR #448, opened 2017-04-16
Add Argmax and Sampling capability

neubig:
This pull request addresses https://github.com/clab/dynet/issues/410 by making it possible to perform argmax, or sample from a log probability distribution. It also introduces a new data type called IndexTensor which is similar to Tensor, but contains integers. This is used to store the result of the argmax.

This still needs python bindings (@yoavg ?), and I'd like to get any comments on the interface before making the final commit.


yoavg:
Hi @neubig , could you elaborate a bit on this?
I see new tensor ops and a new tensor type, but does this connect to the Expressions somehow? how is this supposed to be used within a DyNet program?

(currently, the python API assumes tensors are internal to the implementation and does not expose them to the user)

neubig:
Sorry, let me elaborate: the standard use case would be when you want to do greedy search or sampling, which could be done like this:

    x = start_of_sentence
    while x != end_of_sentence:
      rnn_state = rnn_state.update(lookup(x))
      log_prob = make_prediction(rnn_state)
      x = argmax(log_prob.value())

Currently in order to do something like this, you need to call `.npvalue()` (in Python) or `as_vector()` in C++, then do the argmax over that in the native language. The main reason why this is bad is particularly on GPU, where it requires that memory be transferred to the CPU and then back again to do the lookup, and can't take advantage of the fact that GPUs will plow through an argmax much faster than the CPU.

In this commit I still haven't implemented `lookup(x)` where `x` is an IndexTensor, but that will be forthcoming.

yoavg:
oh. doing this in python can prove a bit tricky, as we currently assume any conversion from Expression to a value (by calling .value(), .npvalue(), .scalarvalue() etc) is done through a copy and returns a native python type. This is not a native python type... and supporting this will seriously change the interface.

I am not sure what would be a good way to support this. the best I can think of right now is maybe adding a `.argmax()` capability to Expression. Not sure what it should return though. Do you think it will be OK (performance wise) if it converted the argmax result into native values (say a numpy vector or a python list of numbers) that can then be passed to `lookup`?

neubig:
Hmm, OK what if we did this:

    x = start_of_sentence
    x_val = None
    while x_val != end_of_sentence:
      rnn_state = rnn_state.update(lookup(x))
      log_prob = make_prediction(rnn_state)
      x = argmax(log_prob)
      x_val = x.npvalue()

Then had the `argmax` function internally call the C++ version of `value()`, which gets a DyNet Tensor. The return of `argmax` would by a DyNet `IndexTensor`, and then you could call `npvalue()` to get a numpy array containing the indices contained therein.

I think your solution would also work, but it might be nice to expose an interface that doesn't do any copying from GPU to CPU in case people want it.

yoavg:
My concern here (which isn't huge but should at least be discussed) is that argmax is very similar syntactically to computation graph creation, but does something else (not returning an expression, but a new kind of object). 

It also overloads the meaning of .npvalue(), from meaning "run a forward computation and return the value" to "copy a value from GPU/c to python". 

Also, because Python does not have type based dispatch, we will have to change lookup to do type checking, making it a bit slower in the common case.

yoavg:
Wait, looking at the code again, these are only tensor ops, I didn't see one taking in an expression. So how would the cpp code using it looks like?

neubig:
OK, I think it's a good idea to make this have separate syntax than the regular expressions to ensure that no-one is confused that this is a regular expression in the computation graph.

The C++ code would look like:

    int x = start_of_sentence;
    while(x_val != end_of_sentence) {
      Expression rnn_state = rnn_builder.add_input(lookup(cg, x));
      Expression log_prob = make_prediction(rnn_state);
      IndexTensor x_tensor = TensorTools::argmax(log_prob.value());
      x = as_scalar(x_tensor);
    }

yoavg:
Hmm.. maybe we should add a `.tensorvalue()` method to `Expression`, which will compute forward and return a tensor (i.e., stay on device)? We can then provide argmax etc on top, as well as methods to convert to (and from?) a numpy array (these conversions will require memory copies of course).

neubig:
Yeah, that seems like a good solution.

yoavg:
Ok, I'll try and do it tomorrow.

yoavg:
Added the python wrapping, though very minimally tested.

The API:
```
t = expr.tensor_value()
# t is of type dynet.Tensor
print t.as_numpy()
ti = t.argmax()
# ti is of type dynet.Tensor
print ti.as_numpy()
```

neubig:
Thanks! This will be merged pending a bit of testing.

neubig:
Upon rudimentary further testing, this seems to be working properly, so I'll go ahead and merge.

Issue #442, opened 2017-04-13
Better dimension-wise reductions


neubig:
Currently DyNet's support for some dimensionwise reductions is lacking (only `sum_dim`) is supported. It would also be nice to do the following by an arbitrary dimension:

- [x] `max`
- [x] `min`
- [ ] `k-max`

neubig:
First two resolved by: https://github.com/clab/dynet/pull/449

neubig:
Resolved by https://github.com/clab/dynet/issues/457, although GPU support for kmax is still pending (https://github.com/clab/dynet/issues/245)

Issue #177, opened 2016-11-21
is there max-out API for one dimension ?


zhang-wen:
i have following code to get a 3-D matrix with batch size is 4

```
m = Model()
lp = m.add_lookup_parameters((10,6))
batch_emb = lookup_batch(lp, [1,2,3,9])
batch_emb = reshape(batch_emb, (2, batch_emb.npvalue().shape[0] / 2), 4)
```

the shape of batch_emb is (2, 3, 4)

i want to get the result of 
```
n1 = batch_emb.npvalue()
numpy.max(n1, axis=0)
```

it there this API for Expression ?

neubig:
There is the `kmax_pooling` expression, but its axis is not configurable. Being able to configure the axis is something that would be nice to have, so a pull request would be welcome.

Peilun-Li:
It seems `kmax_pooling` doesn't support batching and only handles 2-d tensor. Wouldn't it be more general to add another function `maxout` supporting 3-d tensor and batching like the way `PickElement` does?

neubig:
Resolved by: https://github.com/clab/dynet/pull/449

PR #449, opened 2017-04-16
Add max_dim and min_dim functions

Peilun-Li:
This PR solves #177 and part of #442 . I add two functions `max_dim` and `min_dim` to support dimension-wise reductions by max/min values towards a given dimension. (Since there are already `max` and `min` functions for expression-wise comparison I don't use the same name. Are there any more intuitive function names for these dimension-wise reduction functions?) 

In detail, `max_dim` and `min_dim` can support up to 3-d tensor and batching (like PickElement), and the output will be an expression of sub-tensor with max/min values along a given dimension. 

Sample usage:
(Python) `expr = max_dim(batch_emb, 1)`
(C++) `Expression expr = max_dim(batch_emb, 1)`
The second parameter is the dimension to perform reduction on. (default value is 0, i.e., the first dimension)

### DIFF
--- a/dynet/expr.cc
+++ b/dynet/expr.cc
@@ -138,8 +138,8 @@ Expression sum_batches(const Expression& x) { return Expression(x.pg, x.pg->add_
 
 Expression kmh_ngram(const Expression& x, unsigned n) { return Expression(x.pg, x.pg->add_function<KMHNGram>({x.i}, n)); }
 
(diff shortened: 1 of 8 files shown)


neubig:
Thanks, this is excellent. I looked at the code and it seems very well implemented.

For the naming, I propose that we name all dimensionwise reductions "xxx_dim", like "sum_dim" which would make these "max_dim" and "min_dim", or perhaps "maxout_dim", "minout_dim".

Peilun-Li:
Hi @neubig , I've changed the function names to max_dim and min_dim. Thanks!

Issue #445, opened 2017-04-14
delete_dim does not work as expected


sunalbert:
Hi, When implementing dimension-wise mean, I find function: delete_dim in dim.h may not work as expected. 
If a tensor's dim is 
`Dim({2, 5, 7}, 10)`
and argument of delete_dim function `i` is 2, which means I wanna reduce a tensor with 4 rank(including batch dimension) along the third dimension. The expected result is Dim({2, 5}, 10), but in fact delete_dim 
returns Dim({2, 5, 7}, 10) , since the loop requires `i+1<nd`
Besides, in some scenarios, the batch dimension may also need to be reduced, but current delete_dim cannot handle it. Maybe we need a new delete_dim ,which can delete multi dimensions in one call and delete batch dimension when necessary.

neubig:
There are two parts here. The first part of not working when `i` is 2 seems like a bug.

For the second part, deleting the batch dimension is as simple as `d.bd = 1`.

sunalbert:
Thanks for your reply.I think a delete_dim function which can delete multi dimension will be better.In the following case:
the tensor's dim is Dim({2, 5, 7, 8}, 10), I want to  reduce the second(index is 1) and the third dimension(index is 2). The right result is Dim({2,8}, 10).
I have to call delete_dim twice, first calling delete_dim(1), resulting in Dim({2,7,8}, 10). Notice that the original third dimension has become the second dimension, it may result in bug, if user call delete_dim(2).
If we have a function delete_dim([1,2]),we can get right result simply.

neubig:
If you think that's necessary, sure, feel free to add it (maybe with the name `delete_dims`)! `delete_dim` is an internal function that users won't be expected to use directly, so the interface is not super-important.

sunalbert:
Hi, @neubig . I need some some suggestion from you about the new function: delete_dims and mean_reduce.
First, Let me explain why we need delete_dims. In some cases ,we need to reduce multi-dimensions.For example,in batch normalization, a tensor with rank 4({d[0], d[1], d[2]}, bd), we should get the average of it along d[0], d[1], and batch dimension.
In other frameworks, we can make it by calling mean_reduce(tensor, [0,1,3]). Since the batch dimension in dynet is supposed to be different from all others, I think a mean reduce interface implemented as following may be a choice:
```
mean_dim(tensor, dims, reduce_batch )
tensor: tensor
dims: a list of dimensions to be reduced
[...]

PR #450, opened 2017-04-17
fix bug in delete_dim function 

sunalbert:
This PR solves part of #445. I modify the function delete_dim to make it able to handle some edge cases. Now this function can delete batch dimension and work as expected  when i==nd-1.

### DIFF
--- a/dynet/dim.h
+++ b/dynet/dim.h
@@ -184,7 +184,7 @@ struct Dim {
    * \param i index of the dimension to be removed
    */
   inline void delete_dim(unsigned int i) {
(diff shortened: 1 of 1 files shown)


neubig:
Thanks! I actually would prefer not to have delete dimension delete the batch dimension like this though. The batch dimension is supposed to be different from all the others, and is not simply the `n+1`th dimension in an n-dimensional tensor. So basically if you delete the first "if" in the big "if else" statement, I think this should be OK.

sunalbert:
Ok. I have deleted it. Now the delete_dim will work as expected

neubig:
Thanks!

PR #451, opened 2017-04-17
Document python operations

pmichel31415:
This documents every operation available in python. Putting this as a pull request since it's a big commit.

Also adds new operations like 2d conv and maxout.

The result can be seen here http://dynet.readthedocs.io/en/python_doc/python_ref.html#id1


neubig:
Thanks so much! This is excellent.

PR #452, opened 2017-04-17
Fix CUDNN to work with Windows.

mattr1:
Also, the change from "PATHS" to "HINTS" is just so that we use the CUDNN_ROOT instead of system path to find cudnn.


neubig:
Thanks!

PR #453, opened 2017-04-18
Initializing biases to zero

odashi:
Bias parameters are generally initialized to zero when no prior knowledge for data is available. This behavior keeps compatibility with some matrix initialization schema (cf. Goodfellow's "Deep Learning", pp.296-).
This PR naively adds the zero() function just after creating bias parameters.

### DIFF
--- a/dynet/cfsm-builder.cc
+++ b/dynet/cfsm-builder.cc
@@ -22,6 +22,7 @@ StandardSoftmaxBuilder::StandardSoftmaxBuilder() {}
 StandardSoftmaxBuilder::StandardSoftmaxBuilder(unsigned rep_dim, unsigned vocab_size, Model& model) {
   p_w = model.add_parameters({vocab_size, rep_dim});
   p_b = model.add_parameters({vocab_size});
+  p_b.zero();
 }
 
 void StandardSoftmaxBuilder::new_graph(ComputationGraph& cg) {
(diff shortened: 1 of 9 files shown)


neubig:
Thanks! In general I'm in favor of changing this, but the canonical DyNet syntax would be to use `model.add_parameters({vocab_size}, ParameterInitConst(0));`
(It might be worth adding `ParameterInitZero()`?)

neubig:
Also, for LSTMs the bias for the forget gate should probably be set to 1. I realize this wasn't included in the previous version, but this might be a good opportunity to fix this.

odashi:
Okay, I missed the `ParameterInit` argumet.
And, probably it is good to give the forget bias via an additional constant (e.g. Tensorflow's LSTM implementation uses the expression $\sigma(Wx+b+1)$ ).

pmichel31415:
Sorry I saw this PR but didn't comment, basically LGTM and I agree that the +1 bias is a good way to set the forget gate bias to 1.

Just before merging I think there are some inconsistencies in where the +1 bias is implemented and documented, looks like it's documented for the LSTMBuilder but implemented for the VanillaLSTMbuilder?

I think it should be implemented (and documented) for 

- LSTMBuilder (lstm.h/lstm.cc)
- VanillaLSTMBuilder (lstm.h/lstm.cc)
- FastLSTMBuilder (fast-lstm.h/fast-lstm.cc)
- DeepLSTMBuilder (deep-lstm.h/deep-lstm.cc)
- GRUBuilder (gru.h/gru.cc) : not 100% sure if it's also a good thing in grus though

odashi:
The documentation about the bias is of VanillaLSTMBuilder (GitHub's suggestion looks some kind of confusing).
Perhaps we should separate the VanillaLSTM implementation into other files.

In LSTM/FastLSTM/DeepLSTM, the forget gate is designed as CIFG which is *forget_gate = 1 - input_gate*.
I guess there is no similar way to add bias in VanillaLSTM/TreeLSTM that have independent parameters for the forget gate.
Is there any good way?

pmichel31415:
Oh OK I get it now.

Yeah I'm not sure if forget gate bias would be a good thing when the forget and input gate are coupled (this would mean setting `i=sigmoid(Wx+b-1)` I guess)

Anyway this can be addressed later, if everything is ok for you @neubig I think we can merge

neubig:
Yeah, I think this is good. I'll go ahead and merge.

Issue #456, opened 2017-04-18
Convolution doc is_valid is difficult to understand


neubig:
The doc for the is_valid parameter here is difficult to understand. Could we explain it more clearly? https://github.com/clab/dynet/blob/master/dynet/expr.h#L1618

zhisbug:
Assume:
- Input feature maps: XH x XW x XC x N
- Filters: FH x FW x XC x FC 
- Strides: strides[0] and strides[1] are row (h) and col (w) stride, respectively.

For the **SAME** convolution:  the output height (YH) and width (YW) are computed as:

- YH = ceil(float(H) / float(strides[0]))
- YW  = ceil(float(W) / float(strides[1]))

and the paddings are computed as:

- pad_along_height = max((YH - 1) * strides[0] + FH - XH, 0)
- pad_along_width = max((YW - 1) * strides[1] + FW - XW, 0)
- pad_top = pad_along_height / 2
- pad_bottom = pad_along_height - pad_top
- pad_left = pad_along_width / 2
- pad_right = pad_along_width - pad_left


For the **VALID** convolution:  the output height (YH) and width (YW) are computed as:

- YH = ceil(float(XH - FH + 1) / float(strides[0]))
- YW  = ceil(float(XW - FW + 1) / float(strides[1]))

and the paddings are always zeros.

Think about when stride is 1, the distinction:

**SAME**: output size is the same with input size. To do so, one needs to pad the input so the filter can sweep outside of the input maps
**VALID**: output size shrinks by filter_size - 1, and the filters always sweep at valid positions inside the input maps. No padding needed.

neubig:
Thanks! This is awesome. Could we add something to this effect to the doc?

zhisbug:
@neubig Sure, let me file a PR to update the documents.

pmichel31415:
Hi @zhisbug can you also update the python doc [here](https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L1934) and [here](https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L1953)

thanks!

zhisbug:
We can close this issue now as the documents have been fixed.

PR #458, opened 2017-04-19
Add more detailed doc for conv2d

zhisbug:
Add more documentation for issue: https://github.com/clab/dynet/issues/456

### DIFF
--- a/dynet/expr.h
+++ b/dynet/expr.h
@@ -1612,7 +1612,12 @@ Expression kmh_ngram(const Expression& x, unsigned n);
  * \brief conv2d without bias
  * \details
  *   2D convolution operator without bias parameters.
- *   'VALID' and 'SAME' convlutions are supported. Assume:
+ *   'VALID' and 'SAME' convolutions are supported.
+ *   Think about when stride is 1, the distinction:
+ *   - *SAME*: output size is the same with input size. To do so, one needs to pad the input so the filter can sweep outside of the input maps.
+ *   - *VALID*: output size shrinks by filter_size - 1, and the filters always sweep at valid positions inside the input maps. No padding needed.
+ *
+ *   In detail, assume:
  *   - Input feature maps: XH x XW x XC x N
  *   - Filters: FH x FW x XC x FC 
  *   - Strides: strides[0] and strides[1] are row (h) and col (w) stride, respectively.
@@ -1632,10 +1637,6 @@ Expression kmh_ngram(const Expression& x, unsigned n);
  *   - YH = ceil(float(XH - FH + 1) / float(strides[0]))
  *   - YW = ceil(float(XW - FW + 1) / float(strides[1]))
  *   and the paddings are always zeros.
- *
- *   Think about when stride is 1, the distinction:
- *   *SAME*: output size is the same with input size. To do so, one needs to pad the input so the filter can sweep outside of the input maps.
- *   *VALID*: output size shrinks by filter_size - 1, and the filters always sweep at valid positions inside the input maps. No padding needed.
  *
  * \param x The input feature maps in H x W x Ci x N (ColMaj)
  * \param f 2D convolution filters H x W x Ci x Co (ColMaj)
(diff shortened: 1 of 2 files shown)


dynet/expr.h
```diff
@@ -1610,7 +1610,32 @@ Expression kmh_ngram(const Expression& x, unsigned n);
 /**
  * \ingroup convolutionoperations
  * \brief conv2d without bias
- * \details 2D convolution operators without bias parameters.
+ * \details
+ *   2D convolution operator without bias parameters.
+ *   'VALID' and 'SAME' convlutions are supported. Assume:
```
neubig: convlutions -> convolutions

dynet/expr.h
```diff
@@ -1610,7 +1610,32 @@ Expression kmh_ngram(const Expression& x, unsigned n);
...
+ *   - pad_bottom = pad_along_height - pad_top
+ *   - pad_left = pad_along_width / 2
+ *   - pad_right = pad_along_width - pad_left
+ *
+ *   For the *VALID* convolution: the output height (YH) and width (YW) are computed as:
+ *   - YH = ceil(float(XH - FH + 1) / float(strides[0]))
+ *   - YW = ceil(float(XW - FW + 1) / float(strides[1]))
+ *   and the paddings are always zeros.
+ *
+ *   Think about when stride is 1, the distinction:
```
neubig: Maybe move this to the front of the explanation, then describe the details later?

zhisbug:
@neubig Done with the adjustment!

ccsasuke:
Hi,

For "Input feature maps: XH x XW x XC x N":

When I tested this, it seems DyNet only accepts 3D tensors instead of 4D ones for X?
My X with shape (XH x XW x XC) worked but (XH x XW x XC x N) didn't (even with N=1).

zhisbug:
@ccsasuke I think N = 1 is supported. Maybe you want to check your code by taking a look at this test case I wrote here:
https://github.com/clab/dynet/blob/master/tests/test-nodes.cc#L923

ccsasuke:
Ah I see. Looking at the test case, it seems that I directly used a 4D tensor (e.g. Dim({50, 50, 2, 2}, 1)) instead of Dim({50, 50, 2}, 2).
Thanks.

neubig:
Yes, I think the documentation needs to be fixed to indicate that the input is not 4 dimensional, but 3 dimensional (and optionally batched).

ccsasuke:
Sorry for hijacking this thread, but just want to make sure I did not do it wrong before filing a separate issue.

When I used conv2d, it seems it requires the input size larger than the filter size even if I was using "SAME" padding.
The relevant codes are:
  if (xs[0].ndims() != 3 || xs[1].ndims() != 4 ||
      xs[1].d[2] != xs[0].d[2] || xs[0].d[0] < xs[1].d[0] ||
      xs[0].d[1] < xs[1].d[1]) {
    ostringstream s; s << "Bad input dimensions in Conv2D: " << xs;
    throw std::invalid_argument(s.str());
  }

If I understand correctly, when using "SAME" padding, the input size can be smaller than the filter, right? Say my input is 1x2x1, and my filter size is 1x5x1x1 with stride 1, and with "SAME" padding will give me an output of 1x2x1 instead of failing an assertion, right?

Did I miss anything?

zhisbug:
@ccsasuke Yes, you are correct. Under 'SAME' convolution it is possible the feature map is smaller than the filter. However, by convention, this constraint is usually enforced under the conv2d scenario because a conv2d operator is usually supposed to sweep the maps multiple times (also for the simplicity of the logic). In your case, I think a more straightforward way is to directly use a matrix multiplication.

@neubig What do you think?

ccsasuke:
@zhisbug Hi, I was just doing a normal 1d convolution of a sequence of texts. And the code failed on very short sentences (shorter than the filter size).

zhisbug:
@ccsasuke @neubig If this is a common case in applications and there is no easier way to avoid this, I think we should remove this constraint. I think the rest parts of the code won't be affected.

neubig:
@zhisbug Yes, removing this constraint seems like a good idea, as it fits the semantics of SAME. But let's save it for another commit. I'll merge this one.

ccsasuke:
Filed a separate issue for this. #459 

For now, I guess the workaround for me is simply remove this check when using "SAME"?

neubig:
@ccsasuke Another common thing to do is to manually pad your sentences to be larger than the filter. For example, if your filter is 3 words long, add two words to the beginning and end of each output sentence.

zhisbug:
As conv2d will only pad 0 in all cases. Will this affect the semantic of your applications? @ccsasuke

ccsasuke:
Currently I think 0 padding is fine. Maybe later I'll play with the padding (probably just by manually padding them).

PR #464, opened 2017-04-20
Add cyclical learning rate SGD trainer

pmichel31415:
Adds a fun new optimizer from [Cyclical Learning Rates for Training Neural Networks](https://arxiv.org/abs/1506.01186) where the learning rate oscillates according to : 

screenshot from 2017-04-20 17-50-30


neubig:
LGTM!

Issue #457, opened 2017-04-18
KMaxPooling mini-batch support


gth879w:
Is it right that kmax_pooling() doesn't support mini batching?
https://github.com/clab/dynet/blob/master/dynet/nodes-conv.cc#L349

If so, is it possible to add such support? In light of #388, batching in k-max pooling will be very useful. Thank you.

neubig:
Yes, this should definitely be added. This is also related to this:
https://github.com/clab/dynet/issues/158

gth879w:
For k=1, I noticed that max_dim(), which supports batching, was recently introduced. This is helpful. Thanks!

neubig:
Resolved by #457, although GPU support for kmax is still pending (#245)

PR #472, opened 2017-04-21
Add support for mini-batch and dimensionwise kmaxpooling

Peilun-Li:
This PR extends `kmax_pooling` to 3-d tensor, and allows mini-batching and dimension selection on `kmax_pooling`,  which solves #442 and #457. (However this code doesn't work on GPU either, as the original version)

Sample usage:
(Python) `expr = kmax_pooling(batch_emb, 2, 0)`
(C++) `Expression expr = kmax_pooling(batch_emb, 2, 0)`
The second prameter is k, and the third parameter is the selected dimension with default value 1 (which is the original `kmax_pooling`)

Note that for k=1, the result of `kmax_pooling` is not the same as `max_dim` since `kmax_pooling` will not reduce dimensions. For example, if input dimension is (2, 3, 4) and the selected dimension is 0, the output dimension of `max_dim` is (3, 4), while `kmax_pooling` will output dimension (1, 3, 4).


Peilun-Li:
Added! Also added some descriptions. Thanks! @neubig

Issue #470, opened 2017-04-21
Profiling for forward/backward performance of each node


neubig:
It would be really nice if DyNet had a profiling program that ran forward and backward over each of the operations and measured how long each call took. This would help us catch any operations that seem to be taking way longer than expected, or operations where the backward step is much slower than the forward step or vice-versa. I think if we write a nice templated function it should be quite easy to enumerate all of the nodes, so this might not be *too* much work, and it would be super-helpful in making DyNet more speedy.

neubig:
Fixed by https://github.com/clab/dynet/pull/1088

Issue #198, opened 2016-11-28
Test Python bindings on Travis CI


danielhers:
As far as I understand, `.travis.yml` currently only compiles the C++ library and runs C++ tests.
It should also install the Python bindings and run Python tests.
Both 2.7 and 3.5 should be [specified](https://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against) in order to test both.

neubig:
Yes, this is a good idea!

neubig:
This is done by e.g.: https://github.com/clab/dynet/commit/4127a363821f52ad095317bb14bd416e0bd93b21

PR #473, opened 2017-04-21
First attempt at python testing

pmichel31415:
This PR starts to address #198 

- Creates a new folder `tests/python` containing a file `test.py` containing tests with python's `unittest` module. For now the tests are copy pasted from the old `python_tests/test_*` files
- Removes `python_tests`
- Adds testing to Travis CI

Left for future PR : 

- Exhaustive and consistent testing of all classes/functions
- Profiling with timeit? ( #470 )


neubig:
Cool, this is great!

Re: future work. I think profiling is something very different from unit testing (profiling should be computationally intensive, and probably not run on Travis CI), so it will probably be a different program.

Issue #440, opened 2017-04-13
Better support for implicit broadcasting


neubig:
Currently DyNet supports broadcasting for all operations across mini-batches, but only supports broadcasting within tensors for the `colwise_sum` operation and the bias term for `affine_transform`. It should be made easier to implicitly broadcast things, particularly for componentwise operations like `cmult` and `cdiv`. Also `colwise_sum` should probably be renamed something like `csum` and given the same broadcasting behavior to make things consistent.

neubig:
Done by #478 and others.

Issue #355, opened 2017-03-06
about broadcasting multiplication


zhang-wen:
here i have two expressions whoses shapes are {1X80} and  {512X80} with batch size 80

i want to do pair-wise multiplication, so i should broadcast the {1X80} by row, what should i do ? when i use cmult function to multiply them, it showed following error:

```
Mismatched input dimensions in CwiseMultiply: [{1X80} {512X80}]
```

yoavg:
What is the source of these expressions? I suspect the real solution would be to create them differently and not get to this situation. (for example, perhaps the 512x80 should actually be a batch of 512 elements of size 80?)

zhang-wen:
yes, your are right, the {512x80} matrix is a hidden state of one GRU, and it is a batch of 512 elements of size 80.
about the {1x80} matrix, firstly i created a mask expression `mask_expr` which is {lenx80}, and i use `mask_expr[index]` to get the {1x80} matrix.

yoavg:
Try using the DyNet minibatching mechanism. It is a bit different than in other toolkits.
http://dynet.readthedocs.io/en/latest/minibatch.html

If this does not fit your use case, explain why and we can try and figure out a workaround.

pmichel31415:
Just adding to this I still think there are case where having `cmult` automatically broadcasting would be useful (if not necessary). Some I've encountered are : 

- Doing dropout on the embedding matrix for RNNLM, it's better to have a Bernoulli array of size {size of vocabulary,1}
- Multiplying a matrix/vector with a scalar expression (for instance : reparametrization trick when the standard deviation is a parameter)

neubig:
OK. I'm a little bit cautious of broadcasting because I think it can cause you to accidentally do things you weren't expecting if you're not very careful with it. That being said, I think it's worth having the ability to do so.

For number one in your examples, I didn't understand this completely. You mean you're setting the embeddings of some words to zero? Is there a reference for this?

For number two, we should definitely make it possible to broadcast scalar expressions, as we already have this for scalar consts, and this has many obvious use cases.

pmichel31415:
For number one see https://arxiv.org/abs/1512.05287 section 4.2

This is basically applying dropout to the words (but the same words within one sentence)

neubig:
OK, but: "For T ≪ V it is therefore more efficient to first map the words to the word embeddings, and only then to zero-out word embeddings based on their word type." We would probably just do this with an "if" statement in DyNet, right?

[...]

PR #478, opened 2017-04-25
Scalar multiplication/quotient (implicit broadcasting)

pmichel31415:
This partially addresses #355 and #440 

Now,

    cmult(x,y);

Also works if either `x` or `y` are (optionally batched) scalars

Same for `cdiv(x,y)` if `y` is scalar

Should also transfer to the python API.

Basically this creates two nodes `ScalarMultiply` and `ScalarQuotient` and the code of expressions `cmult` and `cdiv` checks the dimensions and adds the appropriate node to the computation graph.


dynet/expr.cc
```diff
@@ -47,8 +47,8 @@ Expression operator-(real x, const Expression& y) { return Expression(y.pg, y.pg
 Expression operator-(const Expression& x, real y) { return -(y - x); }
 Expression operator*(const Expression& x, const Expression& y) { return Expression(x.pg, x.pg->add_function<MatrixMultiply>({x.i, y.i})); }
 Expression operator*(const Expression& x, float y) { return Expression(x.pg, x.pg->add_function<ConstScalarMultiply>({x.i}, y)); }
-Expression cmult(const Expression& x, const Expression& y) {return Expression(x.pg, x.pg->add_function<CwiseMultiply>({x.i, y.i}));}
-Expression cdiv(const Expression& x, const Expression& y) { return Expression(x.pg, x.pg->add_function<CwiseQuotient>({x.i, y.i})); }
+Expression cmult(const Expression& x, const Expression& y) { if (x.dim().batch_size() == 1) return Expression(x.pg, x.pg->add_function<ScalarMultiply>({x.i, y.i})); else if(y.dim().batch_size()==1)return Expression(x.pg, x.pg->add_function<ScalarMultiply>({y.i, x.i})); else return Expression(x.pg, x.pg->add_function<CwiseMultiply>({x.i, y.i}));}
```
neubig: Can we make this multiple lines? Same with the following.

pmichel31415:
OK I fixed both, should be ready to merge when the tests finish.

I agree that a centralized auto formatter would be nice. I personally use the SublimeAStyleFormatter plugin for sublime text, I think that it's based on [AStyle](http://astyle.sourceforge.net/), might be worth considering

michaelcapizzi:
Thanks for implementing this, @pmichel31415.  I'm actually trying to "embedding-level" dropout as you described in the attached paper in #355 .

But I have a question about backpropogation.

If I "zero-out" an embedding after `batch_lookup()` using `cmult()`, will the error *still* get backpropogated to the `lookup` parameters for that embedding *even though* it was `dropped-out` in that instance?

pmichel31415:
The gradient will be 0 since multiplication by 0 will also be performed during the backward computation

ie if `ê = m * e` where `e` is the embedding and `m` is the dropout mask, `dL/de = m * dL/dê` where `L` is your objective.

michaelcapizzi:
Ah!  Makes sense.  Perfect.

Thank you @pmichel31415 for your quick response.

PR #480, opened 2017-04-25
update bindings for upstream changes

joelgrus:


joelgrus:
oops, I really didn't mean to merge upstream without a PR review. shouldn't hurt anything though. :smile:

neubig:
Thanks for this! By the way, it might be worth adding conv2d to the Scalar bindings? This seems like something that quite a few people would want.

PR #482, opened 2017-04-26
Layer norm

pmichel31415:
Adds layer normalization with tests + doc in c++ and python

Also had to add a `scalarAdd` node to handle scalar broadcasting for addition (and substraction), much like `scalarMult`


neubig:
Thanks! Looks good

PR #483, opened 2017-04-26
Remove boost dependency of save/load except example folder.

xunzhang:


neubig:
Thanks, we're almost there! It looks like Python is broken. Should I integrate and leave Python for the next PR?

xunzhang:
I think this can be merged first if possible. To fully remove boost dependency, remaining pull requests are to be finished.
1. Rewrite `example/cpp` folder codes related to save/load(take https://github.com/clab/dynet/pull/483/commits/9f54a6dbd2ec7e6d887d03e6c4e190d1986ee284 for example)
2. Add python binding and rewrite `example/python` folder codes.
3. Update tutorial and add legacy warning when users still use old `model` interface.
4. Remove boost usage in remaining part(mp.h etc), add boost flag to open boost usage only for unit-test, etc.

I am working on 1 now.

neubig:
Great, thanks!

xunzhang:
The latter part of `4. Remove boost usage in remaining part(mp.h etc), add boost flag to open boost usage only for unit-test, etc.` is implemented by https://github.com/clab/dynet/pull/495.

xunzhang:
1. `Rewrite example/cpp folder codes related to save/load(take https://github.com/clab/dynet/pull/483/commits/9f54a6dbd2ec7e6d887d03e6c4e190d1986ee284 for example)` is resolved by https://github.com/clab/dynet/pull/499

yoavg:
Hi @xunzhang @neubig 
could you please clarify here or in the docs what is the difference between `populate` and `load`?
(we've discussed it over email in length, but it is not clear from the current docs, and I want to make sure I understand the semantics right).

in particular, I assume that in `populate`, the `key` parameters specifies what should be populated, right?
* what happens if the key is `""`?
* what is the meaning of `model_name` in `populate`?

what does `load_param` and `load_lookup_param` do?
I assume that they load a parameter with a given names and adds it to the ParameterCollection, right?
but then:
* what is `key`? (is it the name to load? the name to populate into? both?)
* what is `model_name`?
* is there a `load` variant for ParameterCollection ?

xunzhang:
Hi @yoavg , 
Right, the key for `populate` is to specify which `ParameterCollection/Parameter/LookupParameter` should be loaded into.
1. If key is `""`, it assumes the whole file only contains one `ParameterCollection/Parameter/LookupParameter`. In other words, it will seek to the beginning of the file.
2. `model_name` is the model name for holding the wanted parameter. Since there are two types of `Parameter` in model files, one is just independent parameter being saved and the other is the parameter being saved together with the `ParameterCollection`.

xunzhang:
For the second part of the question, `load_param` is to load `Parameter` parameter with `key` into `ParameterCollection` and return the `Parameter` as return value. So,
1. The `key` is `Parameter`'s key when saving.
2. `model_name` + `key` is to specify `Parameter` inside a `ParameterCollection`.
3. Currently, only `populate` to load parameter into a `ParameterCollection`.

xunzhang:
You can find more explanation and example in the files below:
https://github.com/clab/dynet/blob/new-readwrite/dynet/io.h https://github.com/clab/dynet/blob/new-readwrite/tests/test-io.cc

yoavg:
Sorry but I still don't fully get it.

re `populate` (1): what happens if there are multiple saved things in the file, and I pass in `""`? will it get the first one?
(2): I did not understand the answer.

re `load`:
(1) let me try to rephrase: `key` is the name that we want to load from. is that correct?
(2) I don't understand what this means.

neubig:
re `populate` again, looking up my email I see that our original API example was the following:

    // Created the collection
    ParameterCollection pc1;
    Parameter p1 = pc1.add_parameter({100});
    Pack pack1(filename);
    pack1.save(pc1);
    
    // Populate (the recommended usage)
    ParameterCollection pc2;
    Parameter p2 = pc2.add_parameter({100});
    Pack pack2(filename);
    pack2.populate(pc2);
    // or using the name
    pack2.populate(pc2, pc1.get_full_name());
    // or using a single parameter
    pack2.populate(p2, p1.get_full_name());
    
    // Load (another possible usage)
    ParameterCollection pc3;
    Parameter p3 = pack3.load(pc3, p1.get_full_name())

Was there a reason why we diverged from this? If not, would it be possible to revert to just using a single name instead of model+name?

xunzhang:
The reason is for efficiency consideration to store offsets for `ParameterCollections/Parameter/LookupParameter`. It is a two-level offset hashmap for seeking.
In detail, say `ParameterCollection` `pc` contains millions of `Parameter`s. If all offsets are flat structured when loading we need to search in a large space. Maybe there is better design but currently, I use this for simplicity. Do you also feel it is tricky to understand? @neubig 
Another consideration is to avoid name confliction in `ParameterCollection`'s  `Parameter` with single `Parameter`.

yoavg:
1) I don't think we will have more than a couple hundred parameters in a model.
2) what's special about two-level lookups? what if our nesting is 7 levels deep?

xunzhang:
The first level offset is the `ParameterCollection`, `Parameter` or `LookupParameter` offset. If it is a `ParameterCollection`, there is a second level `Parameter` or `LookupParameter` offset.

yoavg:
Having looked at the code, I saw an `offset` field in Packer.

From what I understand it is used by the `save` methods, but this looks like a recipe for trouble, as it is the only state in an otherwise stateless object.

In particular, what will happen if I do:
```
Packer p("f1")
p.save(some_params, "key", true) // append

Packer p2("f1")
p2.save(another_params, "key2", true) // append
```

I think currently the save of `another_params` will write over `some_params`, because the offset in `p2` will be wrong, even though they point to the same file, and save asks for append.

can we get rid of the `offset` field and instead just seek to the end of file when asked to append?

xunzhang:
`append`'s semantics is just for save interface itself of every single packer. Currently it only supports the same packer saving data to multiple files using `packer.reinit`. It is a good point to support multiple packers handling same files. `offset` is not only used  for save methods, it is important for load/populate since it can locate where the parameters are very quick(without scanning the whole model data).

yoavg:
I am not saying not to have the concept of `offset` at all (though I do think that if we can get rid of the the .meta file it will be better), just that it shouldn't be a field in the Packer class, as there is no need for it, and and it makes it hard to support multiple packers for the same file for no reason.

what will be the downside of not having the `offset` field, and instead seeking to the end of the file whenever we need to know where to write?

yoavg:
another question, this time regarding the behavior of populate.

my understanding was that the expected usage is:
```
    ParameterCollection m;
    Parameter a = m.add_parameters({10});
    Parameter b = m.add_parameters({3,7});
    Parameter c = m.add_parameters({2});
    dynet::Packer s("bla");
    s.save(m)
    //...

   ParameterCollection m2;
   Parameter a1 = m.add_parameters({10});
   Parameter b1 = m.add_parameters({3,7});
   Parameter c1 = m.add_parameters({2});
   dynet::Packer s2("bla");
   s2.populate(m2)
   // now a1 should be equal to a, b1 to b, c1 to c
```

However, this does not seem to be the case at all: if I understand correctly, the populate method adds new parameters to the model and not reusing existing ones.

Moreover, the tests have:
```
BOOST_AUTO_TEST_CASE ( test_save_load_parameter_collection ) {
  {
    ParameterCollection m;
    Parameter a = m.add_parameters({10}, "a");
    Parameter b = m.add_parameters({3,7});
    LookupParameter c = m.add_lookup_parameters(10, {2});
    std::remove("test.model"); std::remove("test.model.meta");
    dynet::Packer s("test.model");
    s.save(m, "model1");
    s.save(m, m.get_namespace(), true);

    ParameterCollection m2;
    s.populate(m2, "model1");
    DYNET_CHECK_EQUAL(m2, m);
    ParameterCollection m3;
    s.populate(m3, "/");
  }
```
notice that this is populating the collections `m2` and `m3` that do not have any parameters in them yet.
how is one supposed to access the parameters in `m2` and `m3` then?

yoavg:
if the key cannot contain ":" this should be enforced.

Issue #476, opened 2017-04-24
Access to Input Gradients?


ccsasuke:
Hi,

I'm new to DyNet and I'm wondering whether it is possible to get the gradients w.r.t. some input (or intermediate) X?
The only API I found was to get the gradients w.r.t. parameters.

neubig:
Are you using the Python API, or C++? In C++ I believe you can get it with `expr.value().g`, but I'm not sure if this works in Python.

ccsasuke:
Thanks for the reply. I'm using the Python API. Is there a way to expose this C++ API to Python?

pmichel31415:
@neubig I don't think this works in c++ because `expr.value()` is a tensor. It would require some tweaking in `exec.h` to write a `get_gradient` function but ~~I believe it should be pretty straightforward~~ we'll see.

Opening it up to python would be easy though.

I can try to do that tomorrow

pmichel31415:
Solved by #484

PR #484, opened 2017-04-26
Add access to gradient w.r.t. arbitrary expressions

pmichel31415:
Addresses #476 

This does 2 main things : 

- add an optional `full` parameter to `backward` which, if set to `true`, forces the execution engine to compute the gradient for each node (including input node)
- Add a `gradient` method to `Expression`

All of this in c++ and python with test + doc

Edge cases : 

-  `gradient`  raises an expression if backward hasn't been computed/has been computed w.r.t. an upstream node.
- Gradients of constant nodes are 0 if `full=false` (default)


neubig:
What's the difference between a "constant" node mentioned in the edge cases and an input node?

pmichel31415:
So according to `exec.h` : 

    // here we find constant paths to avoid doing extra work
    // by default, a node is constant unless
    //   1) it is a parameter node
    //   2) it depends on a non-constant node
    // (thus, functions of constants and inputs end up being
    //  false in this computation)

I made the doc more explicit about this

neubig:
Thanks! The reason why I was confused was because of the mistake in the OP. This change looks good.

PR #485, opened 2017-04-26
Access RNN parameters from python

pmichel31415:
Addresses a complaint I've heard here and there.

Now RNNBuilders have two new methods: 

- `get_parameters`, Retrieves the internal parameters of the builder
- `get_parameter_expressions`, Retrieves the internal parameters expressions of the builder. This is more efficient because it retrieves the expressions directly if they've already been loaded in the computation graph

Paging @iamyoungjo , I think we discussed this a while back

Comes with doc and tests

### DIFF
--- a/dynet/gru.h
+++ b/dynet/gru.h
@@ -22,6 +22,12 @@ struct GRUBuilder : public RNNBuilder {
   unsigned num_h0_components() const override { return layers; }
   void copy(const RNNBuilder & params) override;
 
+  // first index is layer, then ...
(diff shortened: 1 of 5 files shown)


neubig:
Awesome, thanks! LGTM but could you fix the conflicts?

pmichel31415:
Done

jcyk:
Hi, This is so good, It provide a way to initialize RNN parameters in whatever way.
but, it does not seem only work in both GPU and CPU.

PR #490, opened 2017-04-27
Adds layer norm to vanilla LSTM

pmichel31415:
This adds a `ln_lstm` optional parameter to the VanillaLSTM which enables layer norm according to [the original paper](https://arxiv.org/abs/1607.06450)

Also adds a bunch of tests for RNNs in c++

Also contains python bindings

Performance:

I've run a super quick test on cpu for character lm. Something I've noticed is that lstm with layer norm is substantially slower (because of layer norm and the fact that the `affine_transform` has to be broken. Didn' record the time but i'd say x 1.5-2 as slow as vanilla. EDIT : I checked and it's **x 6** (on CPU) so much slower. Can't test on GPU right now though

However the results are very good

ln

The tests were done with the `example/python/rnnlm-batch.py` script


neubig:
Great, thanks!

PR #495, opened 2017-04-28
Made everything but tests/mp compile w/o boost

neubig:
Adds a new `INCLUDE_BOOST` definition for cmake, which is disabled by default. Tests and MP are only compiled if it is enabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ before_script:
   - cd build
   - conda create -q -n dynet-environment python=$PYVER $CONDA_PACKAGES
   - source activate dynet-environment
-  - cmake .. -DEIGEN3_INCLUDE_DIR=/usr/local/include/eigen3 -DPYTHON=`which python`
+  - cmake .. -DEIGEN3_INCLUDE_DIR=/usr/local/include/eigen3 -DINCLUDE_BOOST -DPYTHON=`which python`
 
 after_failure:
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -75,7 +75,7 @@ before_script:
   - cd build
   - conda create -q -n dynet-environment python=$PYVER $CONDA_PACKAGES
   - source activate dynet-environment
-  - cmake .. -DEIGEN3_INCLUDE_DIR=/usr/local/include/eigen3 -DPYTHON=`which python`
+  - cmake .. -DEIGEN3_INCLUDE_DIR=/usr/local/include/eigen3 -DINCLUDE_BOOST -DPYTHON=`which python`
```
xunzhang: Use `-DINCLUDE_BOOST=ON` instead, `ON` can also be `On` or `on`.

CMakeLists.txt
```diff
@@ -179,4 +177,7 @@ if(INCLUDE_SWIG)
   add_subdirectory(contrib/swig)
 endif(INCLUDE_SWIG)
 
-enable_testing()
+if(INCLUDE_BOOST)
```
xunzhang: I like this! 😎

dynet/str-util.h
```diff
@@ -6,16 +6,16 @@
 
 namespace dynet {
```
xunzhang: I'll follow this file naming style in the future.

neubig:
Thanks for the review!

Issue #493, opened 2017-04-28
pick_batch bug/unclear behaviour


pmichel31415:
I'm trying to do the same thing as `lookup_batch`, but on a parameter.

I thought 

    lookup_batch(lookup_param, l) == pick_batch(param_expression, l)

But this gives me an error : 

    In PickElement::forward, number of elements in the passed-in index vector (5) did not match
    number of elements in mini-batch elements in expression (of dimension{64})

After looking a little bit in the code it seems that `pick_batch` retruns something with the same batch dimension as the input (https://github.com/clab/dynet/blob/master/dynet/nodes-common.cc#L646-L648).

Is this an intended behaviour? What does pick_batch (or in c++ pick with a vector of indices) do? The doc is very vague.

And if this is intended, is there any way for me to do what I want? ie "lookup" from a parameter?

neubig:
At the moment, `pick_batch` only supports picking when the batch dimension of the expression is equal to the number of elements in the passed-in list of indices. However, this is easy to fix.

pmichel31415:
Hmm, so I guess what I want to do would best be implemented in a separate `lookup(expression x, int i)` expression right?

Or add broadcasting to pick (basically what you said)?

neubig:
I was thinking of adding broadcasting to pick.

pmichel31415:
OK I'll give it a try

neubig:
Fixed by: https://github.com/clab/dynet/pull/497

PR #497, opened 2017-04-28
Add broadcasting along the batch dimension to pick (python pick_batch)

pmichel31415:
Addresses #493


neubig:
LGTM pending tests passing.

PR #499, opened 2017-04-29
Refactor code without using boost serialization in example/cpp folder

xunzhang:
Since current packer only supports mix Parameter, LookupParameter, and ParameterCollection save and load, I removed builder and user-defined class serialization/deserialization functionality in `example/cpp` folder.

### DIFF
--- a/dynet/model.cc
+++ b/dynet/model.cc
@@ -3,6 +3,7 @@
 #include "dynet/aligned-mem-pool.h"
 #include "dynet/dynet.h"
 #include "dynet/param-init.h"
+#include "dynet/io.h"
 
 #include <unordered_set>
 #include <iostream>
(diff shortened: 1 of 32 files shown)


examples/cpp/mnist/mlp.h
```diff
@@ -19,10 +19,6 @@
 
 #include <boost/serialization/utility.hpp>
```
neubig: Still have boost here?

dynet/model.cc
```diff
@@ -425,19 +426,15 @@ const ParameterCollectionStorage& ParameterCollection::get_storage() const {
   return *storage;
 }
 
-/*
 void save_dynet_model(std::string filename, ParameterCollection* model) {
-  std::ofstream out(filename);
-  boost::archive::text_oarchive oa(out);
-  oa << (*model);
+  Pack packer(filename);
```
neubig: Looking at the code, maybe this should be called Packer instead of Pack?

PR #502, opened 2017-05-01
Added multibatch support for euclidean distance

neubig:
This will increase efficiency on GPU by not calling the function for each minibatch element.


pmichel31415:
Is it possible to add this to squarednorm as well?

neubig:
Sure, done. Norms are easier than distances :)

pmichel31415:
Thanks!

PR #507, opened 2017-05-02
Fixing array initialization to be C+ conformant.

mattr1:
And to match initialization done elsewhere in the code. And this way it compiles in MSVC :).


neubig:
Thanks!

PR #513, opened 2017-05-03
Allow manual scaling of the gradient

pmichel31415:
Using `Parameter.scale_gradient(s)` to scale an individual gradient is now possible (same for lookup parameter)


neubig:
Thanks!

PR #526, opened 2017-05-05
Add dropout along one dimension

pmichel31415:
This allows to drop along one dimension, ie to use the same dropout mask along this dimension.

This is useful if one wants to drop rows/columns in a matrix for instance.

I also reactivated tests for the noise operations, right now the tests just run a forward pass to detect bugs.

Tests+doc+python bindings


pmichel31415:
@neubig Can I merge this now?

neubig:
LGTM!

PR #532, opened 2017-05-09
Dropout batch elements

pmichel31415:
Same as dropout_dim but along batch dimension.

Will merge after the test complete


neubig:
Just out of curiosity, what is the use case for this?

pmichel31415:
When decoding in neural MT, if you want to drop out the input words altogether (but you want to drop different words for each batch).

neubig:
Ahh, perfect! Got it.

PR #534, opened 2017-05-09
bring scala bindings up to date

joelgrus:
I don't think this needs a code review, but I'll wait for the CI to run so it doesn't end up with a big X by it. :)

(1) get rid of bindings for `Tensor::bs` (which doesn't exist anymore)
(2) add SWIG + Scala bindings for `conv2d`


neubig:
Yeah, it looks fine with me. Feel free to merge at your discretion, along with https://github.com/clab/dynet/pull/533

joelgrus:
ha, I guess he fixed the same bug

PR #540, opened 2017-05-12
minor fixes to example python code

iceiony:
xor.py - division in python3 turns int to float leading to incorrect xor training example 
cpu_vs_gpu.py - segmentation fault occurs when importing _dynet directly and not calling init()


neubig:
Thanks a lot!

Issue #538, opened 2017-05-11
LSTM without update


jihelhere:
Hi,

I can't find a way to use an LSTM in a deep architecture without updating its parameters during backprop (for instance in a GAN scenario where an LSTM is used as a discriminator).

if RNN/LSTM methods new_graph/new_graph_impl could take an additional boolean flag, 
then internal parameters could be passed as const_parameter if needed.

neubig:
The easiest way to do this is to add a `nobackprop()` operation between the LSTM and whatever you're trying to use it to calculate. Will this solve your problem?

jihelhere:
I'm not sure,

I would like error to backpropagate through the LSTM down to the lower layers of the architecture as usual. I just don't want the lstm parameters to be updated when the trainer calls update.
IIUC, nobackprop would prevent the error to propagate down the network.

pmichel31415:
This should be rather easy to add.

I don't have a lot of time to do it right now though, but if you want to contribute here's what needs to be done : 

- in file `dynet/rnn.h` : change the interface of `RNNBuilder::new_graph` to take an additional optional boolean parameter `update`, by default set to `true`
- in files `dynet/rnn.*`, `dynet/lstm.*`, `dynet/gru.*`, `dynet/fast-lstm.*`, `dynet/deep-lstm.*` : update the interface to add the new parameter and change the `new_graph_impl` method to use `const_parameter` instead of `parameter` depending on the `update` parameter ([here for the lstm for example](https://github.com/clab/dynet/blob/master/dynet/lstm.cc#L57-L69))
- Update the documentation (doxygen comments)
- Optional : add tests in `tests/test-rnn.h`
- Update the python bindings : 
    - Change [this line](https://github.com/clab/dynet/blob/master/python/_dynet.pxd#L418) to take the new parameter into account
    - And [this function](https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L3316-L3318) and [these](https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L3404-L3451)

This would be super helpful!

If you don't have time I'll do it but not before the end of next week.

jihelhere:
Thanks!

I can do this next week.

jihelhere:
Working on the code, I wonder if this could be the opportunity to refactor the `parameter`  method so it would take an additional parameter to signify whether it's "trainable" or not. In this case `const_parameter` would be just a wrapper that sets this boolean to false.
 
[...]

PR #547, opened 2017-05-16
Lstm const : add flag to control update of RNN internal parameters

jihelhere:
Implements feature discussed in #538

The fixup  maintains lines in python files ending with whitespace, as in master.


pmichel31415:
Thanks, that was fast!

LGTM, @neubig if you have nothing to say I will merge this afternoon

PR #550, opened 2017-05-17
Add Eigen note for sudo-less installation

john-hewitt:
Using/Installing DyNet without sudo is possible if one does not have to install Eigen. 
This was of use to me, and may be of use to others using academic grids with possibly slow sysadmin response.
Eigen does not need to be installed; the repository need only be cloned.
As for testing, I at least, have not run into any problems with this.

https://eigen.tuxfamily.org/dox/GettingStarted.html/

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -28,21 +28,12 @@ get Boost, CMake, and Mercurial with either homebrew or macports:
 To compile DyNet you also need the `development version of the Eigen
 library <https://bitbucket.org/eigen/eigen>`__. **If you use any of the
 released versions, you may get assertion failures or compile errors.**
-If you have sudo access and don't have Eigen installed already, you can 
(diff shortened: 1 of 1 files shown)


neubig:
Hmm, I've never used sudo to install DyNet... We should probably remove all mention of sudo from the docs (I don't know when this was added).

john-hewitt:
It's only a problem on linux; `make install` attempts to create, e.g., `/usr/local/include/eigen3`, which requires sudo. We could also consider just removing the eigen build instructions and leaving that code block with only the `hg clone`.

neubig:
@john-hewitt Yep, that's what I meant. I don't think there's any reason to do `make install`.

john-hewitt:
Updated the PR, removing my additions and the build/install code snippet.

neubig:
Thanks!

PR #551, opened 2017-05-18
fixed moment_batches() in case of higher dimensional inputs

msperber:
including a unit test that failed prior to the fix


pmichel31415:
Thanks for this!

I have one comment though, I think instead of using `tb<3>` we should use `tbvec` for the sake of generality (same for `t<3>` => `tvec` and `red_axis[0] = 1`)

msperber:
Thanks for the hint, I updated it.

pmichel31415:
Cool thanks!

PR #557, opened 2017-05-24
Support installation through pip

danielhers:
With this change, DyNet can be installed with the following command line:

    pip install -e git+https://github.com/clab/dynet#egg=dynet

If Boost is installed in a non-standard location, it has to be set in the environment variable `BOOST` prior to installation.

To try this out from my fork before merging the pull request, use:

    pip install -e git+https://github.com/danielhers/dynet#egg=dynet


neubig:
Thanks so much for this! We're reviewing this now and will merge in the next couple days if it seems to be working.

xunzhang:
So great! I will take a look and try this soon.

Abasyoni:
I have two comments here:

First, the pip install command didn't go through, as it only works on g++-4.8 or g++-4.9 and I only have version 5.4, SO I replaced "g++-4.8 or g++-4.9" with simply "g++" and the pip install command worked and printed "Successfully installed dynet".

Second, after it printed this message, I tried to run "python examples/python/xor.py" but it didn't compile saying:
"File "xor.py", line 1, in <module>
    from dynet import *
ImportError: No module named dynet"

is this expected or did I do something wrong?

danielhers:
For me the only versions that worked were 4.8 and 4.9, but if other
versions works then I'll add what you said.
I don't know why it didn't find it though... For me it does.

Abasyoni:
Hi,

I appreciate your prompt response! I look forward to using your pip
addition, so please let me know if it's fixed.

Best,
Al

xunzhang:
Hi after successfully running `pip install -e git+https://github.com/danielhers/dynet#egg=dynet`, I still can not import dynet, my environment is osx 10.12.

BTW, is there possible way to specify GPU or CPU version while building from source?

neubig:
I think it would be reasonable to compile with GPU support if cuda is available, and with only CPU support if it is not.

xunzhang:
Good point.

Abasyoni:
@xunzhang I get the same thing where:
pip install -e git+https://github.com/danielhers/dynet#egg=dynet
succeeds but still can not import dynet. My environment is Ubuntu 16.04

danielhers:
It mostly works now, but the problem is with `libdynet.so`: it's created in the build directory, which is deleted by `pip`. Need to copy it to the installation directory and to set `runtime_library_dirs` to it.

danielhers:
Fixed the shared library problem. @xunzhang @Abasyoni I believe it should work now.

danielhers:
Things I haven't done:
- Compile with GPU support if CUDA is available (not sure how to check that)
- Add check to Travis CI that installation with `pip` works (it would have to replace the existing installation in Travis CI, I guess, and I don't know if it's OK)
- Change the `make install` prefix (and the `runtime_library_dirs`) to be inside the package egg directory rather than at the top of the virtualenv or system path.
- Add version information to allow tracking upgrades.
- [Upload to pypi](http://peterdowns.com/posts/first-time-with-pypi.html) to allow installing by just package name (not git URL).

danielhers:
Note the command line without `-e` should be used to install:

    pip install git+https://github.com/danielhers/dynet#egg=dynet

danielhers:
@neubig Just to clarify, I think this should already be quite useful as it will work for the majority of users. I'll be happy to make changes that you think may be required.

neubig:
@danielhers Thanks! Let me try it out now that it's stable.

xunzhang:
Hi, it still doesn't work for me. I got the error like below:
```
(sandbox) xunzhang@Regulus:~/sandbox$ pip install git+https://github.com/danielhers/dynet#egg=dynet
Collecting dynet from git+https://github.com/danielhers/dynet#egg=dynet
  Cloning https://github.com/danielhers/dynet to /tmp/pip-build-_lb0Fx/dynet
Requirement already satisfied: cython in ./lib/python2.7/site-packages (from dynet)
Requirement already satisfied: numpy in ./lib/python2.7/site-packages (from dynet)
Installing collected packages: dynet
  Running setup.py install for dynet ... error
    Complete output from command /home/xunzhang/sandbox/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-_lb0Fx/dynet/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-OCnYPn-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/xunzhang/sandbox/include/site/python2.7/dynet:
    running install
    running build
    INFO:root:==============================
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-_lb0Fx/dynet/setup.py", line 184, in <module>
        cmdclass={"build": build},
      File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
        dist.run_commands()
      File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
        self.run_command(cmd)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/home/xunzhang/sandbox/local/lib/python2.7/site-packages/setuptools/command/install.py", line 61, in run
        return orig.install.run(self)
      File "/usr/lib/python2.7/distutils/command/install.py", line 601, in run
        self.run_command('build')
      File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
        self.distribution.run_command(command)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/tmp/pip-build-_lb0Fx/dynet/setup.py", line 76, in run
        log.info("CMake path: " + self.cmake_path)
    TypeError: cannot concatenate 'str' and 'NoneType' objects
    
    ----------------------------------------
Command "/home/xunzhang/sandbox/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-_lb0Fx/dynet/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-OCnYPn-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/xunzhang/sandbox/include/site/python2.7/dynet" failed with error code 1 in /tmp/pip-build-_lb0Fx/dynet/

```

danielhers:
Thanks, @xunzhang. It seems like you don't have `cmake` installed, or it is not found on the path (but you can set `CMAKE` to where it is before installing). I'll improve the error message.

xunzhang:
Great! It works now. Thank. +1 for this

neubig:
Sorry about just getting around to trying this now, and basically it seems to be working for me. Thanks so much for this, it's great! Just 3 final things and I think it's ready to merge.

1. Compiling takes a long time, so users might think install has frozen. Could we print a message saying that this may take up to X minutes or something?
2. Is there any way we can compile on all the available processors instead of just using one? For example, on Linux we can get the number of processors using `nproc` and pass it as an argument to the "make" command.
3. Could we add this to the top of the Python doc as the recommended method for installing the DyNet Python bindings? The current method can be listed as an alternative method.

danielhers:
1. Generally the setup output should be succinct (as it is for other packages, even ones with long build time), but you can pass `-v` to `pip` to see all the output that you usually see when building dyNET. In any case, later (maybe once Boost is no longer required) a pre-compiled wheel can be upload to PyPI (per architecture/OS/Python version), and installation will be much faster.
2. Added `-j` with the number of processors (available using Python API).
3. I'll update the documentation too.

neubig:
Awesome, I think this is about ready to merge. Will do so tomorrow-ish if there are no objections. Thanks a bunch!

danielhers:
I went ahead and added the necessary files for PyPi too. This works with the PyPI test repository:
`pip install -i https://testpypi.python.org/pypi dynet`
I did not upload to the real PyPI, though. Before that, I think an official release should be made and then the version number should be set as the `version` argument to `setup()` in `setup.py`.

(Note: it's not the pre-built library yet; it's just to allow installing with a simple `pip install dynet`.)

PR #560, opened 2017-05-24
Autobatch test failing on GPU

neubig:
For some reason auto-batching has regressed and is failing tests on GPU. This fixes one bug, but there are still a few tests that are failing like the `vanilla_lstm` tests. These should be fixed ASAP. Autobatching is working fine on CPU, so it's likely something in memory copy or something.


neubig:
This fixed a couple bugs introduced by accumulating gradients on parallel on GPU. Things should be working properly now.

PR #564, opened 2017-05-25
print fw timing info on debug output

yoavg:
This adds the ability to measure and print out the cumulative time spent on different node types when using forward, on either the autobatch or non-autobatched executors. It is helpful to (a) see which ops benefit most from autobatching on different platforms (b) helps finding bottlenecks and slow ops in general.

### DIFF
--- a/dynet/exec.cc
+++ b/dynet/exec.cc
@@ -87,7 +87,7 @@ const Tensor& SimpleExecutionEngine::incremental_forward(VariableIndex i) {
       dev->pools[(int)DeviceMempool::FXS]->free();
 
   if (i >= num_nodes_evaluated) {
-    NamedTimer timer; string current_node_name;
+    string current_node_name;
     nfxs.resize(i + 1);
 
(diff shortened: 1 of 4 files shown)


neubig:
This looks good, and I think we can merge if you'd like. One thing it might be worth thinking about is whether we can make this do aggregate statistics instead of per-execution statistics. For example, make the NamedTimer global, then have it print statistics when its destructor is called?

yoavg:
Sure, I can make it aggregate across executions. I kinda liked it this way because it also allows you to isolate different calls to forward in an incremental setup, but it's not a strong preference.

neubig:
OK, cool. We could also get the best of both worlds by making it possible to reset/print from the global NamedTimer on the user's side, so the user could choose to print/reset after each call to forward.

yoavg:
where would be a good place to put the global timer?

neubig:
`globals.{h,cc}`

PR #570, opened 2017-05-26
Fixed bad autobatch signature for LookupParameters

neubig:
There was a bug in the signature for lookup parameters that would cause problems when using multiple lookup parameters with the same dimension in a single computation graph. This commit should fix the problem.


yoavg:
Will this hold after the move from Model to nested ParameterCollections (not sure how indexing is implemented there)?

neubig:
Maybe not. We might have to reference the pointer to the lookup parameters instead of the index then.

Issue #572, opened 2017-05-29
LookupParameter related error when autobatching on CPU with Scala bindings


shuheik:
Hi,

I'm experimenting with dynet's scala bindings and encountered an error that looks similar to #567 .

Autobatching seems to work well with scala bindings on GPU, just by adding autobatch flags to corresponding part of dynet_swig.i and specifying them in Initialize.initialize().
It gives around 6 times speedup for a BiLSTM tagger with character-level LSTMs and an Attention-based Encoder-Decoder on GPU.

The tagger benefits from autobatching also on CPU, but when I run the encoder-decoder with autobatching on CPU, it crashes with an error message like this.

> [error] (run-main-0) java.lang.IllegalArgumentException: Out-of-bounds attempt to access index 20006 for LookupParameter of size 19423
> java.lang.IllegalArgumentException: Out-of-bounds attempt to access index 20006 for LookupParameter of size 19423
>         at edu.cmu.dynet.internal.dynet_swigJNI.ComputationGraph_forward(Native Method)

When I modify the encoder-decoder code and remove LookupParameters from the decoding part, it doesn't crash even with autobatching on CPU.

I cloned the latest commit (#571), modified dynet_swig.i and built the binaries including bindings, so it's after the fix of LookupNode::autobatch_sig in #570 .

neubig:
Hmm, sorry about the bug... If you can provide an example of code and data that cause the bug to happen it'd be easier to debug. If this wouldn't be easy, I have a feeling that because multiple people are running into this problem I can probably reproduce it myself if I try hard enough, so I'll try to do so.

shuheik:
I still observe this bug when experimenting with real data from europarl and I'm now trying to reproduce this bug with minimal codes and a synthetic data. But autobatching on CPU hasn't crashed on synthetic data so far.
I'll work on reproducing the bug with small codes and data tomorrow, and if I don't succeed I'll provide the "real" code and data.

shuheik:
I could reproduce the error with a toy code.
[codes.zip](https://github.com/clab/dynet/files/1037612/codes.zip)

It turned out that this error comes out when I'm using two LookupParameters that belong to separate Models like,
>    val model = new Model()
>    val anotherModel = new Model()
>
>    val inputLookup = model.addLookupParameters(inputVcbSize, Dim(embeddingSize))
>    val outputLookup = anotherModel.addLookupParameters(outputVcbSize, Dim(embeddingSize))

and auto-batching on CPU.
[...]

PR #575, opened 2017-05-31
Fixed corner case of identical ID different lookup

neubig:
Fixes https://github.com/clab/dynet/issues/572 (probably)


shuheik:
This update solved the problem #572 for me. Thanks a lot for the quick fix!

PR #578, opened 2017-06-06
Add autobatch and autobatch_debug flags

shuheik:
These changes enable autobatching with scala bindings.


neubig:
Great, thanks!

PR #581, opened 2017-06-08
Small fix for installation with pip

danielhers:
Testing the recently merged #557 with another project on Travis CI, it failed installation because it has an old version of CMake, which looks for the `eigen` directory in the wrong place. Setting it to the full path resolves the problem.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ class build(_build):
         self.make_path = os.environ.get("MAKE") or find_executable("make")
         if not self.make_path:
             raise DistutilsSetupError("`make` not found, and `MAKE` is not set.")
+		self.make_flags = os.environ.get("MAKE_FLAGS", "-j %d" % cpu_count()).split()
         self.hg_path = find_executable("hg")
         if not self.hg_path:
             raise DistutilsSetupError("`hg` not found.")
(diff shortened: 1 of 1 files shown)


danielhers:
Another thing I noticed is that using all available processors on Travis CI exhausts the machine's memory and compilation fails. Perhaps we could go back to using just one process?

xunzhang:
Hi, I think the current setting for Travis CI should work after this fix using multiple processes, right? Check out this trigger https://travis-ci.org/clab/dynet/builds/240694893.

danielhers:
DyNet's Travis CI settings are not installing with `pip` yet. I tried it in another project: https://travis-ci.org/danielhers/tupa/jobs/240688540#L686
It would be a good idea to test installation using `pip` in DyNet's Travis CI settings, but I haven't come round to it yet.

xunzhang:
Got it. Well, let's merge this fix first to make sure the upstream code pass the CI. For adding pip install in Travis, I don't think it is 100% necessary since we only need to test this before releasing a new version. Pip install is supposed to be used for non-developers I think.

danielhers:
By "upstream code" you mean clab/dynet? It already passes CI (since it's not using pip). And even with this PR merged, it won't pass CI if using pip on it, due to the memory problem I mentioned. But @neubig asked for utilization of all processors, so I'm hesitant about removing it.
And since DyNet is not on PyPI, installation using pip must be of the git master. When a new version is released we can add it to PyPI and then this won't be necessary.

neubig:
@danielhers I'm not exactly sure what to do in this case. Perhaps we can switch off the "compile on all processors" thing for now (although pip installation will be horrifically slow, especially when using GPU). Another option would be to have an option to specify the number of processors during a pip install (I'm not sure whether this is possible), and set it to 1 in our travis config files. I'll leave the decision up to you.

danielhers:
@neubig Now the default is to use all processors, and the environment variable `MAKE_FLAGS` can be set to the empty string to disable this, or, for example, to "-j 2" etc.

neubig:
Looks good, thanks!

PR #589, opened 2017-06-14
Add Maxpooling2D CPU (eigen based) and GPU (cudnn based) version. Also fix the conv2d memory allocation issue

### LINKED ISSUE #528: Out of memory error in conv2d
When using conv2d on large matrices I get this error

    terminate called after throwing an instance of 'std::runtime_error'
    what():  aux_mem_pool allocate memory failed: exceed maximally allowed size

aren't dynamic memory pools supposed to take care of these errors?

zhisbug:
This should fix #528 and also concludes #229 .


zhisbug:
However, the CPU backward computation of maxpooling2d could be improved by parallelizing over the batch dimension using multiple computation threads.

neubig:
Looks good, thanks! I'll do some quick testing and merge if everything works

neubig:
Seems to be working for me, thanks a lot!

PR #591, opened 2017-06-14
Change Travis CI matrix.include to build matrix

danielhers:
Avoids repetition in .travis.yml, but does the same thing as before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,57 +1,48 @@
 language: cpp
 
+os:
+  - linux
+  - osx
+compiler:
+  - gcc
+  - clang
+dist: trusty
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - boost-latest
+    packages:
+      - gcc-4.8
+      - g++-4.8
+      - libboost-filesystem1.55-dev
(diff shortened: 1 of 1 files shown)


danielhers:
It will be easier to test `pip` installation with this structure. I tried [experimenting with it](https://github.com/danielhers/dynet/blob/travis_pip/.travis.yml) but it's quite ugly, with lots of `if`s.  Maybe all the "manual" installation steps could be extracted to a script?

neubig:
This change looks good on its own. Regarding your linked code, this looks fine as well. We could extract stuff to a script, but then it'd be harder to read/modify the travis configuration on its own, so I'd favor adding a bit of complexity to the travis config file rather than split over multiple files.

danielhers:
Thanks.
The problem with an `if` is that if a command inside it returns a non-zero exit code, this won't fail the build (the exit code in an `if` is ignored). There are [solutions](http://steven.casagrande.io/articles/travis-ci-and-if-statements) to this problem, though, so maybe I'll try looking into it.

By the way, I'm successfully using DyNet pip installation in my project now: https://travis-ci.org/danielhers/tupa/builds/242788506

PR #592, opened 2017-06-15
Made some improvements to doc to make them stand-alone for Python

neubig:
The requirements for installing Python were previously not clear in the doc for the pip install. This commit should fix this (note to: @danielhers).


danielhers:
Right! It's very clear now.

PR #593, opened 2017-06-15
Updated the install doc for windows

neubig:
Previously there were two problems with the Windows doc:
* In the C++ section the doc was hidden so low in the Hierarchy that it didn't appear on the sidebar by default.
* In the Python section, there was a mention of "install using cmake as above", but the cmake directions above were the ones for Mac.
This commit should make things easier to understand.

(note to @mattr1)

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -25,6 +25,8 @@ get Boost, CMake, and Mercurial with either homebrew or macports:
     brew install boost cmake hg  # Using homebrew.
     sudo port install boost cmake mercurial # Using macports.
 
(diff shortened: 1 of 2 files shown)


mattr1:
Thanks @neubig, looks good. Note that many people use Anaconda in Windows as well. It's possible that dynet works well with Windows anaconda and we could unify the python instructions. 

Also, there are a few other items in the Windows documentation that I've been meaning to clean up. I'll get those into a PR.

As a side note, it took me a while to understand this sentence in your changes: "But if you would like to install them through Anaconda and encounter build errors, try...". I thought you were being funny and saying "If you would like to encounter build errors..." :)

PR #596, opened 2017-06-15
Test installation with pip on Travis CI

danielhers:
It doesn't work on OSX so I excluded it for now.


neubig:
OK, I'll go ahead and merge this, thanks! We should aim to get MacOS support eventually though.

PR #605, opened 2017-06-19
Fix contract batch

pmichel31415:
Adds explicit batching to contract_3d_1d, with CUDA implementation. 10x faster on GPU


neubig:
This commit removes `-march=native` from `CMakeLists.txt`, which will have a negative impact on performance in many cases. Could you please add it back? Other than that, looks good!

pmichel31415:
OK should be fixed now. I also added the flag that prevents compilation on GPU by default because it takes too long (in nodes-contract.cc)

neubig:
Thanks!

Issue #603, opened 2017-06-19
ELU and SELU


speedcell4:
How can I use [ELU](https://arxiv.org/pdf/1511.07289) and [SELU](https://arxiv.org/abs/1706.02515)?

miradel51:
see readme

neubig:
These are not implemented in DyNet yet, but we'd certainly welcome any contributions. Take a look here about information on contributing new nodes: http://dynet.readthedocs.io/en/latest/code_style.html#coding-tips

speedcell4:
@miradel51 could you please show clearly which README you are referring to?

pmichel31415:
Solved by #607

PR #607, opened 2017-06-19
Adds ELU and SELU

pmichel31415:
Addresses #603


neubig:
Looks good. Compile on clang is taking forever, but go ahead and merge if you'd like.

PR #608, opened 2017-06-20
Use a release version of Eigen for building Python package.

bskaggs:
This request uses a tarball release of Eigen downloaded from bitbucket rather than an hg checkout.  By using a released version, we can remove the need for having hg on the system and we get faster downloads. We also get repeatable builds that don't depend on the head of the Eigen repo.

If you want to override default download source, specify the url as the environment variable `DYNET_EIGEN_URL`.  If instead you want to use an already downloaded and extracted version of EIGEN, set `DYNET_EIGEN_PATH`.

This has been tested for CPU on Python 2/3 on Linux; other platforms should be tested.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -200,6 +200,7 @@ class build(_build):
         log.info("MAKE_FLAGS=%r" % " ".join(MAKE_FLAGS))
         # log.info("HG_PATH=%r" % HG_PATH)
         log.info("EIGEN3_INCLUDE_DIR=%r" % EIGEN3_INCLUDE_DIR)
+        log.info("EIGEN3_DOWNLOAD_URL=%r" % EIGEN3_DOWNLOAD_URL)
         log.info("CC_PATH=%r" % CC_PATH)
         log.info("CXX_PATH=%r" % CXX_PATH)
         log.info("SCRIPT_DIR=%r" % SCRIPT_DIR)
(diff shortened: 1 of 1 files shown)


bskaggs:
The Eigen release is 3.3.4, which was released 2017-06-15, so hopefully it's recent enough to use instead of a specific commit; the bitbucket diff comparing them is [here](https://bitbucket.org/eigen/eigen/branches/compare/346ecdb%0D3.3.4#diff).

neubig:
Thanks for this! Overall, I think this is a good idea, but there's a chance that at some point we'll again need bleeding-edge capability that is not included in the Eigen release version. To be ready for that case, could you make sure that it is still possible to revert back to the old way of doing things with mercurial? This could be as simple as leaving the code there, but commented out.

setup.py
```diff
@@ -80,11 +87,17 @@ def run(self):
         self.py_executable = py_executable
         self.py_version = py_version
 
+        self.eigen_path = os.environ.get("DYNET_EIGEN_PATH", None)
```
danielhers: The `None` is redundant here (it's the default).

danielhers:
The nice thing about mercurial is that if you already have eigen it will just update it.
How about checking if the unpacking destination already contains eigen, and in that case skipping the fetch & unpack?
Also, shouldn't the manual installation instructions be the same for consistency?

danielhers:
I think this is a useful alternative if `hg` is not available. Just need to update it following the changes to `setup.py`.

PR #611, opened 2017-06-20
Fix compilation bug with boost 1.59

FilippoC:
BOOST_VERSION is not defined if boost/version.hpp is not directectly imported with my version of boost.


neubig:
Thanks!

Issue #594, opened 2017-06-15
Bug in conv2d test case on GPU


pmichel31415:
I ran `test-nodes.cc` with a CUDA/CuDNN enabled version of dynet to check on something else and I noticed the following error:

    ~/dynet/tests/test-nodes.cc(1245): error: in "node_test/conv2d_same_gradient": check check_grad(mod, z, 0) has failed

@zhisbug any ideas?

zhisbug:
I am aware of this issue (as I wrote the test case by myself). However, I compared my result with TensorFlow (with same input) -- they are consistent. You can increase the verbosity level of grad check function and see the printout. I suspect it is caused by some numeric issues in the grad_check function.

neubig:
If that's the case, one way we could fix this is by ensuring that the inputs to the `conv2d` test are large enough to make sure that the gradients don't suffer from numerical instability. We should definitely check this though. The conv2d operation doesn't seem inherently instable, and because tests are passing on CPU I think it's unlikely that it would be a problem that only appears on GPU. Of course I may be wrong.

zhisbug:
I think I have figured out the problem. The gradient computation is correct (no worry!). The test case does not pass on GPU because some irrelevant (and unused) parameters are overwritten (we create 5 model parameters in the test-node.cc code, some of them are not used when running this test case). It might be caused by some wild pointers but I am not sure at this point. I will have a fix tomorrow.

pmichel31415:
good to know! Yeah I found it weird, especially since the GPU implementation is straight from cudnn

zhisbug:
After 2 days' debugging, I ended up with the following conclusion. This is actually not a bug. This is caused by cuDNN's non-determinism. 

I run the following simple test: I feed the same feature maps and filter parameters into cuDNN and call forward (or backward) for multiple times, the output by different calls will be slightly different (no matter I launch DyNet multiple times or have multiple calls using a for loop within one launch). This explains why the approximated gradients of some irrelevant parameters will be non-zero sometimes (i.e. grad_check not pass.).

This is also a known issue in other communities. Fortunately, I haven't seen any report saying this will affect the statistical convergence.

zhisbug:
Beside this known issue, I will have a PR to fix some other issues (I found during debugging).

neubig:
OK, sounds good. Good to know the problem at least. Perhaps we can add a test that doesn't rely on the gradient checking.

PR #612, opened 2017-06-20
Enhance the conv2d and maxpooling2d code a bit, also fix two bugs.

zhisbug:
#594 

The commit refactors the conv2d and maxpooling2d code a bit. It also resolves two issues: 

- Allow the backward_impl function to execute independently of the forward function, so that you can call backward_impl function without calling foward_impl first (as long as you have a feasible dy)

- The gradients shall be accumulated instead of being assigned. The CPU implementation accumulates them properly but the GPU implementation not. This commit fixes that.

### DIFF
--- a/dynet/cudnn-ops.cu
+++ b/dynet/cudnn-ops.cu
@@ -9,12 +9,11 @@
 namespace dynet {
 
 CudnnConvOp::CudnnConvOp(const std::vector<unsigned>& s, const bool padding_type) {
-  stride.resize(s.size());
-  for (unsigned i = 0; i < stride.size(); ++i) {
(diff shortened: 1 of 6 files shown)


pmichel31415:
What's the advantage of the 1st point? I thought dynet always called forward (eg. https://github.com/clab/dynet/blob/master/dynet/exec.cc#L141-L142) in backward

zhisbug:
In some cases where you might just want to compute the gradients dx given x, y, dy (without computing y one more time).... Also, this allows us to rewrite the aux_mem during backward.

neubig:
Thanks! We discussed memory usage, but I think this can be done in a separate commit.

PR #615, opened 2017-06-21
Add some functionalities to Scala bindings

shuheik:
These commits
1. add AdadeltaTrainer, CyclicalSGDTrainer, FastLstmBuilder, Expression.elu/selu
2. modify some default arguments in Trainer to match c++ version


neubig:
Thanks!

PR #618, opened 2017-06-21
Fix an edgecase of glorot initialization

pmichel31415:
This is a really small change but has a potential impact on a lot of models so I'm putting it as a pull request.

By default dynet initializes parameters using glorot init like this:

4kwe5t

which is technically not correct if W is not a matrix. This commit changes it to
image

In practice this reduces the variance of bias parameters by 2. I've noticed big improvements in eg translation and language modelling in both score and convergence speed.


neubig:
Nice catch, thanks!

PR #628, opened 2017-06-25
Support static libraries of boost.

draplater:
CMake provides Boost_USE_STATIC_LIBS option to make the program
link to boost statically. However, in the current version of
DyNet, BOOST_TEST_DYN_LINK option is set unconditionally. It
causes problem when Boost_USE_STATIC_LIBS is on.

This patch adds condition when setting BOOST_TEST_DYN_LINK. Now
DyNet can statically link to boost with Boost_USE_STATIC_LIBS.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,7 +3,10 @@ enable_testing()
 
 find_package(Boost COMPONENTS system filesystem unit_test_framework serialization REQUIRED)
 include_directories(${TEST_SOURCE_DIR}/src ${Boost_INCLUDE_DIRS})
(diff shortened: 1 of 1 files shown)


neubig:
Thank you very much for the contribution! But actually, with the pending version 2.0 of DyNet, Boost will basically not be necessary unless you want to build unit tests: https://github.com/clab/dynet/pull/637

I believe this will solve the problem, but if you think the pull request would still be useful I can take a look at integrating it into v2.

neubig:
Closing, as this is probably not necessary in v2.0. Thank you for the contribution though!

PR #631, opened 2017-06-26
update swig bindings for v2

joelgrus:
I had to remove some class methods that no longer worked, but I don't think anyone was using them, so it shouldn't be a big deal.

the bigger deal is that the Scala bindings had pretty sophisticated serialization / deserialization wrappers around boost. with boost gone, those features are gone, and what's left is pretty basic.
there was no easy way to keep the functionality without writing a lot of custom code.

this likely affects @jayantk, who should stick with the old version and/or talk to me about how to get back the parts he really needs


neubig:
Thanks! Yes, the serialization interface is now more minimalistic, basically handling only Parameters, LookupParameters, and ParameterCollections. The motivation behind this is that we want to clearly delineate the line between what DyNet should be doing and what is the user's responsibility. Now, the user is responsible for defining, saving, and loading the model configuration, while DyNet will only save and populate the parameters of the model. This might cause a little bit of inconvenience for those who need to port to the new version, but hopefully it won't be too bad.

(On an unrelated note, It looks like the build failed due to an unrelated error, but I've fixed this so I'm sure it should be OK.)

yoavg:
@joelgrus @jayantk you may want to look at how the saving/loading is handled in the python wrapper---I support both the cpp-like version as well as an API that allows saving / loading lists of builders / parameters as before, and the wrapping code was actually quite minimal.

https://github.com/clab/dynet/blob/v2rc/python/_dynet.pyx#L250
https://github.com/clab/dynet/blob/v2rc/python/_dynet.pyx#L304

might be a bit more involved with the less-dynamic typing of scala compared to python, but I think shouldn't be too terrible to add.

PR #638, opened 2017-06-28
MKL_ROOT and EIGEN3_INCLUDE_DIR as env variables

FilippoC:
This change allows the user to simply use "cmake .. -DMKL=true" and provide MKL_ROOT as an environment variable, as already allowed for BOOST_ROOT + same thing for EIGEN3_INCLUDE_DIR.


neubig:
Thanks!

PR #646, opened 2017-06-30
Protected Dict attributes instead of private

FilippoC:
Hello,
I just made the switch to V2. I have a huge codebase that makes an extensive use of dynet Dicts. I moved to the new save/load method for my NN paramaters, however, for convenience reasons, I want to use boost::serialization for Dicts.
This commit only change the scope of Dict attributes from private to protected.

This allows me do to the following non-intrusive hack in my own code :
`
struct Dict : public dynet::Dict                
{  
     template<class Archive> 
     void serialize(Archive& ar, const unsigned int)                                                  
     {       
             ar & frozen;
             ar & map_unk;                       
             ar & unk_id;                        
             ar & words_;                        
             ar & d_;                            
     }                                           
};
`

Note : both documentation and example seems to be leaking of explanation about saving/loading dict.

### DIFF
--- a/dynet/dict.h
+++ b/dynet/dict.h
@@ -68,7 +68,7 @@ public:
   
   void clear() { words_.clear(); d_.clear(); }
 
-private:
+protected:
   bool frozen;
   bool map_unk; // if true, map unknown word to unk_id
   int unk_id; 


neubig:
Thanks! As of v2, DyNet (particularly C++) only supports saving of model parameters, and everything else is saved by the user. I agree that we should modify the examples to show the alternative behavior (where the user saves the dictionaries on their own).

PR #650, opened 2017-07-02
Update version in setup.py

danielhers:
Note: this is to the v1.1 branch since I noticed there is a release from it. I uploaded the package to PyPI, so now it can simply be installed using `pip install dynet`. @neubig please let me know what's your PyPI username so I can add you as an owner.


neubig:
Thank you! My PyPi username is neubig.

PR #651, opened 2017-07-02
Remove Boost from Python setup.py

danielhers:
Boost is only needed for unit tests. Since those are not run when installing using pip, the Boost option is not needed.


neubig:
Thanks!

PR #652, opened 2017-07-02
Some fixes and functionalities

shuheik:
These commits
1) fix bug with CyclicalSGDTrainer.update(). Before this fix Trainer.update() was being called.
2) fix bug around set/disableDropout in LSTMBuilders.
3) add accessor to eta in Trainer.
4) add Expression.constParameter(lp: LookupParameter)

2 is a bit complicated. Before this fix, a) setDropout with one argument changed only dropout_rate, not dropout_rate_h or dropout_rate_c, b) if setDropout with multiple arguments is called and disableDropout and addInput are called after ComputationGraph.renew(), it results in an error inside native methods.
However, if I add one line in lstm.cc like this,
```
void VanillaLSTMBuilder::start_new_sequence_impl(const vector<Expression>& hinit) {
  cerr << "dropout_rate: " << dropout_rate << " dropout_rate_h: " << dropout_rate_h << "\n";
  h.clear();
  c.clear();
```
b) doesn't result in crash. Hopefully, with the fix things work as expected.


neubig:
Thanks!

Issue #648, opened 2017-06-30
Model saving much slower with 2.0?


pmichel31415:
Saving a medium sized seq2seq model takes between 2x and 3x longer as before, is anyone else experiencing something similar?

neubig:
Hmm, maybe @xunzhang can take a look at this?

xunzhang:
Hi @pmichel31415 , I did a performance regression test just now. But in my Linux environment, the save function in v2.0(12s, 266MB model file after saving) is a little faster than in dynet1.1(14s, 525MB model file after saving). For load, since the model file in v2.0 is much smaller, the loading time much shorter(8s vs 14.5s). I tested both cpp and python interface, there is no obvious difference. I also tested saving with nested subcollection, the result is similar.
A possible possibility is that I generated parameters and lookup parameters randomly while you run a real model data. So could you give the dataset and how to reproduce it so that I can tune it, thanks.

pmichel31415:
Hi @xunzhang , thanks for looking into this.

I've noticed that for some other runs the save time was much faster (for the same model/data) sot it might be a problem if read/write speed on my machine?

In any case, do you think it would be hard/a lot of work to implement binary saving/loading with the current interface? I would be willing to code that.

neubig:
I don't think implementing a binary serializer would be too hard, and it would certainly be useful.

xunzhang:
@pmichel31415, I think it shouldn't be too difficult. It is useful if the performance is much faster using binary serializer.

pmichel31415:
Hmm so I inspected the model files manually and I saw a lot of zeros. 

It appears that the `TextFileSaver` object is saving the gradients as well as the values.

I can see the advantage of saving gradients on some cases but I think most of the times people save the parameters after updating them (so the gradients are 0).

Maybe there should be a a flag to enable gradient saving, or some way to indicate that the gradients are 0 and that they don't need to be saved.

That would save a lot of space.

yoavg:
@pmichel31415 I think asking the user to save gradients or not is kinda inelegant. I think an easier solution would be to always save the gradients, but if they are all zero just indicate it with a predefined symbol (say "ZERO_GRAD") instead of a long list of zeros. Then loading needs to be aware of that, and that's it. We'll get both the smaller model and speedier load/save times for this case, without any effort from the user.

Re binary format: should be trivial to do, but need to make sure it is compatible across machines.
[...]

PR #663, opened 2017-07-05
Skip saving zero gradients

pmichel31415:
This addresses #648 (in parts).

- Adds a `ZERO_GRAD`/`FULL_GRAD` flag in the parameter headers of serialized parameters. If the flag is set to  `ZERO_GRAD`, the gradient is not saved (and is set to zero upon loading)
- Adds a `nonzero_grad` field to ParameterStorage. This field is updated when calling `accumulate_grad` or `clear`

Wrt the second point, I initially checked the norm of the gradient before serializing but I think it might cause performance issues for big parameters, so I introduced the `nonzero_grad` field to keep track of the norm of the gradient.

One inconvenient of this approach is that if the value of the gradient is modified in any way other than with `accumulate_grad`, (like if the gradient is modified manually), this variable won't be updated.

### DIFF
--- a/dynet/io.cc
+++ b/dynet/io.cc
@@ -26,9 +26,7 @@ bool valid_pc_key(const std::string & s) {
 }
 
 bool grad_is_zero(const ParameterStorageBase & p){
-  float grad_norm;
-  p.g_squared_l2norm(&grad_norm);
-  return (grad_norm == 0);
+  return !p.has_grad();
 }
 
 void read_param_header(string line, string &type, string &name, Dim& dim,size_t& byte_count, bool& zero_grad){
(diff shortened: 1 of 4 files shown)


neubig:
Thanks!

xunzhang:
Looks good! @pmichel31415 Do you get some the performance gain using this optimization running your seq2seq model?

pmichel31415:
It reduced the size of the file by 12%, which is not huge but still significant especially since I save a lot of models.

As for the speed, I can't really comment. It's much faster now but I suspect there were external factors influencing the disk io speed so the comparison doesn't hold.

PR #664, opened 2017-07-05
Updated Python hinge to work with minibatching

neubig:
In Python, unlike `pickneglogsoftmax`, `hinge` could not process multiple examples at the same time. This fix corrects this problem.


yoavg:
I am perfectly happy with it, but just to be clear that this removes a previous (probably never used...) functionality of static graph construction while updating the hinge coordinates per example.

PR #666, opened 2017-07-06
Clean up includes for faster compile time

neubig:
Compile time for DyNet is somewhat slow. This PR speeds up compilation a bit by optimizing includes with include-what-you-use (https://include-what-you-use.org). There are still a number of places where we could move things from header files to .cc files though, which would further speed things up.


neubig:
This didn't give large gains, likely because some of the implementation is still in the header files. We should try to fix this problem first, then re-optimize the includes.

PR #667, opened 2017-07-06
temporally fix a bug about negative padding in eigen conv2d

zhisbug:
The bug only occurs in some very extreme cases.
More details see here: https://github.com/tensorflow/tensorflow/issues/11092

Before TF community gives a fix I think this patch can be a temporal workaround.


neubig:
Thanks!

PR #673, opened 2017-07-07
new_graph checking and cluster initialization in hsm-builder.cc

jiyfeng:
In hsm-builder.cc: 
(1) fix the bug on checking whether new_graph is called; 
(2) pass the rep_dim explicitly to the cluster children nodes.


neubig:
Thanks!

Issue #522, opened 2017-05-05
File structure refactoring


neubig:
Currently compiling DyNet takes a long time for a couple of reasons:
* Inclusion of header files is not clean, and header files are included when not necessary in a number of places.
* `nodes.cc` is a super-huge file that implements virtually all operations.

I think we should refactor the file structure so that

- [x] `nodes.cc` is split into a finer-grained divisions, perhaps similar to the doc: http://dynet.readthedocs.io/en/latest/operations.html
- [x] `nodes-common.cc` is removed, with all the functionality being moved to the appropriate files from the previous bullet point. This will make sure that the code implementing a particular node will all be in the same file.
- [ ] All header file includes that are not necessary are removed.
- [ ] Other files that have lots of different classes implemented (e.g. `dynet.h` have the functionality split into different files named after the classes so that they're easier to find (e.g. `computation-graph.h` and `node.h`)

pmichel31415:
Would it also be a good opportunity to separate the files in different subfolders? Like all the builders in one folder, all the nodes in another, etc...

Would make it easier to navigate the source code.

Also (but maybe this is a separate issue) we should probably split `_dynet.pyx` in several files as well.

neubig:
This is basically done.

Issue #661, opened 2017-07-05
When compiling on CUDA, unnecessary compute architectures are generated


neubig:
On this line, we are specifying that CUDA should compile on a bunch of different computation architectures from `compute_30` co `compute_52`:
https://github.com/clab/dynet/blob/master/dynet/CMakeLists.txt#L125

In most cases, we probably only need to compile on a single computing architecture, the most recent one supported by the `nvcc` compiler and GPU. Because compiling for different architectures takes a long time and is a significant time bottleneck in getting DyNet installed on GPU, it would be nice to fix this. I'm not sure how to do so though, so if someone knows how help would be appreciated.

mattr1:
I found compiling just for architecture 5.2 makes the compilation roughly twice as fast. The only way I can think of is a parameter passed to cmake telling it what architecture the user wants, with a suitable default (maybe default to all the architectures, as it is currently done, so that people who are just compiling once will have a library that runs everywhere). We could have a page in the documentation on speeding up compilation, or a page for developers, describing options like this.

PR #680, opened 2017-07-11
reduce .cu include files

xunzhang:
Part of https://github.com/clab/dynet/issues/522

### DIFF
--- a/dynet/CMakeLists.txt
+++ b/dynet/CMakeLists.txt
@@ -126,41 +126,41 @@ endif()
 set(dynet_gpu_SRCS
     cuda.cc
     cudnn-ops.cu
-    gpu-ops.cu 
-    gpu-nodes-activations.cu
-    gpu-nodes-affinetransform.cu
-    gpu-nodes-arith-const.cu
-    gpu-nodes-arith-cwise.cu
-    gpu-nodes-arith-scalar.cu
-    gpu-nodes-arith-sum.cu
-    gpu-nodes-arith-unary.cu
-    gpu-nodes-concat.cu
-    gpu-nodes-const.cu
-    gpu-nodes-contract.cu
-    gpu-nodes-conv2d.cu
-    gpu-nodes-conv.cu
-    gpu-nodes-dropout.cu
-    gpu-nodes-flow.cu
-    gpu-nodes-hinge.cu
-    gpu-nodes-linalg.cu
-    gpu-nodes-logsumexp.cu
-    gpu-nodes-losses.cu
-    gpu-nodes-matrixmultiply.cu
-    gpu-nodes-maxpooling2d.cu
-    gpu-nodes-minmax.cu
-    gpu-nodes-moments.cu
-    gpu-nodes-normalization.cu
-    gpu-nodes-norms.cu
-    gpu-nodes-pickneglogsoftmax.cu
-    gpu-nodes-random.cu
-    gpu-nodes-select.cu
-    gpu-nodes-similarities.cu
-    gpu-nodes-softmaxes.cu
-    gpu-nodes-trig.cu
-    gpu-param-nodes.cu
-    gpu-tensor.cu
-    gpu-training.cu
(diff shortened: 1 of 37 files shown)


neubig:
I think this will significantly increase compile time on GPU because it will cause `gpu-nodes-operators.cu` to take a long time, and not allow for multithreading, right? Also, this file will likely be re-compiled any time we make a change to one of the included `.cc` files.

I agree that we'd like to remove all the extra `gpu-...` files though. Perhaps we can find some way to generate them automatically using `cmake`, or some way to specify that `nvcc` should compile `.cc` files.

xunzhang:
@neubig good advice. Done using auto generated file solution.

xunzhang:
This pull request keeps the same compile time as https://github.com/clab/dynet/pull/679. One question here, I noticed that after https://github.com/clab/dynet/pull/679, on my local dev machine using `make -j8`, there is no obvious performance gain under GPU mode. Is that expected? What's your result? Thanks.

mattr1:
@xunzhang: If it's not too much trouble, I think the CMakefile would be much improved if the list of gpu files is not duplicated (in dynet_gpu_SRCS and the foreach loop). Would it be possible to have the foreach append the newly created file to dynet_gpu_SRCS instead? Something like "list(APPEND dynet_gpu_SRCS ${PROJECT_BINARY_DIR}/dynet/gpu-${FILENAME}.cu". Or, I can do this after the PR is merged if you prefer.

mattr1:
Having the many small .cu files compiled independently has significantly increased the compile time of gdynet on MSVC (from 4.5 minutes for a single monolithic .cu file to 18 minutes for the current split of 30 files), because MSVC (unfortunately) doesn't do parallel building of that part of the compilation (they are custom actions calling nvcc, rather than using the built-in compiler). 

I consider that MSVC's fault. However, note there is a lot of overhead in compiling each .cu file because the majority of the code is Eigen that is getting parsed and compiled over and over. That 4x increase in compile time is comparing two sequential compiles, so it shows how much overhead there is. I would imagine this overhead is also hurting the parallel Linux build, and that it would probably be faster to build in Linux if there were just, say, 4-8 files built in parallel instead of the current 30. This would also have the side-benefit of speeding up the MSVC compile.

neubig:
P.S., and I agree that it'd be better to not duplicate the list of files twice in the CMakeLists file.

mattr1:
Agreed on the difficulty of balancing the goals. Though, my experience so far is that changes often affect something like dynet.h or model.h and so the entire gdynet has to get recompiled anyway.

I have found that if I compile only for the latest CUDA architecture, I get something like a 2x speedup in compilation time. I see you opened an issue to track that here: https://github.com/clab/dynet/issues/661. I'll comment there.

At the risk of adding complexity, I could imagine adding a CMake parameter setting that switches between compiling each .cu individually, vs. compiling 4-8 of them, vs. compiling one monolithic one. The default should probably be whatever is fastest for users to just do a single compile, but then developers could modify it to suit their needs. Something like NUM_PARALLEL_CUDA_COMPILES or something like that. I can take a look at that if you think this is a good direction -- it'll probably involve a few lines of CMake doing math and modular arithmetic and appending to files.

xunzhang:
@mattr1 thanks for the comment, very helpful! Modified with removing the duplicate declaration of filenames.

neubig:
Thanks, I'll merge this for now. @mattr1, I think that for now having an option about whether to perform parallel compilation of the `gpu-nodes-*.cu` classes, or more broadly, whether to perform `nodes-*.cc` in parallel would be reasonable, at least for the time being. Also on Travis CI or when installing through, having `nodes-*.cc` compile as one big file might be useful as well, as I don't believe these are making heavy use of multi-threading.

mattr1:
Okay, I'll take a look.

mattr1:
Created PR https://github.com/clab/dynet/pull/689 to allow merging GPU compilation into k files.

PR #686, opened 2017-07-12
Better softmaxbuilders

pmichel31415:
This:

- Adds documentation for all softmax builders
- Adds additional methods to softmax builders
- Adds minimal tests for the standard/classfactored softmax
- Adds python bindings & doc & minimal tests for the standard/classfactored softmax


neubig:
Looks good, thanks!

PR #687, opened 2017-07-12
Fix lstm mask creation condition

FilippoC:
The condition only checks the input dropout. This patch fix this condition for both Coupled and Vanilla LSTM.


neubig:
Nice catch, thanks!

PR #688, opened 2017-07-12
added a scratch memroy allocator and updated conv2d to use it

Abasyoni:
Please review the addition of a scratch memory allocator to the system and how it's used in Conv2D.


neubig:
Thanks! This will be helpful to help conserve memory for operations where extra memory is only needed temporarily.

neubig:
Actually, I realized that I merged this too soon... I didn't realize it hadn't been implemented or tested on GPU. The commit here at least fixes the errors, but the behavior on GPU is rolled back to the original behavior, so please take a look and finish the GPU implementation as well: https://github.com/clab/dynet/commit/4157ba0181e2e0bf72b5c4c38cb4978ecfb581b0

zhisbug:
Could we rollback this merge? I haven't reviewed this commit yet, I think @Abasyoni forgot to mark this one as WIP

neubig:
@zhisbug sorry, I realize now that I should have waited for your review. I did check to make sure that the pull is ok conceptually, and the remaining work is pretty minimal so I don't think we need to roll back. We can just finish the remaining stuff today I think.

neubig:
I have a few minutes to look at this now, so I'll try to come up with a draft.

PR #689, opened 2017-07-13
merge GPU files together for faster compilation. Merging is controlled by GPU_NUMFILES parameter to Cmake

mattr1:
Default values for GPU_NUMFILES are 1 for MSVC (since it compiles CUDA serially) and 4 for Linux (picked arbitrarily -- should probably be number of cores). A few questions:

1) Are you happy with the name "GPU_NUMFILES"
2) Where should we document this? The average user probably doesn't need to know about it. Only developers, who may prefer to set GPU_NUMFILES to 0 in order to get one file per .cc file

### DIFF
--- a/dynet/CMakeLists.txt
+++ b/dynet/CMakeLists.txt
@@ -164,10 +164,6 @@ set(dynet_gpu_SRCS
     cudnn-ops.cu
     gpu-ops.cu)
 
-foreach(FILENAME ${dynet_gen_file_prefix_SRCS})
-    list(APPEND dynet_gpu_SRCS ${PROJECT_BINARY_DIR}/dynet/gpu-${FILENAME}.cu)
-endforeach()
-
 file(GLOB TEST_SRCS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} tests/*.cc)
 if (NOT MSVC)
   set(BUILD_SHARED_LIBS ON)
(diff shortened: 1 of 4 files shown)


neubig:
Thanks a lot, this looks helpful! With regards to documentation, maybe here would be good? http://dynet.readthedocs.io/en/latest/code_style.html#coding-tips

dynet/CMakeLists.txt
```diff
@@ -196,11 +192,28 @@ endif()
 # Build cpu library
 add_library(dynet ${dynet_library_SRCS} ${dynet_library_HDRS})
 target_link_libraries(dynet ${LIBS})
+
 if(WITH_CUDA_BACKEND)
-  foreach(FILENAME ${dynet_gen_file_prefix_SRCS})
-    set(FILEPREFIX ${FILENAME})
-    configure_file(gpu-file-template.cu.in gpu-${FILENAME}.cu)
-  endforeach()
+  if(${GPU_NUMFILES} EQUAL 0)
```
mattr1: I guess I should have left some comments in this part of the code. When GPU_NUMFILES is 0, it generates the ~30 files as before, with specific file naming. If we didn't care about the naming, then this if statement could be simplified to just set GPU_NUMFILES to the length of the list of merge files, but I think it's nice when not doing any merging to keep the files named like "gpu-nodes-flow.cu" rather than "gpu-merge0.cu".

mattr1:
xunzhang: To answer your question, I do see a big improvement on Linux for time to compile gdynet with the merged files. Here is timing for compiling with either 4 or 8 parallel jobs, and varying the GPU_NUMFILES parameter (these are on a 24 core (12 real cores * hyperthreading) system:

make gdynet -j 4
- 3:29 GPU_NUMFILES=4
- 3:32 GPU_NUMFILES=8
- 5:38 GPU_NUMFILES=0 (unmerged)

make gdynet -j 8
- 2:42 GPU_NUMFILES=4
- 2:34 GPU_NUMFILES=8
- 3:44 GPU_NUMFILES=0 (unmerged)

So when GPU_NUMFILES is set to the number of jobs, I get about a 30% speedup. Also, on Windows it's about a 4-5x speedup setting GPU_NUMFILES=1.

neubig:
Apparently this was causing compile errors on some varieties of Linux. The problem was basically that the `log1p` function in `functors.h` was being called after `using namespace std` in some of the `.cc` files, which was making `nvcc` get confused between the `__device__` function `log1p` and the `__host__` function `std::log1p`. I'm not sure how to fix this error easily, so for now I've reverted to compiling each node file separately by default:
https://github.com/clab/dynet/commit/dc2d02e6774651886091cb90ead47b1e50d0507c

xunzhang:
@neubig This is the simple fix: https://github.com/clab/dynet/pull/701

PR #692, opened 2017-07-13
Set conv2d to use scratch memory on gpu

neubig:
@zhisbug @Abasyoni This is a first draft of GPU support. Does it look reasonable? If so, let's do max pooling as well.


neubig:
OK, this should be ready now?

neubig:
I'm going to merge this because I'm pretty sure it's right, and would like to use this functionality.

zhisbug:
@neubig  Fine! Thanks.

Issue #684, opened 2017-07-12
update_epoch changes eta regardless of current eta


shuheik:
Trainer.eta can be modified directly as in
https://github.com/clab/dynet/blob/master/examples/cpp/rnnlm/train_rnnlm.cc#L271-L272
```
if (eta_decay_onset_epoch && completed_epoch >= (int)eta_decay_onset_epoch)
  sgd->eta *= eta_decay_rate;
```

Trainer.update_epoch() increments epoch and changes eta according to eta0, epoch and eta_decay, regardless of current eta.
https://github.com/clab/dynet/blob/master/dynet/training.h#L72
```
void update_epoch(real r = 1) {
  epoch += r;
  eta = eta0 / (1 + epoch * eta_decay);
  }
```

So it looks that if eta is changed directly and update_epoch is called after that, the first change to eta is cancelled.
Maybe it's intended but it looks confusing to me.

neubig:
I agree this is confusing. I'll think about a good way to simplify the interface.

neubig:
OK, we simplified the interface and hopefully made it more intuitive, see here: https://github.com/clab/dynet/pull/695

Issue #641, opened 2017-06-29
Passing scale to Optimizer changes gradient clipping threshold


alexandres:
Trainers accept an initial learning and a decay parameter. To control the learning rate dynamically without using the built-in decay, one can pass a ```scale``` parameter to the ```trainer.update(scale)``` function. Unfortunately, this directly affects gradient clipping:

From https://github.com/clab/dynet/blob/master/dynet/training.cc#L68:

```
if (scale * gg > clip_threshold) {                              
   ++clips;                                                      
   ++clips_since_status;                                         
   gscale = clip_threshold / (scale * gg);                       
}  
```                                                             

Using ```trainer.update(scale)``` to manage the learning rate, ```clip_threshold``` is effectively divided by ```scale```. Though there might be a reason for this, I was surprised by it. To get the same results as I would with TensorFlow, I have to compensante for this by calling ```trainer.set_clip_threshold(clip_threshold * scale)``` before calling ```update(scale)```.

neubig:
Thanks for pointing this out, we'll take a look.

neubig:
This was indeed unintuitive, so we revised the training interface, simplifying it and removing the "scaling" parameter in update, which was not intended to be used to scale the learning rate. Now you need to manage the learning rate in the way that it was intended, by setting the learning_rate member of the trainer class: https://github.com/clab/dynet/pull/695

PR #695, opened 2017-07-14
Simplify training interface by removing weight decay and scaling

neubig:
There is some confusion regarding the interface of the "Trainer" class:
https://github.com/clab/dynet/issues/641
https://github.com/clab/dynet/issues/684

I agree that it's difficult to understand. This commit removes the rate decay and gradient scaling functionality that implicitly changes the learning rate in non-transparent ways. Here is an example of the before/after behavior:

*Rate Decay Before*

    // At beginning of training
    Trainer trainer(initial_learning_rate, rate_decay)
    // After every epoch
    trainer.update_epoch()

*Rate Decay After*

    // At beginning of training
    Trainer trainer(initial_learning_rate)
    // After every epoch
    trainer.learning_rate /= (1 - rate_decay)

*Gradient Scaling Before*:

    cg.backward(loss)
    trainer.update(scaling_factor)

*Gradient Scaling After*:

    cg.backward(loss * scaling_factor)
    trainer.update()

### DIFF
--- a/dynet/mp.h
+++ b/dynet/mp.h
@@ -259,7 +259,9 @@ namespace dynet {
           }
           if (do_update && trainer != nullptr) {
             shared_object->update_mutex.wait();
-            trainer->update(1.0 / counter); 
+            // TODO: The scaling was originally this
+            // trainer->update(1.0 / counter); 
+            trainer->update(); 
             shared_object->update_mutex.post();
           }
           if (batch_counter == header.report_frequency) {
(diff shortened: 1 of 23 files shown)


liesun1994:
I am wondering  if lamtram need to be modified with the issues pulled.

liesun1994:
When I am running lamtram， it may occurs Trainer::update_epoch has been deprecated and doesn't do anything. Please remove it from your code, and control the learning rate of the trainer directly, for example by: 'trainer.learning_rate /= (1 - rate_decay)', see https://github.com/clab/dynet/pull/695 for details .

yuvalpinter:
I can't seem to be able to run the `trainer.learning_rate /= (1 - rate_decay)` code. It gives me the following error: `TypeError: 'property' object is not callable`.
This happens for both `MomentumSGDTrainer` and `AdamTrainer`.

shuheik:
Shouldn't it be 
`trainer.learning_rate *= (1 - rate_decay)` or 
`trainer.learning_rate /= (1 + rate_decay)`
instead of 
`trainer.learning_rate /= (1 - rate_decay)`?

I assume rate_decay to be some small positive value and dividing by (1- rate_decay) would make learning_rate grow larger.

PR #696, opened 2017-07-14
Disable performance tests by default

neubig:
Performance tests are important, but take much longer than other tests to run and will never explicitly fail. This PR disables them by default, which will speed completion of the `make test` command, but allows them to be run on request.


neubig:
Yes, you can still run individual tests from the command line, which will work even for disabled tests.

    test-program --log_level=test_suite --run_test=testA

PR #701, opened 2017-07-15
Fix multithread compile error

xunzhang:


mattr1:
Actually, I think a slightly better solution is to call ::log1p so that if dynet::real was ever changed to double, it would still be valid. At least, that fixes the compile error for me -- I didn't run the tests to verify.

PR #702, opened 2017-07-16
Update Scala bindings

### LINKED ISSUE #683: SerializationSpec fails during build
With current master.

> [info] SerializationSpec:
[info] dynet
[info] - should create models with the right number of parameters
[info] model saver and model loader
/var/folders/41/1rxgyck52f5c72_h11lscr_m0000gn/T/dynet_test7800056718491685052serialization_spec
[info] - should handle simplernnbuilder correctly *** FAILED ***
[info]   0.0 was not -0.5364716 plus or minus 1.0E-5 (SerializationSpec.scala:13)
[info] model saver and model loader
[info] - should handle vanillalstmbuilder correctly *** FAILED ***
[info]   0.0 was not 0.27130622 plus or minus 1.0E-5 (SerializationSpec.scala:13)

Maybe related to #663?

shuheik:
1. adapt to changes in training interface
2. fix Expression.constLookup, let Expression.pickrange use pick_range instead of pickrange
3. fix #683.

It turned out that saver.done() was missing from SerializationSpec.scala and it caused the error. 
The last line of the model is not written to the file until ModelSaver.done() is called, and if a ModelLoader reads the file before the saver is done and tries to populate a new model, it fills missing values with the ones from the last parameter in the unfinished file. 
I guess it was passing the test before #663 because what's not written to the model file until ModelSaver.done() was the gradient.

I think Expression.pickrange should be renamed to Expression.pickRange, but kept it as is for compatibility.
If nobody mind I'll change them and send another PR.

### DIFF
--- a/contrib/swig/dynet_swig.i
+++ b/contrib/swig/dynet_swig.i
@@ -559,7 +559,6 @@ Expression pick(const Expression& x, unsigned v, unsigned d = 0);
 Expression pick(const Expression& x, const std::vector<unsigned>& v, unsigned d = 0);
 Expression pick(const Expression& x, const unsigned* v, unsigned d = 0);
 Expression pick_range(const Expression& x, unsigned s, unsigned e, unsigned d = 0);
-Expression pickrange(const Expression& x, unsigned v, unsigned u);
 
 // Concatenate and ConcatenateCols got changed around, need to implement
 // explicitly now.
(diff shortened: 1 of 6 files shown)


contrib/swig/dynet_swig.i
```diff
@@ -558,6 +558,7 @@ Expression sum_batches(const Expression& x);
 Expression pick(const Expression& x, unsigned v, unsigned d = 0);
 Expression pick(const Expression& x, const std::vector<unsigned>& v, unsigned d = 0);
 Expression pick(const Expression& x, const unsigned* v, unsigned d = 0);
+Expression pick_range(const Expression& x, unsigned s, unsigned e, unsigned d = 0);
 Expression pickrange(const Expression& x, unsigned v, unsigned u);
```
neubig: Should this pickrange function be deleted?

shuheik:
Yes, I've deleted the old pickrange.

PR #703, opened 2017-07-17
make both 1.0*x and x*1.0 work

jcyk:
Note \_cmul and \_cadd are not commutative operations in our
implementation.


pmichel31415:
Thanks a lot!

PR #704, opened 2017-07-17
[Dynet-92]. Multi-device support

xunzhang:
- [x] Refactor dynet to support multi-device more friendly.
- [x] Honor `--dynet-devices` argument.
- [x] Implement interfaces(partly) specifying device type as the argument when defining an expression or using `cg.change_expr_device` before defining an expression.
- [x] Implement memcpy between devices in forward process.
- [x] Test forward in hybird CPU/GPU mode with basic expression: `V * tanh(affine_transform(b, W, x)) + a`.
- [x] Implement memcpy between devices in backward process.
- [x] Test backward in hybird CPU/GPU mode with basic expression: `V * tanh(affine_transform(b, W, x)) + a`.
- [x] Add support for rollback to CPU mechanism when there is no GPU implementation yet.
- [ ] ~~Debug hang issue using multiple GPUs~~
- [ ] ~~Add more feature tests.~~

Original Usage:
```
./a.out --dynet-devices CPU,GPU:0,GPU:1

int main(int argc, char *argv[])
{
  dynet::initialize(argc, argv);

  for (iter) {
    ComputationGraph cg(dynet::devices_map["GPU:0"]); // default device if not specify
    Expression W = parameter(cg, p_W, dynet::devices_map["CPU"]);
    Expression b = parameter(cg, p_b); // default: GPU:0
    Expression x = input(cg, {2}, x_values); // default: GPU:0
    cg.change_expr_device(dynet::devices_map["GPU:1"]); // change default device for future expressions
    Expression h = tanh(affine_transform({b, W, x})); // reside GPU:1
    
    Expression last = ...;
    cg.forward(last);
    cg.backward(last);
    // update
  }
  return 0;
}
```

Modified Usage:
```
./a.out --dynet-devices CPU,GPU:0,GPU:1

int main(int argc, char *argv[])
{
  dynet::initialize(argc, argv);

  for (iter) {
    ComputationGraph cg;
    Expression W = parameter(cg, p_W, dynet::devices_map["GPU:0"]);
    Expression b = parameter(cg, p_b); // default to p_b's device(GPU:0)
    Expression x = input(cg, {2}, x_values, dynet::devices_map["CPU"]);
    Expression x_2 = to_device(x, dynet::devices_map["GPU:0"]);
    Expression h = affine_transform({b, W, x}); // default to b's device(GPU:0)
    Expression h_2 = to_device(h, dynet::devices_map["CPU"]);
    Expression v = tanh(h_2); // default to h2's device(CPU), suppose tanh has no cuda impl in this case 
    
    Expression last = ...;
    cg.forward(last);
    cg.backward(last);
    // update
  }
  return 0;
}
```

To reviewer @neubig , you can have a quick test using code below:
```
// usage: ./a.out --dynet-devices CPU,GPU:0

#include <iostream>
#include "dynet/dynet.h"
#include "dynet/training.h"
#include "dynet/expr.h"
#include "dynet/io.h"
#include "dynet/model.h"
#include "dynet/devices.h"

using namespace std;
using namespace dynet;

int main(int argc, char** argv) {
  dynet::initialize(argc, argv);

  const unsigned ITERATIONS = 30; 

  // ParameterCollection (all the model parameters).
  ParameterCollection m;
  SimpleSGDTrainer sgd(m);

  const unsigned HIDDEN_SIZE = 8;
  Parameter p_W = m.add_parameters({HIDDEN_SIZE, 2});
  Parameter p_b = m.add_parameters({HIDDEN_SIZE});
  Parameter p_V = m.add_parameters({1, HIDDEN_SIZE});
  Parameter p_a = m.add_parameters({1});
  if (argc == 2) {
    // Load the model and parameters from file if given.
    TextFileLoader loader(argv[1]);
    loader.populate(m);
  }

  // Static declaration of the computation graph.
  ComputationGraph cg; 
  Expression W = parameter(cg, p_W);
  Expression b = parameter(cg, p_b);
  Expression V = parameter(cg, p_V);
  Expression a = parameter(cg, p_a);

  // Set x_values to change the inputs to the network.
  vector<dynet::real> x_values(2);
  Expression x = input(cg, {2}, &x_values);
  dynet::real y_value;  // Set y_value to change the target output.
  Expression y = input(cg, &y_value);

  Expression aa = W * x + b;
  Expression hhh = to_device(aa, dynet::get_global_device("CPU"));
  Expression h = tanh(hhh);
  Expression hh = to_device(h, dynet::get_global_device("GPU:0"));
  Expression y_pred = V*hh + a;
  Expression loss_expr = squared_distance(y_pred, y); 

  // Show the computation graph, just for fun.
  cg.print_graphviz();

  // Train the parameters.
  for (unsigned iter = 0; iter < ITERATIONS; ++iter) {
    double loss = 0;
    for (unsigned mi = 0; mi < 4; ++mi) {
      bool x1 = mi % 2;
      bool x2 = (mi / 2) % 2;
      x_values[0] = x1 ? 1 : -1; 
      x_values[1] = x2 ? 1 : -1; 
      y_value = (x1 != x2) ? 1 : -1; 

      loss += as_scalar(cg.forward(loss_expr));
      cg.backward(loss_expr);
      sgd.update();

    }
    loss /= 4;
    cerr << "E = " << loss << endl;
  }

  // Output the model and parameter objects to a file.
  TextFileSaver saver("/tmp/xor.model");
  saver.save(m);
}

```


neubig:
In general, this is great: I think multi-device support will be a great feature for DyNet to have. First, I have a high level comment. In my mind, there are two design decisions here:

### How do we specify the "default" device of a graph node when it is not specified explicitly
1. *Current Implementation:* A default is passed to ComputationGraph, and the default is used.
2. Alternatively, we could have the node default to the device of its first argument.

The first has the advantage of perhaps being easier to understand, but may result in hidden memory moves where people aren't expecting them. It also adds some code complexity.

### When some of the inputs are not on the same device what do you do?
1. *Current Implementation:* The ExecutionEngine is responsible for moving memory.
2. The ExecutionEngine throws an error, telling the user to move the memory themselves (using something like `dy.change_device(x, device)`).
3. A combination of 1. and 2., where 2. is on by default, but 1. can be chosen.

1. and 3. have the advantage of not crashing, but also have the potential to hide memory moves that the user really wouldn't want to be doing. (For example, in the example code, the weight matrix would be passed from CPU to GPU every time it was used, which would be really really bad.) 2. has the advantage of preventing this, but may result in a sightly increased coding burden.

*My opinion:* I tend to prefer 2./2. respectively, but could be convinced otherwise.

yoavg:
This is really great!

I like options 2 and 2 also, but would like to propose a variation of the second (2):

the name `dy.change_device(x, device)` is a bit confusing imo, as we are not changing the device of x as much as copying x to another device (it can also be used on the same device after). So I propose to change @neubig 's proposed interface slightly to:
`Expression y = x.to_device(device)`

letting both `x` and `y` be used.

Another proposal (maybe its already there, I didn't look at the code) is to also allow multiple CPU devices. There, the copying will be NOPs, but we can still run different cpu devices as different threads.

How do things look in terms of synchronization in the current implementation?

xunzhang:
Very helpful comments!! I also prefer 2. / 2. then. And I think to support both copy like `x.to_device(device)` and move like `x.change_device(device)` are necessary. But `to_device` interface is sort of hard to implement since there will be some discrete `VariableIndex` indexes and we need to refactor the executor code to support it. I think I will first finish `change_device` and then implement `to_device`.

Currently, we don't support specifying CPU id. I will think about that in the near future.

xunzhang:
The remark of `to_device` and `change_device` in my last comment might be incorrect. Basically, I think `to_device` is a copy like operation which will create an additional node while `change_device` is sort of changing the device assignment of an expression(not sure if this semantic is useful or not).

neubig:
@xunzhang Yes, `to_device` is an operation that will create a new node (where the memory is stored on a different device than a single node). I don't think that we should have a function to change the device of a particular node for the reasons you mentioned: it would complicate things and require special handling in the executor. Regarding Yoav's comment about having multiple CPU devices, I'm not sure that this is necessary. In order to do things on multiple threads, we'll need to have a multi-threaded execution engine anyway, so we can probably have the multi-threaded execution engine perform multiple operations using the same CPU device. Let's save this discussion for a later commit when we tackle multi-threading the execution engine.

xunzhang:
@neubig Right, cool. I will finish this soon.

xunzhang:
This pull request is review-ready. It will not affect old code and interface and I will split remaining work in the future pull requests.

The remaining things include,
1. Apply the multi-device support to all operations: this is trivial but needs to be carefully coded.
2. Fix the multi GPUs hanging bug: this is not introduced by this pull request and should be applied with multi-device support someplace.
3. Python interface: the code in this pull request will not break current use I think.
4. Add tests and refactor failing GPU unit-tests, update the document.

dynet/exec.cc
```diff
@@ -122,6 +122,10 @@ const Tensor& SimpleExecutionEngine::incremental_forward(VariableIndex i) {
       }
       node->aux_mem = aux_mem;
 
+      // check consistent device
+      for (auto & xs_v : xs) {
+        DYNET_ASSERT(xs_v->device == nfxs[num_nodes_evaluated].device, "Attemp to do tensor forward in different devices");
```
neubig: Attemp -> Attempt. Similarly for all below.

dynet/exec.cc
```diff
@@ -245,15 +254,17 @@ void BatchedExecutionEngine::combine_tensors(std::vector<VariableIndex> batch_id
     const size_t sz = node2size[id];
 
     float* my_src = batches[node2batch[id]].nfx.v + node2offset[id];
+    if (tout.device->type == DeviceType::CPU) {
+      memcpy(dest, my_src, sz * sizeof(float));
+    } else {
```
neubig: For this "else", I think we should add a `tout.device->type == DeviceType::GPU`, then throw an error if we get a device other than the one we're expecting. We might add other device types later, and if we do this logic will break (and it's in a somewhat inconspicuous place, so we should make sure that it's easy to catch through an error).

dynet/expr.cc
```diff
@@ -44,15 +43,17 @@ Expression operator+(const Expression& x, const Expression& y) {
     else if (y.dim().batch_size() == 1)
         return Expression(x.pg, x.pg->add_function<ScalarAdd>({x.i, y.i}));
     else
-        return Expression(x.pg, x.pg->add_function<Sum>({x.i, y.i}));
+        return Expression(x.pg, x.pg->add_function<Sum>({x.i, y.i}, x.pg->nodes[x.i]->device));
```
neubig: This shouldn't need a "device" argument, right? It can just inherit its device from its inputs.

dynet/expr.cc
```diff
@@ -76,7 +77,10 @@ Expression contract3d_1d(const Expression& x, const Expression& y, const Express
 Expression sqrt(const Expression& x) { return Expression(x.pg, x.pg->add_function<Sqrt>({x.i})); }
 Expression abs(const Expression& x) { return Expression(x.pg, x.pg->add_function<Abs>({x.i})); }
 Expression erf(const Expression& x) { return Expression(x.pg, x.pg->add_function<Erf>({x.i})); }
-Expression tanh(const Expression& x) { return Expression(x.pg, x.pg->add_function<Tanh>({x.i})); }
+Expression tanh(const Expression& x, Device *device) {
```
neubig: Similarly, we should remove the device argument from this and all others below, except the ones that deal with input, or explicitly changing devices.

dynet/expr.h
```diff
@@ -35,17 +36,38 @@ struct Expression {
   VariableIndex i;
   unsigned graph_id;
 
-  Expression() : pg(nullptr), i(0), graph_id(0) { }
-  const bool is_stale() const {return (get_number_of_active_graphs() != 1 || graph_id != get_current_graph_id());}
+  Expression() : pg(nullptr), i(0), graph_id(0) {}
+
+  Expression(Device *device) : pg(nullptr), i(0), graph_id(0) {}
```
neubig: This is identical to the empty constructor, so we should delete it.

dynet/expr.h
```diff
@@ -35,17 +36,38 @@ struct Expression {
...
+    i(i), graph_id(pg->get_id()) {}
+  /**
+   * \brief Base expression constructor
+   * \details Used when creating operations
+   *
+   * \param pg Pointer to the computation graph
+   * \param i Variable index
+   * \param device Device placement for the Expression
+   */
+  Expression(ComputationGraph *pg, VariableIndex i, Device *device) : pg(pg),
```
neubig: Again, this is identical to the constructor above and should be deleted.

dynet/expr.h
```diff
@@ -35,17 +36,38 @@ struct Expression {
...
+    i(i), graph_id(pg->get_id()) {}
+
+  inline std::string get_device() const { return pg->nodes[i]->device->name; }
```
neubig: Maybe `get_device_name()`? Also, this should check if the device is a null pointer, at least with a `DYNET_ASSERT()` call.

dynet/expr.h
```diff
@@ -35,17 +36,38 @@ struct Expression {
...
+
+  const bool is_stale() const {
+    return (get_number_of_active_graphs() != 1 || graph_id != get_current_graph_id());
+  }
+
+  // TODO
```
neubig: Delete.

dynet/expr.h
```diff
@@ -103,7 +125,23 @@ Expression f(const T& xs, const T1& arg1) {
   for (auto xi = xs.begin(); xi != xs.end(); ++xi) xis[i++] = xi->i;
   return Expression(pg, pg->add_function<F>(xis, arg1));
 }
+template <typename F, typename T>
```
neubig: Give this a more descriptive name. Also it might be good to comment, unless what the function is doing is obvious from the name.

dynet/nodes-affinetransform.h
```diff
@@ -3,12 +3,14 @@
...
 #include "dynet/dynet.h"
 #include "dynet/nodes-macros.h"
+#include "dynet/devices.h"
 
 namespace dynet {
 
 // y = x_1 \sum_{i=2, 4 ...} A_i * x_{i+1}
 struct AffineTransform : public Node {
   template <typename T> explicit AffineTransform(const T& a) : Node(a) {}
+  template <typename T> explicit AffineTransform(const T& a, Device *device) : Node(a, device) {}
```
neubig: Remove this, and just get the device from the previous nodes' device.

dynet/nodes-arith-sum.h
```diff
@@ -3,12 +3,14 @@
...
 #include "dynet/dynet.h"
 #include "dynet/nodes-macros.h"
+#include "dynet/devices.h"
 
 namespace dynet {
 
 // y = \sum_i x_i
 struct Sum : public Node {
   template <typename T> explicit Sum(const T& a) : Node(a) {}
+  template <typename T> explicit Sum(const T& a, Device *device) : Node(a, device) {}
```
neubig: Similarly, remove the device option.

dynet/nodes-change-devices.cc
```diff
@@ -0,0 +1,48 @@
+#include <vector>
```
neubig: Might want to rename this file `nodes-to-device.cc`?

dynet/nodes-linalg.h
```diff
@@ -19,7 +21,7 @@ struct Transpose : public Node {
 // y = inv(x)
 // x = an invertible matrix
 struct MatrixInverse : public Node {
-  explicit MatrixInverse(const std::initializer_list<VariableIndex>& a) : Node(a) {}
+  explicit MatrixInverse(const std::initializer_list<VariableIndex>& a, Device *device) : Node(a, device) {}
```
neubig: Again, no device here.

dynet/nodes-matrixmultiply.h
```diff
@@ -3,12 +3,14 @@
...
 #include "dynet/dynet.h"
 #include "dynet/nodes-macros.h"
+#include "dynet/devices.h"
 
 namespace dynet {
 
 // y = x_1 * x_2
 struct MatrixMultiply : public Node {
-  explicit MatrixMultiply(const std::initializer_list<VariableIndex>& a) : Node(a) {}
+  explicit MatrixMultiply(const std::initializer_list<VariableIndex>& a,
```
neubig: No device here.

dynet/nodes-trig.h
```diff
@@ -3,12 +3,13 @@
...
 #include "dynet/dynet.h"
 #include "dynet/nodes-macros.h"
+#include "dynet/devices.h"
 
 namespace dynet {
 
 // y = tanh x_1
 struct Tanh : public Node {
-  explicit Tanh(const std::initializer_list<VariableIndex>& a) : Node(a) {}
+  explicit Tanh(const std::initializer_list<VariableIndex>& a, Device *device) : Node(a, device) {}
```
neubig: No device here.

dynet/tensor.cc
```diff
@@ -94,11 +95,13 @@ float TensorTools::access_element(const Tensor& v, int index) {
 }
 
 float TensorTools::access_element(const Tensor& v, const Dim& index) {
+  if (v.device->type == DeviceType::CPU) {
+    return (*v)(index[0], index[1]);
+  } else {
 #if HAVE_CUDA
```
neubig: Similarly here, I think we should check if the device is GPU, and throw an error if we get an unknown device.

dynet/tensor.cc
```diff
@@ -152,6 +156,28 @@ void TensorTools::copy_elements(const Tensor& v, const Tensor& v_src) {
...
+void TensorTools::copy_elements(Tensor& v, const Tensor& v_src) {
+  if (v.device->type == DeviceType::CPU) {
+    if (v_src.device->type == DeviceType::CPU) {
+      memcpy(v.v, v_src.v, sizeof(real) * v.d.size());
+    } else {
+#if HAVE_CUDA
+      cudaMemcpyAsync(v.v, v_src.v, sizeof(real) * v.d.size(), cudaMemcpyDeviceToHost);
+#endif
+    }
+  } else {
```
neubig: Similarly here, check if device is GPU explicitly.

dynet/expr.h
xunzhang: Thanks! All comments seem reasonable. But here, if we infer the device type of inside the executorEnginer(in the place where initializing the `nfxs`), the `get_device` call will get an error before running `forward`, right?

xunzhang: A solution for this is to create a variable to record device in ComputationGraph class I think.

neubig: I don't understand what you mean here. My comment was just that you should change the name of the function, and also make sure that if we call this when `pg->nodes[i]->device` is null, we should get an error instead of a segmentation fault.

neubig:
Thanks! I confirmed that this is working as expected, so I merged. This is great to have :)

duyvuleo:
Is it working actually? I tried the "./examples/train_xor-multidevice" example and got the following error:

> terminate called after throwing an instance of 'std::runtime_error'
>   what():  Invalid device name: GPU:0
> Aborted (core dumped)

neubig:
I think documentation is not finished yet, but you need to add `--dynet-devices CPU,GPU:0` to the command line I think.

duyvuleo:
It works. Thanks!

PR #707, opened 2017-07-17
Add more info to Readme

pmichel31415:
This PR adds info about install, getting started, etc...  directly into the readme.

Also puts the build/doc/pip status bar on top and centers the logo.

I'm not sure what should be less/more verbose, especially the installation and getting started parts, so feel free to modify it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The Dynamic Neural Network Toolkit
 
 - [General](#general)
 - [Installation](#installation)
-  - [C++](#c++-installation)
+  - [C++](#c-installation)
   - [Python](#python-installation)
 - [Getting Started](#getting-started)
 - [Citing](#citing)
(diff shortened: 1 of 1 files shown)


neubig:
Hmm, I intentionally removed this from the README to prevent divergence from the doc on readthedocs. Do you think this is necessary? If so, we'll have to have some strategy to make sure that both places get updated when changes are made.

pmichel31415:
I still think it's more convenient for new users that come to the github first.

I recon synchronizing things might be an issue.

One solution I've been considering is including the readme.md in the documentation as the home page or something like this.

I think this could be done in readthedocs using pandoc to convert from md to rst. As a corollary, other md files could be included in the docs, eg. the scala readme.

I'll give it a try once I have a bit more time.

neubig:
OK. For now, I'll merge this though. Thanks!

PR #709, opened 2017-07-18
Clarifying how to do learning rate decay

jkkummerfeld:
Modified comment to describe the current strategy for learning rate decay.


neubig:
Thanks!

PR #710, opened 2017-07-19
Refactror rnnlm.py to run on current API

akkikiki:
rnnlm.py is currently failing due to some API changes. 
Confirmed that it is running on ab31157b2f150a198fc264640fad56c798a619ab (i.e. latest commit as of now).

### DIFF
--- a/examples/python/rnnlm.py
+++ b/examples/python/rnnlm.py
@@ -22,10 +22,10 @@ class RNNLanguageModel:
         self.bias = model.add_parameters((VOCAB_SIZE))
 
     def save_to_disk(self, filename):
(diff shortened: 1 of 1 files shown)


examples/python/rnnlm.py
```diff
@@ -98,7 +98,7 @@ def sample(self, first=1, nchars=0, stop=-1):
     VOCAB_SIZE = vocab.size()
 
     model = Model()
```
pmichel31415: Ideally this should also be updated to `ParameterCollection()`

neubig: @pmichel31415 Good point! I think we need to make a full sweep through the Python examples to make them consistent.

PR #712, opened 2017-07-19
add verbose error message to even-length BiRNN hidden dim assertion

john-hewitt:
This may be domain knowledge, but I was unaware that the hidden dimension of a BiRNN was considered to be split between the two constituent RNNs (instead of, say, assigning the full dimension assigned to each.)

When an odd dimension is given to BiRNN builder 
 - *Previous behavior:* die with Python assertion, no message 
 - *New behavior:*  die with Python assertion, descriptive message

I also added the constraint (and splitting-the-dimension behavior) to the BiRNN docs.


neubig:
Thanks!

PR #713, opened 2017-07-19
[WIP] ThreadPool Device Support

zhisbug:
This PR adds the support for Eigen ThreadPool device. Users can enable the feature and set the number of threads by setting the -DTHREADPOOL flag in CMake.

TODOs:
- [x] Some conflicts to be resolved.
- [x] Hong reviewed.
- [ ] Graham reviewed. 

Additional comments: due to some engineering difficulties, I have to remove a forward declaration in devices.h, see this commit: https://github.com/clab/dynet/pull/713/commits/7f0819768c207738a228a5b64085f00b1bc1abaa , which might increase the compilation time a bit. The reason is that I have to create an `Eigen::ThreadPool* tp` as a member variable inside `class Device_ThreadPool`. 
However, as this struct is a typedef'ed struct which cannot be forward declared in C++. A workaround is to directly new an `Eigen::ThreadPool` here: https://github.com/Abasyoni/dynet/blob/tmp-branch/dynet/devices.cc#L137, but this probably suffers a memory leak issue.


zhisbug:
Conflicts resolved. @neubig @xunzhang Could you please review the code? (especially in devices.h where I remove the forward declarations).

xunzhang:
The compile check failed, could you take a look first? Thanks!
BTW, I am a little worried about the conflicts with #704 after this has been merged. What do you think about the order to check in between this and #704?

dynet/devices.h
```diff
@@ -1,19 +1,16 @@
...
-namespace Eigen {
-  struct DefaultDevice;
-  class CudaStreamDevice;
-  struct GpuDevice;
-}
-
 namespace dynet {
 
-enum class DeviceType {CPU, GPU};
+enum class DeviceType {CPU, GPU, ThreadPool};
```
xunzhang: A little confused here, it seems that ThreadPool is a sub-option for both CPU and GPU(correct me if I am wrong). But here it seems it is a third option after CPU and GPU. What's the relationship between it with CPU and GPU? I don't suggest to add an enum value here for ThreadPool.

dynet/tensor.cc
```diff
@@ -19,7 +19,7 @@ namespace dynet {
 #ifndef __CUDACC__
 
 ostream& operator<<(ostream& os, const Tensor& t) {
-  if (t.device->type == DeviceType::CPU) {
+  if ((t.device->type == DeviceType::CPU) || (t.device->type == DeviceType::ThreadPool)) {
```
xunzhang: What is the reason in `model.cc` use `else if` but here use `||` together with the CPU if branch?

CMakeLists.txt
```diff
@@ -132,6 +132,11 @@ if(ENABLE_BOOST)
   endif()
 endif()
 
+# if this flag is set, Eigen will use ThreadpoolDevice with the number of thread as specified
+if (THREADPOOL)
```
xunzhang: So overall it will use the thread-pool functionality of Eigen code, Right? Could you please add some tests and show sort of the performance gain after opening this? If that is the case, could we open it by default when using Eigen + GPU, or so? If I am a user, I may feel uncertain about the relationship between `-DBACKEND=cuda` with `-DTHREAD_POOL`.

dynet/tensor.cc
zhisbug: Because they are different. Print functions can not be multi-threaded.

dynet/devices.h
zhisbug: You can understand ThreadPool as a device which relies on CPU but using multiple threads.

CMakeLists.txt
zhisbug: See my previous answer. I agree with you that we can enable it as default. @neubig comments are welcome.

neubig:
Thanks, this is great! I'm a little bit concerned about conflicts with https://github.com/clab/dynet/pull/704 though. I think we can try to integrate #704 first, then fit the threadpool device within this framework. I'll do a full review at that point.

neubig:
Sorry this was delayed so long! @xunzhang's multi-device pull has been merged, so maybe we can reconcile the threadpool code with his now.

xunzhang:
Will close this and file another pull request.

PR #716, opened 2017-07-20
Added transduce and add inputs to VanillaLSTM

philip30:
This will make the BiRNN and VanillaLSTM builder can be used interchangeably.

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -4107,6 +4107,12 @@ cdef class VanillaLSTMBuilder(_RNNBuilder): # {{{
             self.thisvanillaptr = self.thisptr = new CVanillaLSTMBuilder()
         self.cg_version = -1
 
+    def transduce(self, inputs):
+      return self.initial_state().transduce(inputs)
+
+    def add_inputs(self, inputs):
+      return self.initial_state().add_inputs(inputs)
+
     @property
     def spec(self): return self._spec
 


yoavg:
I'm not sure I like the idea of having transduce/add_inputs on the builder and not on the state (why is it needed?), but if we do it, might as well put it in the base builder so it is available for all RNN builders.

philip30:
I actually just saw that there was one in the BiRNN builder, so I put it here. Also, I think it is a good idea to have the capability to build a graph based on a sequence input.

I think the state should be encapsulated in the builder, because if we can modify the state directly, why do we need the builder anyway? This is just my opinion, need feedback though :)

neubig:
I think the bigger issue is just that the interface is inconsistent, where in BiLSTMBuilder `transpose()` is associated with the builder, and LSTMBuilder it's associated with the state.

neubig:
@yoavg Any thoughts on making the interface more consistent? This is a Python binding thing, so I'll leave the decision up to you.

philip30:
I think because there are still some issues, I will revert this request. Thank you!

yoavg:
Some (late, sorry) comments on this issue:

- In the RNNs, it is essential for the transduce / add_inputs to be on the state, because of the flexibility to transduce from arbitrary states, and not just the initial one. It is also needed for incremental cases (i.e. truncated BPTT). 

- In the bi-rnn builder, someone added this functionality on the builder without me noticing, causing an inconsistency with the LSTM interface. In retrospect, I am not really opposed to this, especially as I do not see a way to make the analogue case (transduce on the state) for the bi-rnn, as its not clear what the analogue of state is.

- The current PR suggests adding the functionality *also* to the LSTMBuilder. The upside is the LSTM and BiLSTM builders can be used interchangeably in some cases. The downside is redundancy in the API that can lead to confusions. I tend against this API change, but not strongly so, and am open to being convinced otherwise.

- In any case, the functionality should be implemented in the base RNN class and not the specific builders.

Issue #636, opened 2017-06-27
how to use a specific initializer for VanillaLSTMBuilder?


rasoolims:
I was not able to find an example. For example, how can I use a UniformInitializer with a specific stdv?

pmichel31415:
I don't think this is possible for now, although it would be cool if it were.

One solution is to set the value of the parameters manually after intialization.

If you are using python, something like:

    lstm=dy.VanillaLSTMBuilder(...)
    lstm.get_parameters()[0][0].set_value(np.random.uniform(...))

pmichel31415:
Actually this is a duplicate of #529

jcyk:
@pmichel31415  I tried the way you suggested. It does seem to work on GPU. Any idea?

Issue #529, opened 2017-05-07
How to change LSTM ParameterInit?


speedcell4:
if I want to init LSTM parameter with other initialization method, e.g. [`Orthogonal`](https://keras.io/initializers/#orthogonal) in Keras, how can I do that?

neubig:
I think at the moment adjusting the LSTM initialization is not supported directly in DyNet's LSTMBuilders, but this is something that might be nice to support.

Also, I believe `ParameterInitSaxe` does orthonormal initialization, but is not supported on GPU at the moment.

yoavg:
As a workaround, you could get the parameter matrices from the LSTMBuilder, and change their values. This is a bit cumbersome and less efficient than init (because you set the values twice) but it happens just once in the beginning, so should be fine. (this will be tricky to do at the moment using the python bindings, but should be doable with the cpp version).

danielhers:
Has anyone done what @yoavg suggested here and could share an example? Now it should be possible in Python too, I think.

PR #722, opened 2017-07-21
allow Parameter.set_value to work on both GPU and CPU

jcyk:
Old version of `set_value()`  does not work on GPU. This PR may fix it.

Give a way to set the value of the parameters manually after initialization is useful.
For example, manully initialize an LSTM #529 and #636


neubig:
Thanks! This is a good idea. Could you add a unit test to `test-params.cc` that ensures that this functionality is working correctly?

neubig:
Hmm, the Python test is giving me a segfault on GPU (but not on CPU):

    test_layer_norm (test.TestOperations) ... ok
    test_grad (test.TestParameters) ... ok
    test_is_updated (test.TestParameters) ... ok
    test_set_updated (test.TestParameters) ... ok
    test_set_value (test.TestParameters) ... Segmentation fault (core dumped)

Are you able to reproduce this? If not, I can take a look.

jcyk:
Uh. I think it is because my transfer from numpy.ndarray to vector<float> is too naive. It only works when type of value is float32 and in successive memory. There is a solution in _dynet.pyx (https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L475 )but commented with "to make more efficient ". Any idea?

jcyk:
@neubig 
I tested on my machine, but could not reproduce the segfault, things just work fine.
It seems the direct transfer is okay at least in my machine.

neubig:
Sorry it took so long to get back to this. I tried one more time and the segmentation fault was resolved. I'm guessing that my environment was broken when I tried before, so it looks like things are working. Thanks!

Issue #715, opened 2017-07-20
Saving a model causes a large spike in memory


neubig:
I have a simple DyNet program:

    import dynet as dy
    @profile
    def main():
      pc = dy.ParameterCollection()
      param = pc.add_parameters((1024,1024,4))
      pc.save("pc.txt")    
    main()

when run with the Python memory profiler it gives the following result:

    [dynet] random seed: 2179783019
    [dynet] allocating memory: 512MB
    [dynet] memory allocation done.
    Filename: saving.py
    
    Line #    Mem usage    Increment   Line Contents
    ================================================
         3  540.289 MiB    0.000 MiB   @profile
         4                             def main():
         5  540.297 MiB    0.008 MiB     pc = dy.ParameterCollection()
         6  540.348 MiB    0.051 MiB     param = pc.add_parameters((1024,1024,4))
         7  848.461 MiB  308.113 MiB     pc.save("pc.txt")

We can see that just saving the model causes the memory usage to greatly increase, which may be a problem with people running out of memory when they try to save their models. I think this is a result of how things are implemented within the DyNet serialization code. @xunzhang any ideas about how to solve this without losing functionality?

xunzhang:
Could you reproduce it? In my environment, the result are below. The first is CPU mode and the second is GPU mode. The memory profiler version I use is 0.4.7.

```
Filename: test.py

Line #    Mem usage    Increment   Line Contents
================================================
     3  558.301 MiB    0.000 MiB   @profile
     4                             def main():
     5  558.305 MiB    0.004 MiB   	pc = dy.ParameterCollection()
     6  558.742 MiB    0.438 MiB   	param = pc.add_parameters((1024, 1024, 4))
     7  559.730 MiB    0.988 MiB   	pc.save('pc.txt')
```

```
Filename: test.py

Line #    Mem usage    Increment   Line Contents
================================================
     2  543.094 MiB    0.000 MiB   @profile
     3                             def main():
     4  543.094 MiB    0.000 MiB     pc = dy.ParameterCollection()
     5  543.531 MiB    0.438 MiB     param = pc.add_parameters((1024,1024,4))
     6  543.836 MiB    0.305 MiB     pc.save("pc.txt")    
```

redpony:
I suspect this will vary significantly with the version of the standard
library that is being used. but the way we write to a std::string to figure
out how many bytes each parameter is going to need on disk is very
inefficient. If we always write n digits per float, we can easily
precompute this and stop using strings.

[...]

PR #723, opened 2017-07-21
Made text saving be fixed-width for efficiency

neubig:
Addresses  https://github.com/clab/dynet/issues/715 by making floats fixed-width in the saved file and pre-calculating the size in bytes:

Here are some stats to help whether this is a good way of doing things:
* Memory required on save for above example: Old=308MB, New=16MB
* Space on disk required for above example: Old=50MB, New=64MB
* Time required for `tests/test-io`'s `test_save1_perf`: Old=9.40s, New=9.28s (stddev: ~0.14)

So this is faster, and more importantly doesn't have the memory spike, but results in a non-trivial increase in disk space. I'd tend to favor incorporating this, as memory spiking on model saving could cause processes to crash or hang after an epoch of training, which could be quite annoying.

### DIFF
--- a/dynet/io.cc
+++ b/dynet/io.cc
@@ -47,6 +47,8 @@ TextFileSaver::TextFileSaver(const string & filename, bool append) :
         datastream(filename, append ? ofstream::app : ofstream::out) {
   if(!datastream)
     DYNET_RUNTIME_ERR("Could not write model to " << filename);
+  datastream.precision(FLOAT32_PRECISION);
+  datastream << std::scientific << std::showpos;
 }
 
 void TextFileSaver::save(const ParameterCollection & model,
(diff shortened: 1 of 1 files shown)


dynet/io.cc
```diff
@@ -47,6 +47,8 @@ TextFileSaver::TextFileSaver(const string & filename, bool append) :
         datastream(filename, append ? ofstream::app : ofstream::out) {
   if(!datastream)
     DYNET_RUNTIME_ERR("Could not write model to " << filename);
+  datastream.precision(FLOAT32_PRECISION);
+  datastream << std::scientific << std::showpos;
```
xunzhang: `std::showpos` here is good.

dynet/io.cc
```diff
@@ -84,37 +86,30 @@ void TextFileSaver::save(const LookupParameter & param,
 
 void TextFileSaver::save(const ParameterStorage & p,
                          const string & key) {
-  std::ostringstream buffer;
-  buffer.precision(FLOAT32_PRECISION);
-  buffer << dynet::as_vector(p.values) << endl;
+  datastream << "#Parameter# " << (key.size() > 0 ? key : p.name) << ' ' << p.dim << ' ';
+  size_t strsize = p.dim.size() * (FLOAT32_PRECISION + 8) + 1;
```
xunzhang: If I understand correct, 8 here is for `+x.e-xx`. I think the value range is fine if there is no the spefical case.

xunzhang: According to [this](http://www.cplusplus.com/reference/ios/scientific/), it says there is three exponential digits, but on my machine, there is only two.

neubig:
I agree that this was not clear. But 32-bit floats only have 2 digits worth of exponents (the example you cited is using doubles, 64-bit floats that have three digits). This will need to be made more flexible if DyNet is to support multiple levels of floating-point precision, but that is far in the future.

PR #730, opened 2017-07-22
fix #714

### LINKED ISSUE #714: Need to precisely set the amount of memory for FOR,BACK,PARAM?
Hi, 
Initial memory allocation really matters!
Though the memory use for Dynet will be expanded automatically,  I found I need to precisely set the memory allocation at the initialization stage (```--dynet-mem MEM```) to get my code work through my dataset. Otherwise, I will see `CUDA failure in cudaMalloc(&ptr, n)
out of memory`

It is a pain to do so. 
Also the command line actually accepts either one or four arguments
`--dynet-mem MEM`
`--dynet-mem FOR,BACK,PARAM,SCRATCH`
The document should be updated.

jcyk:
Expand memory pool by a multiple of 16MB every time.

### DIFF
--- a/doc/source/commandline.rst
+++ b/doc/source/commandline.rst
@@ -8,13 +8,13 @@ specified at the very beginning of the command line, before other
 options.
 
 -  ``--dynet-mem NUMBER``: DyNet runs by default with 512MB of memory,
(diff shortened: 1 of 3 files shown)


neubig:
Thanks for fixing this!

PR #731, opened 2017-07-22
Better way to set configuration from python.

yoavg:
Instead of the `import _dynet/_gdynet` and then setting `DynetParams` followed by
`init`, vs `import dynet` with commandline args, we can now do:

```
    import dynet # as before
```

or:
```
    import dynet_config
    dynet_config.set_gpu()
    dynet_config.set(mem=4,random_seed=9)
    import dynet
```

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -29,6 +29,7 @@ if(PYTHON)
   
   # Things to copy the files
   file(GLOB DepFiles ${CMAKE_CURRENT_SOURCE_DIR}/dynet.py
+                     ${CMAKE_CURRENT_SOURCE_DIR}/dynet_config.py
                      ${CMAKE_CURRENT_SOURCE_DIR}/_dynet.pyx
                      ${CMAKE_CURRENT_SOURCE_DIR}/_dynet.pxd
                      ${CMAKE_CURRENT_SOURCE_DIR}/dynet_viz.py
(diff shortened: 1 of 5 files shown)


pmichel31415:
Thanks, that looks like a better and more intuitive way to go about it

neubig:
This is a good idea, but we might need to rethink it once the (coming very soon) multi-device support is implemented. Do you think it's worth merging now? Or shall we think again once multi-device is working and we start doing things like removing the `libgdynet.so` library and instead just use a single library for both CPU and GPU?

yoavg:
I'm happy waiting, but in any case it'd be nice to have a transparent mechanism for controlling the commandline arguments, esp. memory allocation, from python code.

neubig:
OK, multi-device support is working now! It actually looks like this pull has no conflicts with the multi-device support, but perhaps we should check to make sure that nothing is broken. https://github.com/clab/dynet/pull/704

pmichel31415:
I didn't follow the multi-device PR super closely, but does it still make sense now to have two modules `_dynet` and `_gdynet`?

Like if the module has been compiled for GPU isn't it possible to specify cpu usage in the command line? Not very familiar with the details

neubig:
Yes, @xunzhang will probably be refactoring this in a future PR to create a single library.

pmichel31415:
Is this still relevant with #781 ?

xunzhang:
Oh, I added `dynet_config` in https://github.com/clab/dynet/pull/781. I think now that works and this can be closed.

PR #740, opened 2017-07-25
Fix readme to match http://dynet.readthedocs.io/en/latest/install.html

altosaar:
There was a `cmake` flag missing from the README, and a git checkout.

Without the flag, it didn't compile the tutorials and examples and I got confused.


neubig:
Thanks!

PR #744, opened 2017-07-25
Create a virtual cudnn to be compatible with different versions of cuDNN

zhisbug:
Allow compatibility with different versions of CUDNN.


zhisbug:
This should be ready unless you want to take a second look.. @neubig @xunzhang

neubig:
Thanks, will try to do a review today.

dynet/cudnn-ops.cu
```diff
@@ -88,35 +89,13 @@ void CudnnConvOp::forward_impl(const Device_GPU& dev, const std::vector<const Te
...
-  CUDNN_CHECK(cudnnSetConvolution2dDescriptor(conv_desc_,
-              pad_w/2, pad_h/2, stride_[1], stride_[0], 1, 1,
-              CUDNN_CROSS_CORRELATION, CUDNN_DATA_FLOAT));
-#else
-  CUDNN_CHECK(cudnnSetConvolution2dDescriptor(conv_desc_,
-              pad_w/2, pad_h/2, stride_[1], stride_[0], 1, 1,
-              CUDNN_CROSS_CORRELATION));
-#endif
+  //set cudnn descriptors
+  setTensor4dDescriptor(&x_desc_, XN, XC, XW, XH);
```
neubig: I think it'd be better to be consistent with the use of "descriptor" and "desc". I can see people being confused about which one to use. If there is no strong preference, I'd suggest exactly matching the CUDNN name, minus the `cudnn`.

zhisbug: Sure, good suggestion!

zhisbug:
I think it is ready to be merged now.

neubig:
Thanks! I've merged this.

PR #745, opened 2017-07-26
Fix bug if incremental_forward(i) called for i<num_nodes_evaluated.

mattr1:
Also update get_value(i) to only do incremental_forward to the variable requested.


neubig:
Thanks!

PR #746, opened 2017-07-26
Verify the deletion of nodes for cg::revert.

zzsfornlp:
This is about fixing the possible memory leak problem for the checkpoint/revert mechanism, which might get severe when lots of nodes are created between checkpoint() and revert().


neubig:
Good catch, thanks!

PR #749, opened 2017-07-28
Make conv2d with a single channel accept 2-dimensional inputs

neubig:
Sometimes, particularly when handling text, we'll have a 2-dimensional array as input to conv2d, which is treated as having a single channel. This allows us to pass this as-is without reshaping to a 3-dimensional tensor.


neubig:
I'm going to go ahead and merge this because I've confirmed that it's working both in tests and on real data and need it to do something. (Please tell me if you notice anything off)

PR #765, opened 2017-08-07
Restart method for optimizers

pmichel31415:
This adds a restart method to optimizers to reinitialize the momentum/rescaling parameters to 0.

Also adds python bindings

### DIFF
--- a/dynet/training.cc
+++ b/dynet/training.cc
@@ -322,9 +322,9 @@ void RMSPropTrainer::alloc_impl() {
 
 
 void RMSPropTrainer::restart() {
-  for (auto sp : hsmg)
+  for (auto sp : hmsg)
     TensorTools::zero(sp.h);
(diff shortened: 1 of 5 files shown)


neubig:
Thanks! Does this have any advantage over just calling the constructor and creating a new trainer?

pmichel31415:
Does the memory of shadow-parameters get freed properly? I assumed it wasn't the case (like for standard parameters), hence it's more efficient to keep the same shadowparams and just reset their value)

neubig:
That may be the case. Thanks for the clarification!

pmichel31415:
OK I also added the option of setting a new learning rate as an argument.

I added a correctness test for the MomentumSGD which required making the velocity shadow-param public but I think that's OK.

Anyway for me it's good to merge once the tests are done

PR #766, opened 2017-08-08
A possible method to pre-determine the needed memory.

zzsfornlp:
This commit concerns the part of memory allocation, there are mainly two changes.

1. Record the maximum used memory `peak_used` for each MemoryPool and re-allocate memory according to this one rather than capacity (there might be some fragments).

2. Add an `--dynet-mem-test` flag to enter a mode for only memory testing. In this mode, all forward and backward calculations are skipped (using a global var `mem_test_flag`, maybe not an elegant way). Also, it will report the memory usage and suggested memory value for each MemoryPool whenever `AlignedMemoryPool::clear` is called.

The running could be finished in two steps, using python for example:
1. First determine the needed memory for each Memory-Pool (this run will be fast since there will be no real calculations, and you can cut it at any time once the rough memory sizes are determined):
`python run.py [other-options] --dynet-mem 4 --dynet-mem-test`
Some of the memory information will be printed like this:
`[dynet-mem-test] CPU forward memory, Previously (in MB): #pools/#last-used/#peak-used/#last-cap = 1/676/3338/3500; suggesting 3338MB.`
`[dynet-mem-test] CPU backward memory, Previously (in MB): #pools/#last-used/#peak-used/#last-cap = 1/581/3244/3500; suggesting 3244MB.`
`[dynet-mem-test] CPU parameter memory, Previously (in MB): #pools/#last-used/#peak-used/#last-cap = 45/654/0/769; suggesting 769MB.`
`[dynet-mem-test] CPU scratch memory, Previously (in MB): #pools/#last-used/#peak-used/#last-cap = 1/0/0/1; suggesting 1MB.`
2. Next will be the real run with exact memory setting and no `--dynet-mem-test` flag.
`python run.py [other-options] --dynet-mem 3500,3500,800,10`

Although currently the MemoryPool allows dynamic growing & memory re-allocating and pre-determing memory sizes should only make the first run more efficient, for reasons I have not figured out, I still find it will cost more memory, sometimes +1GB, than that when pre-determing the exact sizes for the pools using `--dynet-mem`. Thus it might be helpful to pre-determine the needed memory, which could be also utilized for the mechanisms of auto-batch and reverting. This might not be the best way for doing this and only a small step towards memory-profiling, but it might be a convenient and transparent way for almost any running cases.


neubig:
Thanks! But I think the better way to do this is to figure out why it's using more memory than expected, instead of trying to work around what might potentially be a bug that is causing excess memory usage.

Could you create a minimal example of code/command line arguments that causes the problem so we can debug?

zzsfornlp:
Here is a piece of code with which that I encounter the problem (running on Ubuntu-14.04 with CPU):

    import dynet as dy
    import random, os
    N = 1000
    ITER = 5
    STEP = 50
    BATCH = 4000
    m = dy.Model()
    w = m.add_parameters((N,N))
    ini = [random.random() for _ in range(N*BATCH)]
    for i in range(ITER):
        dy.renew_cg()
        ww = dy.parameter(w)
        h_rec = dy.inputVector(ini)
        h_add = dy.inputVector(ini)
        h_rec = dy.reshape(h_rec, (N,), batch_size=BATCH)
        h_add = dy.reshape(h_add, (N,), batch_size=BATCH)
        for s in range(STEP):
            h_rec = ww * h_rec + h_add
        loss = dy.dot_product(h_rec, h_add)
        loss = dy.sum_batches(loss)
        loss.backward()
        print("Step: %s-end" % (i,))
        os.system("cat /proc/%s/status | grep VmRSS" % os.getpid())
        os.system("nvidia-smi | grep %s" % os.getpid())

When running in memory-growing mode, it will reach to 7GB Resident Set Size.
`python test.py --dynet-mem 4`
But when setting enough memory (without the need of re-allocating), it will be about 5GB.
`python test.py --dynet-mem 2000,2000,200,10`

Strangely enough, this problem only happens with the python-version on CPU, when run on GPU or using c++, things seem fine. My initial guess is the fragments of the InternalMemoryPool slots, but that seems only accounts for at most hundreds of MBs. In fact, I'm not quite sure with this problem, could you please provide some help with this?

neubig:
Sorry about being late on this! I was planning on taking a look. Has the problem been resolved?

zzsfornlp:
Sorry for the late reply. In fact, I'm not quite sure about this and thus close this one.

I assume dynet takes extra memory by looking at RES memory using `top` or `"cat /proc/<pid>/status | grep VmRSS"`. The "problem" seems to be caused by no CPU memory freeing when freeing old memories of the `InternalMemoryPools`. But actually when the system runs out of memory, it seems to release some and there might be no troubles at all (despite the somewhat scaring reporting of the system). Thus I'm not sure whether this is really a problem or is it just because of some properties of my machine?

Also, since this only happens with CPU+python, wonder whether this is some kind of late releasing interactions between system-malloc and python/cython?

By the way, the problem happens when allocating large blocks of memory (simulated by large batch_size in the example) up to MBs, therefore maybe fragmentation also plays a role in it. And thus point 1 is about recording the maximum used memory peak_used and set later memories according to this.

PR #769, opened 2017-08-08
Replace node Zeroes with Constant and add "constant" expression

pmichel31415:
Also 

- `zeroes` -> `zeros` (same as numpy), also [better](https://english.stackexchange.com/a/3825)
- add `ones` and `constant` expressions

### DIFF
--- a/dynet/expr.cc
+++ b/dynet/expr.cc
@@ -26,7 +26,10 @@ Expression const_lookup(ComputationGraph& g, LookupParameter p, unsigned index) 
 Expression const_lookup(ComputationGraph& g, LookupParameter p, const unsigned* pindex) { return Expression(&g, g.add_const_lookup(p, pindex)); }
 Expression const_lookup(ComputationGraph& g, LookupParameter p, const vector<unsigned>& indices) { return Expression(&g, g.add_const_lookup(p, indices)); }
 Expression const_lookup(ComputationGraph& g, LookupParameter p, const vector<unsigned>* pindices) { return Expression(&g, g.add_const_lookup(p, pindices)); }
-Expression zeroes(ComputationGraph& g, const Dim& d) { return Expression(&g, g.add_function<Zeroes>(d)); }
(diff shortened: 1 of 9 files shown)


neubig:
If I remember correctly, `TensorTools::zero()` is actually a bit faster than `TensorTools::constant()`. Have you profiled this by any chance?

Also, at least make `zeroes()` throw an error that the function was deprecated instead of deleting it altogether.

pmichel31415:
hmm OK I'll modify the forward pass a bit to use zero then.

I haven't deleted `zeroes` though, it's just removed from the doc but I've kept it backward compatible

pmichel31415:
Should be fixed now @neubig

dynet/expr.cc
```diff
@@ -26,7 +26,10 @@ Expression const_lookup(ComputationGraph& g, LookupParameter p, unsigned index)
 Expression const_lookup(ComputationGraph& g, LookupParameter p, const unsigned* pindex) { return Expression(&g, g.add_const_lookup(p, pindex)); }
 Expression const_lookup(ComputationGraph& g, LookupParameter p, const vector<unsigned>& indices) { return Expression(&g, g.add_const_lookup(p, indices)); }
 Expression const_lookup(ComputationGraph& g, LookupParameter p, const vector<unsigned>* pindices) { return Expression(&g, g.add_const_lookup(p, pindices)); }
-Expression zeroes(ComputationGraph& g, const Dim& d) { return Expression(&g, g.add_function<Zeroes>(d)); }
+Expression zeros(ComputationGraph& g, const Dim& d) { return Expression(&g, g.add_function<Constant>(d, 0.f)); }
+// Expression zeroes(ComputationGraph& g, const Dim& d) {return zeros(g, d);}
```
neubig: It looks like "zeroes" is commented out here?

pmichel31415: It is actually here as an inline function:
https://github.com/clab/dynet/pull/769/files#diff-dca57b9f24bff4f4b8ab89e90a3057a8R389

neubig: Ahh, OK. Thanks for the clarification!

Issue #739, opened 2017-07-25
using CompactVanillaLSTMBuilder with dropout produces error


jcyk:
I tried the new CompactVanillaLSTMBuilder, but when I set dropout on it, I got
```
CUDA failure in cudaMemcpy(val.v, t, sizeof(real) * val.d.size(), cudaMemcpyHostToDevice)
an illegal memory access was encountered
```
It only occurs on GPU.

jcyk:
hi!
For nodes in `Dynet`, does the output of `dim_forward` equal to `fx.d` ?

neubig:
@msperber, have you tested dropout on GPU? If not, could you take a look quickly?

msperber:
Hmm I had tried it and it worked for me.. @jcyk could post a minimal example so I can reproduce this?

jcyk:
Sorry for the late reply.
```
import dynet as dy
import numpy as np

dropout_x = 0.3
dropout_h = 0.3
batch_size = 5
inputs = np.random.randint(0, 100, (6, 5), dtype=np.int32)

pc = dy.ParameterCollection()
embs = pc.add_lookup_parameters((100, 10))
builder = dy.CompactVanillaLSTMBuilder(1, 10, 20, pc)

emb_inputs = [ dy.lookup_batch(embs, x) for x in inputs]
lstm = builder.initial_state()
builder.set_dropouts(dropout_x, dropout_h)
builder.set_dropout_masks(batch_size)
outputs = lstm.transduce(emb_inputs)
print outputs[-1].npvalue()
```
The error information:
```
CUDA failure in cudaMemcpy(&res[0], v.v, sizeof(real) * res.size(), cudaMemcpyDeviceToHost)
an illegal memory access was encountered
terminate called after throwing an instance of 'dynet::cuda_exception'
  what():  cudaMemcpy(&res[0], v.v, sizeof(real) * res.size(), cudaMemcpyDeviceToHost)
Aborted (core dumped)
```

msperber:
Thank you! I'm currently traveling but will take a look as soon as I have a chance.

msperber:
@jcyk I couldn't reproduce your problem, but would you mind checking if it still persists after the most recent commit?

jcyk:
@msperber  Thanks! But the new PR didn't fix the error occurs in my environment.

Note the same error information as above.

However, I tried the following modification, and the problem disappeared. 
[...]

PR #775, opened 2017-08-12
Compact-LSTM improvements

msperber:
This PR addresses the following:
- Improved speed by avoiding a broadcast operation and using temporary memory to avoid striding operations. I measured similar speed to the non-compact LSTM code on a seq2seq example on both CPU and GPU.
- New feature: vanilla_lstm_gates can now take several x_t inputs that are concatenated on-the-fly. This can help save memory because we can avoid concat operations which are often required before an LSTM, e.g. for bidirectional LSTMs, pyramidal LSTMs, ..; note that this might come at the expense of a slightly slower backward pass, since we cannot batch-compute backpropagation to the individual inputs, but I haven’t tested speed yet.
- A small bit of restructuring of temporary memory usage. This may be related to #739. Though I could not reproduce the error in the first place, it would be worth checking if this solved it.

A few notes:
- The code concerning autobatching could use a review since I don’t understand the details here
- The multi-input feature is not used by the LSTM builder yet. We could either leaves things as is and leave it to users to exploit that feature, or in a future PR add a ```add_input_concat()``` to either ```RNNBuilder``` or ```CompactVanillaLSTMBuilder```


msperber:
I re-added the non-list version to the interface as discussed, hope it'll be ok like this!

PR #776, opened 2017-08-13
Multidim operations & speed-improvements

msperber:
New features:
- change ```+``` operator to broadcast among arbitrary dimensions, including batch dimension
- extend ```cmul``` and ```cdiv``` to broadcast among arbitrary dimensions, including batch dimension
- broadcasting works as follows (similar to numpy):
  - When number of dimensions differ, we add dimensions of size 1 to make the number of dimensions match
  - Now, every dimensions is required to have matching size, or one of the dimensions must equal 1 (in which case it will be broadcasted)
  - In the same way, the batch dimension must match, or equal 1 in which case it will be broadcasted
  - The resulting tensor's dimensionality is thus determined as the max of both inputs at every position
- ```mean_dim```, ```std_dim```, ```moment_dim``` can now reduce along multiple dimensions at once, including the batch dimension; this makes the following redundant: ```mean_elems```, ```moment_elems```, ```std_elems```, ```moment_batches```, ```mean_batches```, ```std_batches```; they are kept in the code though and simple link to the new code
  - Speed up +, cmult, cdiv by avoiding Eigen broadcast when inputs only differ in batch dimension

Things I'll leave for a future commit:
  - Note that currently, CwiseQuotient only supports broadcasting the second expression; I would leave the other way for someone else to implement in a future PR if needed
  - Remove ```sum_elems``` and ```sum_batches```, change```sum_dim``` to also reduce along multiple dimensions
  - similar broadcasting behavior for pow() operator

Speed improvements:
- ```std_dim```
- ```weight_norm```
- ```cdiv``` (backward pass)
Previously operations performed before a broadcast were recomputed for every value duplicated via the broadcast; instead, we now precompute them using temporary memory. This can improve speed several orders of magnitude in some cases, depending on how much is broadcasted.

TODO:
- [X] make broadcasting more flexible
- [X] deprecate redundant operators
- [X] have the plus operator use CwiseSum; remove ScalarAdd, ScalarMultiply, ScalarQuotient nodes from the code, as these are now redundant.
- [x] fix errors


pmichel31415:
Thanks this is very cool. I think it would make more sense to have the `+` operation with automatic broadcasting rather than a separate `cadd` op.

Also removing all the *_elems operation breaks compatibility, I think it would be better to keep them and make them give an error/warning (and remove them from the docs).

msperber:
Yes, I'll do both of these; As of what's implemented so far, cadd can't handle things like tensor + scalar, but I'll update this PR once we're ready to do that.

neubig:
Travis seems to be passing. Is this done? If so, I'll be ready to merge when merge conflicts are resolved.

pmichel31415:
Is there anything preventing this from being merged?

neubig:
Ahh, I was planning on making sure there were no speed regressions when using `a + b` over identically sized matrices (potentially with different batch dimensions, and particularly on CPU). I don't have time today, but if you could run a quick test that'd be appreciated.

pmichel31415:
Hmm not sure I'll have time to do it this week.

neubig:
Ok, I came up with a simple profiling script (we might want to turn this into a "performance" unit test, like the ones in the `tests-io.cc`):

    import dynet as dy
    import time
    import numpy as np
    
    model = dy.ParameterCollection()
    
    a = model.add_parameters((1024,1024))
    b = np.random.rand(1024,1024,32)
    
    start = time.time()
    
    ad = dy.parameter(a)
    bd = dy.inputTensor(b, batched=True)
    
    sums = []
    for x in range(10):
      sums.append(dy.sum_elems(ad + bd))
    
    loss_exp = dy.sum_batches(dy.esum(sums))
    
    start = time.time()
    loss = loss_exp.value()
    print("forward {}".format(time.time()-start))
    
    start = time.time()
    loss_exp.backward()
    print("backward {}".format(time.time()-start))

Here are the results before this PR on CPU:

    forward 1.3658833503723145
    backward 10.38591480255127

Interesting, there was already a performance problem in backward *before* this PR. And here are the stats after:

    forward 7.674404144287109
    backward 10.910429239273071

So it looks like there was a significant performance regression. I'm guessing that this is because we changed to using Eigen broadcasting instead of the more optimized implementation that we had before. It's OK if it's slow when we're doing broadcasting (for now), but for identically sized matrices I'd like to prevent the performance regression (and also fix the problem in backward ideally). Mind taking a look?

msperber:
This should probably fix it, I just copied over the pre-PR code for the special case of unequal batch size, equal other dimensions, and the forward pass sped up from 10.5 to 2 seconds on my CPU.

neubig:
Hmm, tests seem to be failing. Is there an obvious reason why?

msperber:
Had forgotten to push some files, but seems to be passing tests now.

neubig:
Looks good, I added a few changes that improve speed of backward as well, so pending these passing tests I'll merge.

neubig:
Hmm, manual install seems to be uniformly failing... I'm not sure whether this is the fault of this PR or something else.

pmichel31415:
I think it's caused by this PR, maybe even the last commit.

The tests are failing, here's the errors I get when I try to run test-nodes:

```
/home/pmichel31415/dev/dynet/tests/test-nodes.cc(136): error in "add_gradient": check check_grad(mod, z, 0) failed
/home/pmichel31415/dev/dynet/tests/test-nodes.cc(146): error in "cadd_gradient": check check_grad(mod, z, 0) failed
/home/pmichel31415/dev/dynet/tests/test-nodes.cc(156): error in "cadd_scalar_gradient": check check_grad(mod, z, 0) failed
/home/pmichel31415/dev/dynet/tests/test-nodes.cc(169): error in "cadd_broadcast2_gradient": check check_grad(mod, z, 0) failed
test-nodes: /home/pmichel31415/libs/eigen/unsupported/Eigen/CXX11/src/Tensor/TensorEvaluator.h:380: Eigen::TensorEvaluator<const Eigen::TensorCwiseBinaryOp<BinaryOp, LeftArgType, RightArgType>, Device>::TensorEvaluator(const XprType&, const Device&) [with BinaryOp = Eigen::internal::scalar_sum_op<float, float>; LeftArgType = const Eigen::TensorMap<Eigen::Tensor<float, 1> >; RightArgType = const Eigen::TensorMap<Eigen::Tensor<float, 1> >; Device = Eigen::DefaultDevice; Eigen::TensorEvaluator<const Eigen::TensorCwiseBinaryOp<BinaryOp, LeftArgType, RightArgType>, Device>::XprType = Eigen::TensorCwiseBinaryOp<Eigen::internal::scalar_sum_op<float, float>, const Eigen::TensorMap<Eigen::Tensor<float, 1> >, const Eigen::TensorMap<Eigen::Tensor<float, 1> > >]: Assertion `dimensions_match(m_leftImpl.dimensions(), m_rightImpl.dimensions())' failed.
unknown location(0): fatal error in "cadd_broadcast2_gradient": signal: SIGABRT (application abort requested)
```

So I think the backward pass is broken for the sum (even when there is no eigen error the gradient check fails)

dynet/expr.cc
```diff
@@ -176,28 +158,30 @@ Expression sum_dim(const Expression& x, unsigned d) { return Expression(x.pg, x.
...
-Expression std_batches(const Expression& x) { return Expression(x.pg, x.pg->add_function<StdBatches>({x.i})); }
 
-Expression mean_dim(const Expression& x, unsigned d) { return Expression(x.pg, x.pg->add_function<MomentDimension>({x.i}, d, 1)); }
-Expression moment_dim(const Expression& x, unsigned d, unsigned r) { return Expression(x.pg, x.pg->add_function<MomentDimension>({x.i}, d, r)); }
-Expression std_dim(const Expression& x, unsigned d) { return Expression(x.pg, x.pg->add_function<StdDimension>({x.i}, d)); }
+Expression mean_dim(const Expression& x, const vector<unsigned>& dims, bool b) { return Expression(x.pg, x.pg->add_function<MomentDimension>({x.i}, dims, 1, b)); }
+Expression moment_dim(const Expression& x, const vector<unsigned>& dims, unsigned r, bool b) { return Expression(x.pg, x.pg->add_function<MomentDimension>({x.i}, dims, r, b)); }
+Expression std_dim(const Expression& x, const vector<unsigned>& dims, bool b) { return Expression(x.pg, x.pg->add_function<StdDimension>({x.i}, dims, b)); }
+
+Expression moment_elems(const Expression& x, unsigned r) { DYNET_RUNTIME_ERR("moment_elems has been deprecated, use moment_dim instead"); }
```
neubig: I don't think we need to deprecate these, do we? I think they can just be an alias to the appropriate call to MomentDimension. These are convenient for a number of applications I think.

neubig:
OK, I've fixed bugs in backwards (introduced by me, sorry...), and also de-deprecated some functions. I think this should basically be ready to merge.

msperber:
Thank you! I've also updated the description accordingly.

PR #777, opened 2017-08-13
Fix incorrect function calls in cudnn-ops.cu

xunzhang:


xunzhang:
cc @zhisbug @Abasyoni I am not sure about this, but the original argument name is kind of confusing to me.

xunzhang:
I will close this then. But the naming issue still confuses me a little.

PR #780, opened 2017-08-14
Fix eigen gcc version

### LINKED ISSUE #59: Segfault on CUDA when using GCC 4.9.3 and optimization -O2
Currently, when using cnn on CUDA, compiling with GCC 4.9.3, segfaults occur when using the default optimization setting of -O2. This is not a problem on GCC 4.8.*, but may be a problem on other versions of GCC.

We don't currently know the cause, but a patch fixing the bug would be welcome. If you want a quick fix, find the line including `APPEND CUDA_NVCC_FLAGS` in `cnn/CMakeLists.txt` and change `-O2` to `-O1`.

QiJune:
Fix #59 
I am a committer of paddle. When I work on refactoring paddle code using unsupported tensor module of Eigen3, I meet the same problem. GCC5.4 with -O1 is good, but -O2 will cause segment fault.

I found that Tensorflow also used tensor module of eigen3, but have no such error. Tensorflow has implemented [EigenCudaStreamDevice](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L77). 
It's interesting that in the constructor of [EigenCudaStreamDevice](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L79), no cuda stream will passed, but in [Reinitialize](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L84), a cuda stream will be passed.

So, I make a test, use default stream(means initialize in constructor) or set a created stream later(just implement a class EigenCudaStreamDevice like tensorflow).

The gcc version is 5.4.

In  branch [just_for_debug](https://github.com/QiJune/dynet/tree/just_for_debug), example/train_xor will cause segment fault.
In branch [fix_eigen_gcc_version](https://github.com/QiJune/dynet/tree/fix_eigen_gcc_version), example/train_xor works good.

In paddle, the solution also works good. So, maybe dynet can have a test to verify my solution.

### DIFF
--- a/dynet/CMakeLists.txt
+++ b/dynet/CMakeLists.txt
@@ -225,17 +225,7 @@ if(WITH_CUDA_BACKEND)
   if(CUDNN_FOUND)
     list(APPEND CUDA_NVCC_FLAGS "-DHAVE_CUDNN")
   endif()
-  if(CMAKE_COMPILER_IS_GNUCXX)
-    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.9 OR CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 4.9)
-      # gcc 4.9 or later versions raise SEGV due to the optimization problem.
-      # Use -O1 instead for now.
-      list(APPEND CUDA_NVCC_FLAGS "-O1")
-    else()
-      list(APPEND CUDA_NVCC_FLAGS "-O2")
-    endif()
-  else()
-    list(APPEND CUDA_NVCC_FLAGS "-O2")
-  endif()
+  list(APPEND CUDA_NVCC_FLAGS "-O2")
   if(MSVC)
     # If MSVC, we need the boost flag because nvcc doesn't properly parse part of the boost template definitions
     list(APPEND CUDA_NVCC_FLAGS "-DBOOST_NO_CXX11_ALLOCATOR")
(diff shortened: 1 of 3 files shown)


QiJune:
There is a bug in unspported tensor module, [m_devicePropInitialized](https://github.com/RLovelett/eigen/blob/master/unsupported/Eigen/CXX11/src/Tensor/TensorDeviceCuda.h#L41) and are [m_deviceProperties](https://github.com/RLovelett/eigen/blob/master/unsupported/Eigen/CXX11/src/Tensor/TensorDeviceCuda.h#L40) defined as a static variable in a header file!!
If the TensorDeviceCuda.h header file is included by several .cc file, the value of m_devicePropInitialized will be different. Each .cc file will have its own value. The first file value maybe is true(device properties have been inited), but the second file value can still be false. 
So, it will cause segment fault, when get value from m_deviceProperties[0](m_deviceProperties is actually nullptr!)

neubig:
@QiJune Thanks for pointing this out! This sounds like a bug in Eigen upstream. Have you reported it to them?

QiJune:
@neubig I have not reported to Eigen team. The tensor module is unsupported. I will have a try.
By the way, the tensor module is mainly developed by google, and I guess that Tensorflow has implemented [EigenCudaStreamDevice](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L77) and [Reinitialize](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L84) to set cuda stream. Maybe this is the appropriate way to use Eigen.

neubig:
Thanks again for this! I confirmed that this is working.

PR #787, opened 2017-08-16
some fixes to dropout

### LINKED ISSUE #739: using CompactVanillaLSTMBuilder with dropout produces error

msperber:
- fixed a minor bug (dropout_rate_h > 0.f) should be => (dropout_rate > 0.f || dropout_rate_h > 0.f), thanks @jcyk  @FilippoC 
- made pointer management a bit more foolproof, may or may not fix the encountered issue #739


neubig:
Thanks!

PR #791, opened 2017-08-17
Fix broken Scala serialization tests.

adampauls:
These tests are failing in v2.0 because the ModelSaver wasn't being cleaned up after saving. Also removed a (probably errant?) println.


neubig:
Thanks!

PR #792, opened 2017-08-17
fix scala bindings to account for `Device`

joelgrus:
I just plugged in `default_device` everywhere. It builds and works (in the sense of passing all the Scala unit tests) on both CPU and GPU, but I don't know if my solution is the right one.


neubig:
Thanks! This is better than not breaking, so I'll merge it now. Eventually it might be nice to add multi-device support, but maybe this can wait until we document it better and have a good example in the Python bindings.

Issue #794, opened 2017-08-18
CoupledLSTMBuilder Not Accessible from Python


neubig:
It seems that the CoupledLSTMBuilder cannot be accessed from the Python bindings since v2. This should be re-added.

pmichel31415:
Fixed by #799

Issue #793, opened 2017-08-18
LSTMBuilder Python doc is outdated


neubig:
This needs to be updated to reflect the fact that LSTMBuilder is now an alias for VanillaLSTMBuilder:

http://dynet.readthedocs.io/en/latest/python_ref.html#dynet.LSTMBuilder

pmichel31415:
Fixed by #799

PR #799, opened 2017-08-18
Attempt to fix python LSTMs

pmichel31415:
Addresses #793 and #794 : Adds coupledLSTM to python bindings and makes LSTMBuilder an alias of VanillaLSTMBuilder.

Also fixes the docs.

Untested


neubig:
Thanks! I confirmed that this was working as expected.

PR #805, opened 2017-08-21
Fix #803: make regex in setup.py Python 2-compatible

### LINKED ISSUE #803: Error of re "unmatched group"
I encountered an python error at the final step of building the latest version of dynet.

Here is the cmake command:

     CC=/usr/bin/gcc-4.9 CXX=/usr/bin/g++-4.9 cmake .. -DEIGEN3_INCLUDE_DIR=~/libs/eigen-170714 -DBACKEND=cuda -DCUDNN_ROOT=~/cudnn/ -DMKL=TRUE -DMKL_ROOT=/opt/intel/mkl/ -DPYTHON=`which python`

And the error is like this:

      File "/home/z/tmp/dynet/setup.py", line 113, in <module>
        append_cmake_lib_list(GPULIBRARIES, ENV.get("CUDA_CUBLAS_FILES"))
      File "/home/z/tmp/dynet/setup.py", line 34, in append_cmake_lib_list
        l.extend(map(strip_lib, var.split(";")))
      File "/home/z/tmp/dynet/setup.py", line 39, in strip_lib
        return re.sub(r"^(?:(?:lib)?(.*)\.(?:so|a|dylib)|(.*)\.lib)$", r"\1\2", filename)
      File "/usr/lib/python2.7/re.py", line 151, in sub
        return _compile(pattern, flags).sub(repl, string, count)
      File "/usr/lib/python2.7/re.py", line 278, in filter
        return sre_parse.expand_template(template, match)
      File "/usr/lib/python2.7/sre_parse.py", line 799, in expand_template
        raise error, "unmatched group"
    sre_constants.error: unmatched group

thx ...

danielhers:


neubig:
Thanks!

PR #809, opened 2017-08-22
Made handling of pypandoc errors more lenient

neubig:
It seems that pypandoc commands in `setup.py` are failing on Python 2.7. This is a temporary fix to prevent things from dying.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -296,7 +296,7 @@ class build_ext(_build_ext):
 try:
     import pypandoc
     long_description = pypandoc.convert("README.md", "rst")
-except (IOError, ImportError):
+except:
     long_description = ""
 
 setup(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -296,7 +306,7 @@ def run(self):
 try:
     import pypandoc
     long_description = pypandoc.convert("README.md", "rst")
-except (IOError, ImportError):
+except:
```
danielhers: Just adding `RuntimeError` should suffice.

neubig: Thanks, I agree that this would suffice in this case, but really regardless of the reason why this fails we wouldn't like it to be a fatal error that prevents install, as we have a reasonable fallback option (don't add a long description). So I think I'll leave it as-is.

PR #812, opened 2017-08-22
draw LSTM dropout masks per sentence, not per minibatch

msperber:
Previously, the dropout mask was the same for every sentence in the minibatch, which is probably not correct (the variational dropout paper doesn't mention minibatches, and the reference implementation seems to draw noise for each sentence: https://github.com/yaringal/BayesianRNN/blob/master/LM_code/main_new_dropout_SOTA_v3.lua#L166 )

### DIFF
--- a/dynet/lstm.cc
+++ b/dynet/lstm.cc
@@ -16,7 +16,7 @@ enum { X2I, H2I, C2I, BI, X2O, H2O, C2O, BO, X2C, H2C, BC };
 CoupledLSTMBuilder::CoupledLSTMBuilder(unsigned layers,
                          unsigned input_dim,
                          unsigned hidden_dim,
-                         ParameterCollection& model) : layers(layers), input_dim(input_dim), hid(hidden_dim) {
+                         ParameterCollection& model) : layers(layers), input_dim(input_dim), hid(hidden_dim), dropout_masks_valid(false) {
   unsigned layer_input_dim = input_dim;
   local_model = model.add_subcollection("lstm-builder");
   for (unsigned i = 0; i < layers; ++i) {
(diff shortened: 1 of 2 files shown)


neubig:
Thanks! And nice catch on this.

jcyk:
I don't think this is a **bug**. Because the users could choose to dropout in either way (per sentence or per batch) through the `set_dropout_masks` function in previous design.

The document already explain it well, check [here](https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L4417)
For `set_dropout_mask`
"If this function is not called on batched input, the same mask will be applied across all batch elements. Use this to apply different masks to each batch element"

So it was actually more flexible before. Now, we can only do it 'per sentence'. However, I think there is also an unnatural constrain in previous design:  "You need to call this __AFTER__ calling `initial_state`", and should be improved.

Yet, I think this PR is okay, because "per sentence" may be generally better than "per batch"? if so, I agree it is annoying to `set_dropout_masks` every time.

At least the document should be updated correspondingly if we want this PR.

msperber:
We _could_ change it to sample per sentence by default, optionally per minibatch (before it was the opposite, and probably few people were even aware of this). The question is why anybody would want to sample per minibatch? In general we want minibatched code to behave identically to non-minibatched code wherever possible, and in this case it would likely make training much more stable especially for larger batch sizes. I can really only think of advantages of per-sentence masks, and no disadvantages (other than marginally increasing memory requirements).

jcyk:
I also believe "per sentence" is generally better than "per batch". 

Just want to clarify that the previous version supported both, while default one may be "unnecessary". 

At least document should be updated then.

PR #819, opened 2017-08-25
Add device consistency check in definition time

xunzhang:
You can only review the first and the last commit. The second commit is the formating.


neubig:
Thanks!

PR #821, opened 2017-08-25
Fix setup.py when using relative paths

neubig:
This PR makes manual install more robust to relative paths.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,6 @@ if (EIGEN3_INCLUDE_DIR is not None and
     not os.path.isdir(EIGEN3_INCLUDE_DIR) and
     os.path.isdir(os.path.join(os.pardir, EIGEN3_INCLUDE_DIR))):
     EIGEN3_INCLUDE_DIR = os.path.join(os.pardir, EIGEN3_INCLUDE_DIR)
-EIGEN3_INCLUDE_DIR = os.path.abspath(EIGEN3_INCLUDE_DIR)    
     
 # Remove the "-Wstrict-prototypes" compiler option, which isn't valid for C++.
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -87,15 +87,24 @@ def get_env(build_dir):
...
 CC_PATH = ENV.get("CC", find_executable("gcc"))
 CXX_PATH = ENV.get("CXX", find_executable("g++"))
 INSTALL_PREFIX = os.path.join(get_python_lib(), os.pardir, os.pardir, os.pardir)
 PYTHON = sys.executable
 
+# Try to find Eigen
+EIGEN3_INCLUDE_DIR = ENV.get("EIGEN3_INCLUDE_DIR")  # directory where eigen is saved
+# Default to under the build directory if specified
+if EIGEN3_INCLUDE_DIR is None:
+    EIGEN3_INCLUDE_DIR = os.path.join(BUILD_DIR, "eigen")
```
danielhers: Note that `BUILD_DIR` here might not be what's given in `--build_dir` yet, since that's handled only in `run()`.

setup.py
```diff
@@ -174,8 +183,6 @@ def initialize_options(self):
     def run(self):
         global BUILD_DIR, BUILT_EXTENSIONS, EIGEN3_INCLUDE_DIR
         BUILD_DIR = os.path.abspath(self.build_dir)
-        if EIGEN3_INCLUDE_DIR is None:
```
danielhers: This breaks the `pip` install because the `hg` below clones it into the current directory, which is the updated `BUILD_DIR` and not the original one (e.g., `.../py-3.5-x64/build` and not `../build`).

PR #824, opened 2017-08-25
Adapt scala binding to #781

shuheik:
Scala binding can't be compiled with CUDA on the latest master (af30c4b).
This change makes it compile again.


neubig:
Thanks!

PR #827, opened 2017-08-25
Give python build targets more descriptive names

mattr1:
This is really minor, but I wanted the python build targets to have more descriptive names. "target" has become "pydynet", and "copy" has become "pydynet_precopy". I don't care what the names are, if anyone has a better suggestion. I also considered "dynet_python" and "dynet_py" (and correspondingly "dynet_python_precopy" and "dynet_py_precopy"), but settled on pydynet for brevity.


neubig:
Yeah, looks fine. Thanks!

PR #831, opened 2017-08-27
Fixes #830 and allows skip-build on Windows

### LINKED ISSUE #830: pip install after build gives error in TypeError in setup.py:180 on Windows
On trying to pip install dynet after building the solution, I get the following `TypeError message` which causes the install to fail. 

```
G:\Code\C++\dynet\build\python>python ..\..\setup.py EIGEN3_INCLUDE_DIR="G:\Code\C++\eigen" build --build-dir="G:\Code\C++\dynet\build" --skip-build install
running build
INFO:root:CMAKE_PATH=F:\Program Files\CMake\bin\cmake.exe
  File "..\..\setup.py", line 337, in <module>
    py_modules=["dynet", "dynet_viz", "dynet_config"],
  File "F:\ProgramData\Anaconda3\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "F:\ProgramData\Anaconda3\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "F:\ProgramData\Anaconda3\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "..\..\setup.py", line 180, in run
    log.info("MAKE_PATH=" + MAKE_PATH)
TypeError: must be str, not NoneType
```

A possible fix to all of these messages is to use the python string formatting function to create the log messages. I will send a pull request on this.

napsternxg:
This commit takes care of the following errors when passing --skip-build
flag on Windows.

* It uses proper python string formatting options to create log messages
* It skips building extensions when --skip-build flag is passed.

Fixes #830

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -173,23 +173,26 @@ class build(_build):
 
     def run(self):
         global BUILD_DIR, BUILT_EXTENSIONS, EIGEN3_INCLUDE_DIR
+        if self.skip_build:
+            print("Skipping Build.")        
         BUILD_DIR = os.path.abspath(self.build_dir)
         if EIGEN3_INCLUDE_DIR is None:
             EIGEN3_INCLUDE_DIR = os.path.join(BUILD_DIR, "eigen")
-        log.info("CMAKE_PATH=" + CMAKE_PATH)
-        log.info("MAKE_PATH=" + MAKE_PATH)
+        log.info("CMAKE_PATH={}".format(CMAKE_PATH))
+        log.info("MAKE_PATH={}".format(MAKE_PATH))
         log.info("MAKE_FLAGS=" + " ".join(MAKE_FLAGS))
         if HG_PATH is not None:
-            log.info("HG_PATH=" + HG_PATH)
-        log.info("EIGEN3_INCLUDE_DIR=" + EIGEN3_INCLUDE_DIR)
-        log.info("CC_PATH=" + CC_PATH)
-        log.info("CXX_PATH=" + CXX_PATH)
-        log.info("SCRIPT_DIR=" + SCRIPT_DIR)
-        log.info("BUILD_DIR=" + BUILD_DIR)
-        log.info("INSTALL_PREFIX=" + INSTALL_PREFIX)
-        log.info("PYTHON=" + PYTHON)
-        run_process([CMAKE_PATH, "--version"])
-        run_process([CXX_PATH, "--version"])
+            log.info("HG_PATH={}".format(HG_PATH))
+        log.info("EIGEN3_INCLUDE_DIR={}".format(EIGEN3_INCLUDE_DIR))
+        log.info("CC_PATH={}".format(CC_PATH))
+        log.info("CXX_PATH={}".format(CXX_PATH))
+        log.info("SCRIPT_DIR={}".format(SCRIPT_DIR))
+        log.info("BUILD_DIR={}".format(BUILD_DIR))
+        log.info("INSTALL_PREFIX={}".format(INSTALL_PREFIX))
+        log.info("PYTHON={}".format(PYTHON))
+        if not self.skip_build:
+            run_process([CMAKE_PATH, "--version"])
+            run_process([CXX_PATH, "--version"])
 
         # This will generally be called by the pip install
         if not self.skip_build:
@@ -204,12 +207,12 @@ class build(_build):
 
             # Prepare folders
             if not os.path.isdir(BUILD_DIR):
-                log.info("Creating build directory " + BUILD_DIR)
+                log.info("Creating build directory {}".format(BUILD_DIR))
                 os.makedirs(BUILD_DIR)
 
             os.chdir(BUILD_DIR)
             if os.path.isdir(EIGEN3_INCLUDE_DIR):
-                log.info("Found eigen in " + EIGEN3_INCLUDE_DIR)
+                log.info("Found eigen in {}".format(EIGEN3_INCLUDE_DIR))
             elif HG_PATH is None:
                 raise DistutilsSetupError("`hg` not found.")
             else:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -272,14 +276,14 @@ def run(self):
...
             LIBRARY_DIRS.append(BUILD_DIR + "/dynet/")
-        log.info("Building Cython extensions...")
-        log.info("INCLUDE_DIRS=" + " ".join(INCLUDE_DIRS))
-        log.info("LIBRARIES=" + " ".join(LIBRARIES))
-        log.info("LIBRARY_DIRS=" + " ".join(LIBRARY_DIRS))
-        log.info("COMPILER_ARGS=" + " ".join(COMPILER_ARGS))
-        log.info("EXTRA_LINK_ARGS=" + " ".join(EXTRA_LINK_ARGS))
-        log.info("RUNTIME_LIB_DIRS=" + " ".join(RUNTIME_LIB_DIRS))
-        _build_ext.run(self)
+            log.info("Building Cython extensions...")
```
danielhers: Why should these be inside the `if`? `BUILT_EXTENSIONS` means "extensions have already been built".

setup.py
```diff
@@ -255,8 +259,8 @@ def run(self):
                 EIGEN3_INCLUDE_DIR = os.path.join(EIGEN3_INCLUDE_DIR, os.pardir)
             if not os.path.isdir(EIGEN3_INCLUDE_DIR):
                 raise RuntimeError("Could not find Eigen in EIGEN3_INCLUDE_DIR={}. If doing manual install, please set the EIGEN3_INCLUDE_DIR variable with the absolute path to Eigen manually. If doing install via pip, please file an issue at the github site.".format(EIGEN3_INCLUDE_DIR))
+            BUILT_EXTENSIONS = False  # already build repo
```
danielhers: The idea in setting `BUILT_EXTENSIONS` to `True` even if `--skip-build` is given is that in that case we assume everything has already been built. This causes the `error: [Errno 2] No such file or directory: 'build'` failure on Travis CI.

setup.py
```diff
@@ -173,23 +173,26 @@ def initialize_options(self):
...
-        run_process([CXX_PATH, "--version"])
+            log.info("HG_PATH={}".format(HG_PATH))
+        log.info("EIGEN3_INCLUDE_DIR={}".format(EIGEN3_INCLUDE_DIR))
+        log.info("CC_PATH={}".format(CC_PATH))
+        log.info("CXX_PATH={}".format(CXX_PATH))
+        log.info("SCRIPT_DIR={}".format(SCRIPT_DIR))
+        log.info("BUILD_DIR={}".format(BUILD_DIR))
+        log.info("INSTALL_PREFIX={}".format(INSTALL_PREFIX))
+        log.info("PYTHON={}".format(PYTHON))
+        if not self.skip_build:
```
danielhers: Should be combined into the next `if`, as it's the same condition.

setup.py
```diff
@@ -173,23 +173,26 @@ def initialize_options(self):
 
     def run(self):
         global BUILD_DIR, BUILT_EXTENSIONS, EIGEN3_INCLUDE_DIR
+        if self.skip_build:
```
danielhers: Can be in the `else` below.

neubig:
P.S. I just want to note that we really appreciate the contribution! We'll just need to fix the cosmetic issues and ensure that it passes integration tests.

(Also @danielhers on an auxiliary note, I feel that we should really refactor `setup.py` to make the control flow more clear. This isn't urgent, but I think it'd make it more maintainable and save time in the long run.)

setup.py
napsternxg: But if I am building everything externally say using VS on Windows, the code shouldn't be running the Build extensions code. Would you add a bit more clarification on this ?

danielhers:
I completely agree. In my efforts to understand the old setup logic I got quite attached to it and to this code, but after this merging work it's still overly complicated for no good reason.

setup.py
napsternxg: @danielhers looking at the travis file it looks like the following line should be edited to fix the issue

https://github.com/clab/dynet/blob/feb01a17560eeafb8c5fdb0a39634457ef62d5fa/.travis.yml#L92

It has
```
      python ../../setup.py build --build-dir=.. --skip-build install --user || travis_terminate 1;

```

If I am skipping the build then it should not be required to pass the build argument to setup.py. 

What do you think? Without using that piece of code, the build fails on Windows as the setup.py tries to compile the files using g++. 

If I keep the `BUILD_EXTENSIONS` line as before then I get the error described in #830

danielhers:
`build-dir` is required even with `--skip-build` because `build_ext` needs to know where the build directory is in order to copy the files from it, and because install needs to find the files too.

napsternxg:
@danielhers I am unable to see my previous comment, but my question was not about the `build-dir`, but about the `build` argument which is passed to `setup.py` when using `--skip-build`. I don't think we should be passing `build` argument along with `--skip-build` as they are contradictory (as per my thinking).

danielhers:
We need to know where the build path was even when doing install without actually building, because we need to copy files from the build directory.

danielhers:
Oh, sorry, now I see what you mean. `build` and `install` are `setup.py` *commands*. I made the `build` command accept the `--build-dir` and `--skip-build` arguments, but the `install` command doesn't know them, so in order to read them and set `BUILD_DIR` we need the `build` command.

napsternxg:
OK great got it. Thanks.

neubig:
@napsternxg @danielhers is this still being worked on? I don't have a good windows environment to use to test this unfortunately...

danielhers:
I think most of the changes here have already been done in other commits, and I don't understand the rationale for the rest. It seems pip install still fails on Windows, though. I don't have a Windows environment either, but with #914 this can be tested (currently AppVeyor is only doing manual install, but pip uninstall will only require uncommenting a few lines in .appveyor).

neubig:
OK, I'm going to close this if most of the other things have already been resolved. If there are any outstanding issues we can re-open a new PR that addresses them.

PR #835, opened 2017-08-28
Adding missing header files to CMakeLists

mattr1:
BTW: I would have just done this as a commit directly rather than a pull request, but I wanted to make sure it passed the travis-ci tests. Is there any way to kick off a travis-ci to check a branch, or check after committing, rather than making a pull-request?

### DIFF
--- a/dynet/CMakeLists.txt
+++ b/dynet/CMakeLists.txt
@@ -100,9 +100,39 @@ set(dynet_library_HDRS
     matrix-multiply.h
     mem.h
     model.h
+    nodes-activations.h
+    nodes-affinetransform.h
+    nodes-arith-const.h
+    nodes-arith-cwise.h
+    nodes-arith-scalar.h
+    nodes-arith-sum.h
+    nodes-arith-unary.h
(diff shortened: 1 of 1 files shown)


pmichel31415:
It used to be automatic, for some reason it got deactivated.

I just reactivated it. Also if you're a contributor on the repo you can trigger builds in travis directly by logging in with your github account.

xunzhang:
According to the current setting, It should trigger a CI run after committing to `addheaders` branch. My guess is that when creating a new branch, it will not trigger, please try making another commit in `addheaders`?

A better way to do this is cloning to local repo then open Travis hook in that personal repo, this will relieve the burden of upstream's pull requests and master commits.

mattr1:
Sorry for the ignorance, but how would I see if travis-ci has run on a given branch (or can I only see that if I log in to travis-ci)? The only place I ever see it is when creating a pull request.

pmichel31415:
Either in the branch tab on the github front-page or in travis directly

mattr1:
Aha, great. Thank you.

Issue #774, opened 2017-08-10
Better Support for "in-place" Operations


neubig:
Many operations such as `reshape`, `no_backprop`, etc. don't need to make a memory copy in the forward pass. However, these nodes still result in doing memory allocation and copying memory (before autobatching they didn't result in a memory copy, but the autobatching functionality required them to move memory, so now everything moves memory).

It would be nice to have a way to note in-place operators and prevent them from using extra memory, etc. The one thing we need to be careful about is that they don't break autobatching.

neubig:
Added by #840 and following PRs.

PR #840, opened 2017-08-29
Some partial supports for inplaced operations.

zzsfornlp:
This PR adds certain partial enhancements for in-place operations (concerning issue [#774](https://github.com/clab/dynet/issues/774)) to save some memories.

About the changes:

0. For usage, call the enhanced expressions with an additional parameter `inplaced` set to `true`, those added parameters are set to `false` by default, thus no conflicts with previous codes. 
1. For inplace operations, the `Node` does not hold its own memory, it borrows the forward-value/backward-gradient memories of its args and the corresponding operations are done in-place. Two kinds of inplace nodes are considered: READ(no re-writing the forward-value) and WRITE(scrambling the original value, but don't need the original value for gradient calculations, for eg., tanh). `Reshape` is an example of the former while `Tanh` is an example of the latter.
2. The borrowing is simply implemented by sharing (float*) pointers in Tensor::v, and if the node is `inplaced`, no new memory will be allocated for them and its arg's value pointer will be directly utilized. The forward process is not changed since Eigen seems to be able to handle the situation of alias, while for the backward, all `+=` are changed to `=`. (This breaks the convention of using `+=` for backward, but is the most convenient way of sharing gradient memories, thus some constrains are placed, the main of which is that the borrowed node could only be used by this borrowing inplaced node.)
3. Currently only support certain unary operations: Reshape, Identity (READ); Rectify, LogisticSigmoid, SoftSign, Tanh, Dropout* (WRITE).

Constrains:
1. Only support certain unary operations as described above.
2. For simplicity, break the convention of using `+=` for backward, thus adding somewhat strict rules for the inplaced nodes and the nodes whose memories are borrowed. Briefly, WRITE-inplaced Nodes cannot borrow memories of the nodes whose output have been used (read) by other nodes or another WRITE-inplaced Node (nested WRITE-nodes might bring certain problems); READ-inplaced and plain Nodes cannot rely on nodes whose memories have been borrowed by WRITE-inplaced Nodes. See function `ComputationGraph::set_inplace_for_new_node`.
3. Forbid `revert`/`get_value` when we need to restore some scrambled values.
4. For simplicity, disable the supports of auto-batch for all inplaced nodes, since we cannot guarantee they are in contagious memories.
5. Currently no supports for `NoBackprop` and `FlipGradient`, whose gradients need special manipulations and thus are not that easy to be correctly shared.

Example (in python):

    import dynet as dy
    import numpy as np
    import random
    N = 100
    Ns = 10
    m = dy.Model()
    ini = [random.random() for _ in range(N)]
    w1 = m.add_parameters((N,), dy.NumpyInitializer(np.array(ini)))
    w2 = m.add_parameters((N,), dy.NumpyInitializer(np.array(ini)))
    tr = dy.SimpleSGDTrainer(m)
    assert all(w1.as_array() == w2.as_array())
    for inplaced in [True, False]:
        wp = dy.parameter(w1 if inplaced else w2)
        x1 = dy.inputVector(ini)
        x2 = dy.reshape(x1, (Ns, Ns), inplaced)
        wp2 = dy.reshape(wp, (Ns, Ns), inplaced)
        y1 = wp2 * x2
        y2 = dy.reshape(y1, (N,), inplaced)
        z1 = y2 + wp
        z2 = dy.tanh(z1, inplaced)
        loss = dy.sum_elems(z2)
        loss.backward()
        tr.update()
    assert all(w1.as_array() == w2.as_array())

Here the two iter should have the same effect with only the differences of calling `reshape` and `tanh` with different `inplaced` values.

I'm not sure whether this implement for inplaced operations is fine, or there are hidden problems with this. From the user's perspective, the only changes are the additional parameters of `inplaced` for certain operations. Since dynet is dynamic calculation graph, and users can always add later operations depending on previous values, it can be hard to automatically optimize with in-place operations. Thus to use inplace, maybe the users could only manually specify that like this?

### DIFF
--- a/dynet/dynet.cc
+++ b/dynet/dynet.cc
@@ -164,8 +164,22 @@ void ComputationGraph::_revert(CGCheckpoint p) {
   default_device->revert(p.device_mem_checkpoint);
   // clear all nodes at position >= p.node_idx
   if ((int)nodes.size() > p.node_idx) {
-    for(int i = p.node_idx; i < (int)nodes.size(); i++)
-      delete nodes[i]; // the deletion of nodes.
+		for(int i = p.node_idx; i < (int)nodes.size(); i++){
+			// revert some info
+			Node* one = nodes[i];
+			for(VariableIndex arg : one->args) {
+				if(one->inplace_state == Node::INPLACE_TYPE::WRITE){
+					if(arg <= p.node_idx){
+						DYNET_RUNTIME_ERR("Invalid revert for the node "<<arg<<" cannot be recovered from certain inplaced operation");
+					}
+					nodes[arg]->write_num -= 1;
+				}
+				else{
+					nodes[arg]->read_num -= 1;
+				}
+			}
+			delete one; // the deletion of nodes.
+		}
     nodes.resize(p.node_idx);
     ee->invalidate(p.node_idx - 1); // clear precomputed forward values
   }
@@ -347,6 +361,40 @@ VariableIndex ComputationGraph::add_const_lookup(LookupParameter p, const std::v
   nodes.back()->device = p.get_storage().device;
   set_dim_for_new_node(new_node_index);
   return new_node_index;
+}
+
+// check for the inplace operations
+void ComputationGraph::set_inplace_for_new_node(const VariableIndex& i) {
+	Node* node = nodes[i];
+	// 1. check current node
+	if(node->inplaced()){
+		// currently, only unary operations support inplacing
+		DYNET_ARG_CHECK(node->args.size() == 1, "Failed input count check for inplaced-operation");
+		Node* x0 = nodes[node->args[0]];
+		// check for validity, currently somewhat strict, only allow multiple READ or one WRITE
+		if(node->inplace_state == Node::INPLACE_TYPE::READ){
+			DYNET_ARG_CHECK(x0->write_num == 0, "Failed input for READ inplaced-operation");
+		}
+		else if(node->inplace_state == Node::INPLACE_TYPE::WRITE){
+			// currently only allow one layer of inplace for WRITE since letting it spreading out might cause certain problems
+			DYNET_ARG_CHECK((x0->write_num + x0->read_num == 0) && (!x0->inplaced()), "Failed input for WRITE inplaced-operation");
+		}
+		else{
+			DYNET_RUNTIME_ERR("Impossible inplace staus");
+		}
+	}
+	// 2. check input arguments (make sure their value is preserved)
+	for(VariableIndex arg : node->args) {
+		Node* x = nodes[arg];
+		DYNET_ARG_CHECK(x->write_num == 0, "Failed input argument (already being inplaced)");
+	}
+	// 3. update info
+	for(VariableIndex arg : node->args) {
+		if(node->inplace_state == Node::INPLACE_TYPE::WRITE)
+			nodes[arg]->write_num += 1;
+		else
+			nodes[arg]->read_num += 1;
+	}
 }
 
 // factory function should call this right after creating a new node object
--- a/dynet/dynet.cc
+++ b/dynet/dynet.cc
@@ -169,7 +169,7 @@ void ComputationGraph::_revert(CGCheckpoint p) {
       Node* one = nodes[i];
       for(VariableIndex arg : one->args) {
         if(one->inplace_state == Node::INPLACE_TYPE::WRITE){
-          if(arg <= p.node_idx){
+          if(arg <= (unsigned)p.node_idx){
             DYNET_RUNTIME_ERR("Invalid revert for the node "<<arg<<" cannot be recovered from certain inplaced operation");
           }
           nodes[arg]->write_num -= 1;
@@ -377,7 +377,9 @@ void ComputationGraph::set_inplace_for_new_node(const VariableIndex& i) {
     }
     else if(node->inplace_state == Node::INPLACE_TYPE::WRITE){
       // currently only allow one layer of inplace for WRITE since letting it spreading out might cause certain problems
-      DYNET_ARG_CHECK((x0->write_num + x0->read_num == 0) && (!x0->inplaced()), "Failed input for WRITE inplaced-operation");
+      DYNET_ARG_CHECK((x0->write_num + x0->read_num == 0), "Failed input (used by others) for WRITE inplaced-operation");
+      DYNET_ARG_CHECK((!x0->inplaced()), "Failed input (already inplaced) for WRITE inplaced-operation");
+      DYNET_ARG_CHECK((x0->rewritable), "Failed input (not rewritable) for WRITE inplaced-operation");
     }
     else{
       DYNET_RUNTIME_ERR("Impossible inplace staus");
--- a/dynet/dynet.cc
+++ b/dynet/dynet.cc
@@ -164,22 +164,22 @@ void ComputationGraph::_revert(CGCheckpoint p) {
   default_device->revert(p.device_mem_checkpoint);
   // clear all nodes at position >= p.node_idx
   if ((int)nodes.size() > p.node_idx) {
-		for(int i = p.node_idx; i < (int)nodes.size(); i++){
-			// revert some info
-			Node* one = nodes[i];
-			for(VariableIndex arg : one->args) {
-				if(one->inplace_state == Node::INPLACE_TYPE::WRITE){
-					if(arg <= p.node_idx){
-						DYNET_RUNTIME_ERR("Invalid revert for the node "<<arg<<" cannot be recovered from certain inplaced operation");
-					}
-					nodes[arg]->write_num -= 1;
-				}
-				else{
-					nodes[arg]->read_num -= 1;
-				}
-			}
-			delete one; // the deletion of nodes.
-		}
+    for(int i = p.node_idx; i < (int)nodes.size(); i++){
+      // revert some info
+      Node* one = nodes[i];
+      for(VariableIndex arg : one->args) {
+        if(one->inplace_state == Node::INPLACE_TYPE::WRITE){
+          if(arg <= p.node_idx){
+            DYNET_RUNTIME_ERR("Invalid revert for the node "<<arg<<" cannot be recovered from certain inplaced operation");
+          }
+          nodes[arg]->write_num -= 1;
+        }
+        else{
+          nodes[arg]->read_num -= 1;
+        }
+      }
+      delete one; // the deletion of nodes.
+    }
     nodes.resize(p.node_idx);
     ee->invalidate(p.node_idx - 1); // clear precomputed forward values
   }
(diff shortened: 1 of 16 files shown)


neubig:
Wow, this is great. Thanks! We will need to test this carefully to make sure it doesn't break various things, such as autobatching (or at least throws errors instead of silently failing), or when you try to use the value of an expression that has been invalidated via an in-place operation, but I'm looking forward to testing this.

zzsfornlp:
Some additional notes:

Currently, only support inplaced operations for certain nodes with the following conditions:
(1) has one and only one input, whose dim() should be the same as the output.
(2) no need for the original input value in the backward gradient calculation, thus is ok to optimize out.
(3) no special gradient manipulations thus backward will not be broken.
Therefore, only selecting certain nodes: Reshape, Identity (READ); Rectify, LogisticSigmoid, SoftSign, Tanh, Dropout* (WRITE).

------------------------------------------------------

Since inplaced operations directly share memories, there indeed might be certain silent failures which will result in wrong calculations without any warnings. Thus, currently the rules for inplaced nodes are relatively strict. (Most of them are checked in `ComputationGraph::set_inplace_for_new_node` by DYNET_ARG_CHECK).

Here are the failing situations checked (default action is throwing runtime_error), wonder if there are anything remiss or anything breaking the procedure, might need some help for this.

-- failing situation 1: values cannot be restored or used for it is already changed by WRITE-inplace nodes.

    # failing situation 1.1: cannot restore value for revert
    import dynet as dy
    x = dy.zeroes(100)
    dy.cg_checkpoint()
    y = dy.tanh(x, True)
    y.value()
    dy.cg_revert()
    
    # failing situation 1.2: cannot restore value for get_value
    import dynet as dy
    x = dy.zeroes(100)
    y = dy.tanh(x, True)
    y.value()
    print(x.value())
    
    # failing situation 1.3: cannot use the node whose value has been WRITE-borrowed
    import dynet as dy
    x = dy.zeroes(100)
    y = dy.tanh(x, True)
    z = x + 1

-- failing situation 2: cannot add inplace nodes due to the current exclusive property of WRITE-inplace. (This is relatively strict for the purpose of a simpler implementation: the principle is READ is shareable while WRITE is exclusive.)

    # failing situation 2.1: *-inplace + WRITE-inplace
    import dynet as dy
    x = dy.zeroes(100)
    y = dy.tanh(x, True)
    y2 = dy.tanh(y, True)
    
    # failing situation 2.2: used node (by plain nodes or READ-inplace nodes) + WRITE-inplace
    import dynet as dy
    x = dy.zeroes(100)
    z = x + 1
    y = dy.tanh(x, True)

(In fact, some WRITE+WRITE operations might be possible to implement, like `tanh+dropout` which is typical for hidden layers, which could be extended but need to carefully specify the rules . However, most of them seems to unfeasible because we need the intermediate values for gradient calculation.)

-- ok situation: multiple READ-inplaces are allowed since they are not changing the original values.

    import dynet as dy
    x = dy.zeroes(100)
    x1 = dy.reshape(x, (10,10), True)
    x2 = dy.reshape(x, (10,10), True)
    x11 = dy.reshape(x1, (100,), True)
    for z in [x,x1,x2,x11]:
        print(z.dim())

(Currently no supports for NoBackprop and FlipGradient, whose gradients need special manipulations and thus are not that easy to be correctly shared.)

------------------------------------------------------

For simplicity, currently make all the inplaced nodes non-auto-batchable (return 0 for autobatch-sig). Since it seems that auto-batch need the inputs in contagious memory, which usually cannot be guaranteed. Thus this simple strategy might be ok? since the purpose of inplace operations is saving memories.

------------------------------------------------------

Inplace operations should only be a way of optimization for memories and will never bring any other changes. Thus two programs with the same codes except the inplace operations should obtain exact the same results. Tests will be needed on real tasks, expecting the same results and models; I'll plan to test this soon.

zzsfornlp:
Sorry for remissing a huge potential risk (silent failing) that breaks the gradient: some nodes need the original forward value to calculate the gradients (like tanh), thus they cannot be rewritten by any inplaced operators (in simple nodes we can specifically restore the values, but this might not be general and efficient). Really need to carefully inspecting this feature about inplacing operations ...

Thus adding a method `set_rewritable` to explicitly declare that the expression(node) gives up its memory. This could also be used for all the nodes in testing mode since we do not need to calculate gradients. Unfortunately, this has to be done by the users. (currently all the nodes set rewritable=false by default, thus `set_rewritable` needs to be called before all inplaced operations, but this could be further relaxed for nodes that are safe.)

Here is a simple python MLP code for testing inplaced-tanh:

    import dynet as dy
    import random, sys
    if sys.argv[1].startswith("in"):
        inplaced = True
        print("inplace", file=sys.stderr)
    else:
        inplaced = False
        print("notinplace", file=sys.stderr)
    Ni, Nh, BS = 5, 5, 1
    m = dy.Model()
    random.seed(12345)
    ini1 = [([random.random() for _1 in range(Ni)], 1) for _ in range(BS)]
    ini0 = [([random.random() for _1 in range(Ni)], 0) for _ in range(BS)]
    ws = { "w1": m.add_parameters((Nh, Ni)), "b1": m.add_parameters((Nh,)), "w2":m.add_parameters((1, Nh)), "b2":m.add_parameters((1,))}
    ps = {}
    for k in ws:
        ps[k] = dy.parameter(ws[k])
    tr = dy.SimpleSGDTrainer(m)
    losses = []
    for _ in range(100):
        for x, y in ini0+ini1:
            h0 = dy.inputVector(x)
            h1 = dy.affine_transform([ps["b1"], ps["w1"], h0])
            h1.set_rewritable(True)
            h1t = dy.tanh(h1, inplaced=inplaced)
            h1d = dy.dropout(h1t, 0.2)
            out = dy.affine_transform([ps["b2"], ps["w2"], h1d])
            losses.append(dy.square(y-out))
        loss = dy.esum(losses)
        loss.backward()
        tr.update()
    for w in sorted(ws.keys()):
        print("%s: %s" % (w, ws[w].as_array()))

After calling it with inplaced(`sys.argv[1].startswith("in")`) or not-inplaced(others), the same weights should be obtained.

redpony:
Sorry for commenting late on this, but I'm **extremely** concerned about
this line of work.

The problem with this is that there is no way to know whether you can
"borrow" the argument memory in a dynamic network since you might decide
later to  access the original value again in the future. You can let the
caller declare "feel free to reuse this!" but this exposes the programmer
to hard to track down bugs, and it makes the toolkit hard to understand and
hard to learn.

I realise memory usage is important, but simplicity is important too l.
Next year, GPUs will have more memory but complex code will still be
complex. DyNet is designed to be simple and fast, not the most resource
efficient. Let TF and the static toolkits handle extreme memory efficiency
(after all, they can do the static analysis that will reveal whether a
value will be needed again in the future).

neubig:
There are certain operations for which we basically get memory and speed savings for free because the operation doesn't change the output (e.g. reshape and flip_gradient, no_backprop). I would like to make sure that we at least handle these properly, especially reshape, as it's one we tend to use frequently.

For the other ones, I agree that this is a tradeoff that we should discuss, but given that the default will be to not do in placing, this is somewhat of a "use at your own risk" feature that only advanced users would be using. So the main tradeoff is whether it would significantly complicate the code internal to DyNet. That we should discuss before incorporating this feature. (Sorry about the slow action on this, but I want to review it properly before incorporating anything, and doing a proper review takes some time)

redpony:
Yes we should definitely do our best to handle things like reshape properly
(I actually thought we already did that). But we should be wary of "use at
your own risk" since it represents a leakage of internal implementation
into user land. This means we have less freedom to innovate on the inside
because we are left supporting a few power users. This is another bad
tradeoff.

neubig:
Fair enough, I'll look at the code carefully during the review and see if there's anything we can do to reconcile.

zzsfornlp:
Thanks for the comments. I see that memory issue is not the most important one and including inplace operations inevitably brings maybe-hard-to-control complexities. And with dynamic nets, we can only rely on the users themselves to decide whether to do the inplace operations. Those indeed makes these inplace operations not that appealing.

I still think using methods to save memories could be helpful, especially there are situations when we might need to save some. In fact, I experiment using ``inplace-tanh'' in the attention part of att-seq2seq, and it seems to be able to save at least up to 500MB memories.

Although using inplace operations is the choices for users, but the system can also check whether the operations are legal (whether they are using rewritten/staled values) (Currently this is done when building the graph). Thus the risks should be controlled and checked by the system (throwing error when illegal things happen), currently I'm doing this in `set_inplace_for_new_node`.

Yes, I agree that complexing the internal codes might not be a good idea, and it is only until recently (sorry about this) that I find this bug, which could be really dangerous and break down the gradient calculating (silently). And maybe this really opens too much to the users?  But I cannot think of other ways to do that without slightly changing the internals, I don't know whether this is done appropriately.  Thanks for the review and help.

zzsfornlp:
Ah..., maybe constrcuting some special combining nodes could be another way without complexing and potential bugs? In fact what we need is some commonly used combinations, like `affine+tanh+dropout` for the hidden layers, and this is like what lstm-nodes do. But this might not be that general, for example, for the hidden layers in attention, we might need another one with `colwise_add+tanh`. Will this be a better method to deal with memory issues?

neubig:
I had another chance to look at this. I think a good strategy would be to first create a pull request that implements this only for totally safe operations such as reshape, nobackprop, and flip_gradient. Then we can think about what to do next. I see two paths, the first easier and the second more interesting:

1. Allow in-place operations. The pre-condition for doing this is that it doesn't add significant code complexity (over the implementation for reshape, nobackprop, and flip-gradient), and is not able to fail silently.
2. Allow execution-time operation fusing. This would mean that we could change ExecutionEngine a bit so it would detect which operations could be fused together to save memory and execution speed. I don't think that we should force the user to do this, but it'd be good if the backend could discover these opportunities for optimization automatically. One difficulty here is that we'd have to be sure that the intermediate nodes in the operations would not be used later when we expand the graph further. We could be sure of this by adding some capacity for reference counting, and only fuse nodes if there are no references to the intermediate nodes. Needless to say, this is a big job, but I think it'd very much in the spirit of what DyNet is trying to do (not bother the user with the details of optimizing for efficiency), so it might be an interesting project.

zzsfornlp:
Thanks for the advices, maybe indeed a better stratedgy than this PR is needed. About the two paths:

1. Reshape might be easily handled, but for `nobackprop `& `flip-gradient`, I think even more complexities will be included since they are manipulating the gradients, which means that maybe they should keep their own gradient memory. So, these nodes only share other's memory of forward values? Otherwise we may need to add special checking only to detect the existence of these two kinds of special nodes when calculating gradients.

In fact, for the first path, I think maybe the method in this PR is already the most convenient way to do this. Currently, the added complexity is only the checking for invalid inplace-operations and invalid usage of values which have been re-written. If all the conditions are covered in these checkings, there will be no way to fail silently. (And also for future enhancements, we should remember that memory sharing might exist). The biggest problem is the one about interface as discussed before, the inplace opration is assigned by the users.

2. This is an interesting path, and more powerful because maybe more optimizations could be included. But indeed there should be methods to tell the system that the nodes are not needed for further expansion and could be fused. One of the way is depending on the users to add some tags, but this is again violating the spirit as the current PR does. The other implicit way to automatically do this might need to rely on specific languages features like the construction/destruction of C++ objects, to ensure that the users cannot "hold" any to-be optimized nodes. But this still depends on the users' destructing them, and for python which deals with gc without the knowing of users this might be harder.

Indeed, the problem is that the system has no ideas of which part of the computation-graph could be safely optimized. Thus I think to perform these kinds of optimizations, the users should at least provide some information, maybe?

----------

Anyway, thanks for the discussions and indeed I have not been thinking from the perspective of users.
Therefore, should I close this PR? For read-only operations, I can only think of `Reshape` for full memory optimization, for `nobackprop `& `flip-gradient`, will introducing half-sharing (only forward values) be a good idea?

neubig:
First, I'd like to note that we really appreciate the commit! This is just a tough design decision, so we need to figure out how to do it right.

I think there are a couple of options. One would be to only implement memory savings for the forward pass first (where we can do this for `reshape` `nobackprop` and `flip_gradient`). Then we can think about the backward pass, which is more complicated. Please note that any changes that you make will have to be consistent with BatchedExecutionEngine, in addition to SimpleExecutionEngine.

I think it'd be a good idea to close the PR for now, but keep the code that you implemented. Then, we can re-introduce the changes a little bit at a time as separate pull requests.

zzsfornlp:
ok, I will try to deal with the read-only operations ...

PR #846, opened 2017-09-01
Autobatch device

neubig:
This is a fix by @mattr1 regarding a fix for a bug that popped up in the introduction of multi-device support. I'm creating a PR so it will be easy for me to merge after I test.

### DIFF
--- a/dynet/dynet.h
+++ b/dynet/dynet.h
@@ -488,6 +488,7 @@ struct ComputationGraph {
   // flag of checking Inf/NaN of each layer. Only performing checking when
   // immediate_compute is also set to true.
   bool check_validity;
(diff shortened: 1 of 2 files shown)


neubig:
Seems to be working in my environment. Thanks @mattr1 !

xunzhang:
I noticed this commit has been merged using `Squash and Merge` maybe, and the Author info of https://github.com/clab/dynet/commit/37f61e6c21dad11297d30a34066931a0a792ca0c removed mattr1. A suggestion is using `Rebase and Merge` instead when you create a pull request with other's commits, or manually do squash in the command line(this way is better IMO since you can keep the commit info the same format as `Squash and Merge` adding pull request's link in the message). Just FYI @neubig

PR #852, opened 2017-09-03
Remove deprecated Trainer::update_epoch

tetsuok:
This PR removes the deprecated method call used in the examples.
The method was deprecated in #695.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/852)

### DIFF
--- a/examples/cpp/embed-cl/train_embed-cl.cc
+++ b/examples/cpp/embed-cl/train_embed-cl.cc
@@ -145,7 +145,6 @@ int main(int argc, char** argv) {
   unsigned si = training.size();
   vector<unsigned> order(training.size());
   for (unsigned i = 0; i < order.size(); ++i) order[i] = i;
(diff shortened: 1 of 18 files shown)


examples/cpp/embed-cl/train_embed-cl.cc
```diff
@@ -156,7 +156,7 @@ int main(int argc, char** argv) {
     for (unsigned i = 0; i < report_every_i; ++i) {
       if (si == training.size()) {
         si = 0;
-        if (first) { first = false; } else { sgd->update_epoch(); }
```
neubig: Thanks, this is great. We can also probably remove the "first" variable entirely, right?

tetsuok: Done. Please take a look.

neubig:
Thanks!

PR #853, opened 2017-09-03
Use std::unique_ptr.

tetsuok:
This PR takes advantage of std::unique_ptr added in C++11.  Given that
dynet uses C++11, we can take advantage of the feature for RAII if
applicable.

For the code of dynet library, it would be helpful to clarify the
ownership of objects. In particular, it is useful for debugging or avoiding memory
related bugs such as memory leaks, use-after-free.

For the code in examples, we can remove the code to manually delete
objects allocated by new, using std::unique_ptr.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/853)

### DIFF
--- a/examples/cpp/segrnn-sup/train_segrnn-sup.cc
+++ b/examples/cpp/segrnn-sup/train_segrnn-sup.cc
@@ -1074,7 +1074,7 @@ int main(int argc, char** argv) {
 
     ParameterCollection model;
     // auto sgd = new SimpleSGDTrainer(model);
-    std::uniqu_ptr<Trainer> sgd(new AdamTrainer(model, 0.0005, 0.01, 0.9999, 1e-8));
+    std::unique_ptr<Trainer> sgd(new AdamTrainer(model, 0.0005, 0.01, 0.9999, 1e-8));
     int max_seg_len = DATA_MAX_SEG_LEN + 1;
     if(vm.count("train_max_seg_len")){
       max_seg_len = vm["train_max_seg_len"].as<int>();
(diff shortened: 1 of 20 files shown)


examples/cpp/segrnn-sup/train_segrnn-sup.cc
```diff
@@ -1074,7 +1074,7 @@ int main(int argc, char** argv) {
 
     ParameterCollection model;
     // auto sgd = new SimpleSGDTrainer(model);
-    auto sgd = new AdamTrainer(model, 0.0005, 0.01, 0.9999, 1e-8);
+    std::uniqu_ptr<Trainer> sgd(new AdamTrainer(model, 0.0005, 0.01, 0.9999, 1e-8));
```
neubig: Typo here.

tetsuok:
> I'm not so familiar with unique_pointer, but are there any implications for speed/memory efficiency compared to plain pointers?

Assuming we're talking about Release builds.
My understanding is that there is no noticeable overhead of memory and speed unlike `std::shared_ptr` as long as we don't pass customized deleters to std::unique_ptr's constructor.
According to Scott Meyers's Effective Modern C++:

>"It’s reasonable to assume that, by default, std::unique_ptrs are the same size as raw pointers, and for most operations (including dereferencing), they execute exactly the same instructions. This means you can use them even in situations where memory and cycles are tight. If a raw pointer is small enough and fast enough for you, a std::unique_ptr almost certainly is, too."

BTW, if you have used boost's scoped_ptr before, `std::unique_ptr` is standardized version of it.
MeCab and Moses essentially use very similar smart pointer types.

tetsuok:
Fixed typo.

neubig:
Great, thanks!

PR #857, opened 2017-09-04
Reduce global non-POD objects by adding DeviceManager class

tetsuok:
This attempts to remove global non-POD objects (devices, devices_map) by
adding DeviceManager class. Using global non-POD objects can trigger nasty bugs.
In general, it is very hard to reproduce and debug. I think it is a good practice to reduce the use of non-POD objects in global scope.

The idea is simple. 
- Move devices and devices_map move to private members of DeviceManager class. The class provides the access to the variables. 
- get_device_manager is used to get the unique manager object
- Initialization of the manager object is done when get_device_manager is called for the first time (this initialization is thread safe since C++11)

I'm not sure the destructor of Device class is declared as protected instead of public.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/857)


neubig:
Thanks! I don't think there's any particular reason why the destructor is protected instead of public.

PR #862, opened 2017-09-05
Improve installation process with relative Eigen path

xunzhang:
Current `setup.py` reads variables from `CMakeCache.txt`, so we should replace the variables in the cache file with absolute paths. Otherwise, the paths might be problematic.

For example, if we use relative Eigen path like `cmake .. -DEIGEN3_INCLUDE_DIR=../../../../eigen`, we can only do `cd python; python ../../setup.py install --user` which is not good. This pull request optimises this: users can do `python setup.py` anywhere.


xunzhang:
cc @danielhers @neubig

PR #864, opened 2017-09-06
examples/cpp: Use appropriate naming based on the type of training algorithms

tetsuok:
This PR fixes the names of the variables used in C++ examples.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/864)


neubig:
Hmm, I'd actually prefer something like "trainer", but I agree that this is better than before.

xunzhang:
I agree, let's keep it as it is now since there are lots of places using `sgd` name. Maybe we can do it when refactoring the example folder structure.

PR #866, opened 2017-09-06
Use consistently 'trainer' as the variable name of Trainer objects

tetsuok:
As per the discussion at #864, it might be better to use `trainer` consistently rather than individual name based on the type of training algorithms such as `sgd`, `adam` or `adagrad`.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/866)


xunzhang:
Also update [here](https://github.com/clab/dynet/blob/master/doc/source/cpp_basic_tutorial.rst)

tetsuok:
Done.

PR #868, opened 2017-09-06
dynet-867. Fix segfault when only lookup_parameters.

xunzhang:


xunzhang:
cc @davidweichiang

neubig:
@xunzhang Thanks! Could you add this as a unit test as well?

xunzhang:
Sure

PR #874, opened 2017-09-07
Use 'import dynet as dy' in the examples.

### LINKED ISSUE #869: don't import * in Python examples
In a lot of our Python examples, we do `from dynet import *`. According to PEP-8 (the Python style guide), wildcard imports are dispreferred. Instead, we should use by default `import dynet as dy`.

tetsuok:
Fixes #869.

Not sure about whether we should fix the examples in tutorials (*.pynb files). I'm fine either way.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/874)


neubig:
Thanks a lot!

tetsuok:
You’re welcome :)

PR #880, opened 2017-09-09
Optionally enable Debug build, ASan enabled or TSan enabled binaries

tetsuok:
Currently, DyNet is built with all optimizations enabled by default.
Optionally allowing debug builds would be useful for debugging crashes
or data races, enabling AddressSanitizer (aka ASan) or ThreadSanitizer
(aka TSan).

With this change, developers can tell cmake to create Debug build,
by specifying `-DCMAKE_BUILD_TYPE=Debug` without editing CMakeLists.txt.
If `-DCMAKE_BUILD_TYPE` is not specified, Release build is built as
before.
So, this change requires no burden for standard users.

Typical workflow to create Debug builds, ASan enabled binaries, or
TSan enabled binaries would be like:

$ mkdir Debug
$ cd Debug
$ cmake .. -DCMAKE_BUILD_TYPE=Debug -G Ninja
$ ninja

ASan enabled binaries
$ mkdir Debug-asan
$ cd Debug-asan
$ cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-fsanitize=address" -G Ninja
$ ninja

TSan enabled binaries
$ mkdir Debug-tsan
$ cd Debug-tsan
$ cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-fsanitize=thread" -G Ninja
$ ninja

Note: I'm using ninja as a build system. This applies to other build
systems such as make.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/880)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,10 @@ function(find_mkl)
   endif()
 endfunction()
 
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE "Release")
+endif()
+
 ######## Cross-compiler, cross-platform options
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DEIGEN_FAST_MATH")
 if (MKL OR MKL_ROOT)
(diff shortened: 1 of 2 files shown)


neubig:
Thanks, this is great.

Could the explanation here be added to the documentation? Perhaps in the C++ build instructions? I'm afraid that this might become lost if we don't document it.

tetsuok:
Sure. Which doc file is appropriate? This change is for DyNet developers. I think standard users don't need to know.

neubig:
Maybe either the "Installing w/ C++" file (at the end?) or on the "Debugging/Reporting Issues" page.

tetsuok:
OK. Got it.

tetsuok:
Added the explanation to "Debugging/Reporting Issues". Please take a look.
I'm not sure how developers are familiar with the tools. I'm happy to improve
the explanation based on feedback.

PR #882, opened 2017-09-10
Eliminate dependency on libdynet from _dynet.so

tetsuok:
Currently, the compiled Cython file, _dynet.so depends on libdynet.
This has a couple of disadvantages such as:

- Installation can be clumsy (e.g., setting LD_LIBRARY_PATH in Linux,
  DYLD_LIBRARY_PATH in macOS to load libdynet)
- Not easy to deploy to servers.

This change eliminates the dependency by creating a static library of
dynet, making the installation and deployment easier.  The idea is to
link the static library rather than the shared/dynamic library when
generating _dynet.so.

The static and shared/dynamic libraries are generated from an object
library (it's just a collection of object files) [1].  By creating an
object library, we can avoid compiling object files for both libraries.

[1] https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/882)


neubig:
Oh, this is great! Maybe this will fix @danielhers 's problems with installing on MacOS?

tetsuok:
@neubig #146?

danielhers:
Yes, I do believe it should help.

tetsuok:
@neubig could you check this commit works in Linux + CUDA backend? I don't have the access to Linux machine with NVIDIA GPUs.

neubig:
Sure, will do!

danielhers:
FYI I'm experimenting with CUDA build on Travis CI (not yet with this PR's change, just want to see it works first): https://travis-ci.org/danielhers/dynet/jobs/274130100

neubig:
@danielhers FYI: It fails the C++ tests because if you compile with CUDA they try to run on the GPU by default, but the GPU is not present.

danielhers:
Makes sense. Any way to force it to use the CPU? Or I'll just disable the
C++ tests for this configuration. Should still be useful for testing the
build.

neubig:
@danielhers We could probably force it to use the CPU by passing `--dynet-devices CPU`?

neubig:
@tetsuok This apparently doesn't work with CUDA, as the `cuda_add_library` function of `cmake` doesn't support the "OBJECT" option. I'm not sure the solution and don't have a lot of time to debug right now. If you want, it's possible to install CUDA on a machine even without a GPU, so you could potentially do that to help debug.

tetsuok:
@neubig OK. Let me investigate.

tetsuok:
I tried a bunch of different approaches. Turned out there is no "easy" way to create a collection of object files using CUDA cmake macros. There might be some ways of doing this, but I think it would be too sophisticated ways nobody can maintain.

So, here is a list of approaches I can think of:
1. Leave the CUDA related part as it is in this patch.
2. Change to build static library only by default when CUDA backend is specified. Build shared library if `BUILD_SHARED_LIBS=ON` is specified for those who want to build the shared library. (The latter case, users need to set LD_LIBRARY_PATH or DYLD_LIBRARY_PATH properly)
3. Build static and shared libraries by compiling twice, then link the static library when generating "_dynet.so".

I'm wondering how DyNet is used in Linux environment. If most users don't care about installing the shared libraries under / (e.g., /usr/lib, /usr/local), the first approach sounds better. Otherwise, the second approach seems good to me.

neubig:
How about we build only the static library by default, and *also* build the shared library if `BUILD_SHARED_LIBS=ON`? I think we only need the shared library if we're using the C++ bindings from an external program, so we could add this info to the C++ documentation. This seems like the simplest solution?

neubig:
@tetsuok Have you had a chance to look at this?

tetsuok:
@neubig Not yet. I will take a look tomorrow.

tetsuok:
Seems like building both static and shared libraries didn't work if `-DBUILD_SHARED_LIBS=ON` is specified (see  [if block](https://github.com/clab/dynet/pull/882/commits/d0fce31581dcd7d2e25f5de38f3afc8191b05e26#diff-cee02f37e34c24f359d88b7308c2a803R272) in commit d0fce31581dcd7d2e25f5de38f3afc8191b05e26). I got compile errors on cuda.cc when building shared library. I didn't investigate why that happened yet.

Building a shared library only worked out (see commit 1ef12069b8dad2fadc16792545a5cf7a70782ee7)

I now feel it makes sense for users to be able to choose either a static library or shared library by specifying `-DBUILD_SHARED_LIBS=ON`. Users can link their C++ programs no matter if the library is static or shared. It's okay to abandon this pull request, though.

neubig:
I'm going to close this, as I'm not sure if there's any easy way around the CUDA problem. Thanks for the contribution anyway @tetsuok !

PR #883, opened 2017-09-11
Fix -Wsign-compare warnings

tetsuok:
Currently, we see lots of warnings when building with `-DBACKEND=cuda`.
This attempts to fix -Wsign-compare warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/883)

### DIFF
--- a/dynet/nodes-lstm.cc
+++ b/dynet/nodes-lstm.cc
@@ -31,7 +31,7 @@ namespace dynet {
     unsigned input_dim=xs[num_inputs+1][1];
     unsigned batch_size=xs[0].bd;
     unsigned inputs_dim_sum=0;
-    for(int i=0; i<num_inputs; i++){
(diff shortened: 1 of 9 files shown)


tetsuok:
Sure. I'd like to fix the other types of warnings in a separate PR.
Also, I think `_dynet.cpp` is the file automatically generated by Cython.

xunzhang:
ok, totally fine. I know, could we generate no-warning `_dynet.cpp` in some way? Not sure though.

tetsuok:
Sorry I don't know right now. I was actually wondering, too.

xunzhang:
Never mind, let's push it back then.

tetsuok:
As far as I know, what we can do now is to disable particular warnings in compiling Cythoned files.
Please take a look at #887.

PR #890, opened 2017-09-11
More fine grained access to numpy LookupParams

pmichel31415:
This makes it possible to get the numpy value (or gradient value) of a lookup parameter for certain rows only (so that the whole array doesn't need to be transferred). !tested and documented


pmichel31415:
OK I'm going to merge this since the test passes and it's a relatively minor change

PR #893, opened 2017-09-12
Allow adding model parameters after update

danielhers:
Instead of allocating shadow parameters (e.g. gradient history) just at the first call to `update()`, now the trainers check if there are more parameters in the models than there are shadow parameters, and if so, allocate additional ones.

### DIFF
--- a/dynet/training.cc
+++ b/dynet/training.cc
@@ -82,10 +82,10 @@ void Trainer::update() {
   
   // Allocate if necessary
   if(aux_allocated < params.size()) {
-    aux_allocated += alloc_impl();
+    aux_allocated = alloc_impl();
   }
   if(aux_allocated_lookup < lparams.size()) {
-    aux_allocated_lookup += alloc_lookup_impl();
+    aux_allocated_lookup = alloc_lookup_impl();
   }
 
   // Perform gradient clipping and cycle through parameters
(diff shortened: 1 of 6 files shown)


danielhers:
Still getting the assertion error sometimes (randomly, it seems): https://travis-ci.org/clab/dynet/jobs/274557551
@pmichel31415 any idea what's wrong with my change?

danielhers:
The build actually succeeded, Travis just timed out on `apt-get` in one job (should try to find a way to work around that...).
I think this PR is ready.

dynet/training.cc
```diff
@@ -191,9 +194,15 @@ void MomentumSGDTrainer::update_lookup_params(real gscale, size_t idx) {
   auto & p = model->lookup_parameters_list()[idx];
   update_rule(gscale, {&p->all_values, &p->all_grads, &vlp[idx].all_h});
 }
-void MomentumSGDTrainer::alloc_impl() {
-  vp = allocate_shadow_parameters(*model);
-  vlp = allocate_shadow_lookup_parameters(*model);
+unsigned MomentumSGDTrainer::alloc_impl() {
+  auto v = allocate_shadow_parameters(*model, aux_allocated);
+  vp.insert(vp.end(), v.begin(), v.end());
```
neubig: There's a bunch of copy-pasted inserts here. It'd be better if we didn't have to copy-paste this every time. Is there any way around this by doing the modification within `allocate_shadow_parameters`, etc?

danielhers: Sure, we could pass a reference to the vector we want to append to. Would that be better?

neubig: Yep.

neubig:
Thanks!

Issue #224, opened 2016-12-25
Initializing lookup matrix


chaitanyamalaviya:
Hi everyone,

Is there a way to initialize the dynet lookup matrix with a numpy matrix?
I know that the lookup parameter object does not support item assignment, but I was hoping there would be a workaround to this.
Would appreciate any help!

mitchellstern:
Currently DyNet supports two methods of initializing `LookupParameters` from a `numpy` array:
```
>>> import numpy as np
>>> import dynet as dy

# Create a model
>>> model = dy.Model()

# Generate some sample embeddings
>>> array = np.arange(15.0).reshape((3, 5))
>>> array
array([[  0.,   1.,   2.,   3.,   4.],
       [  5.,   6.,   7.,   8.,   9.],
       [ 10.,  11.,  12.,  13.,  14.]])

# Method 1: the embeddings should be arranged as rows
>>> lp1 = model.add_lookup_parameters((3, 5))
>>> lp1.init_from_array(array)
>>> lp1[0].npvalue()
array([ 0.,  1.,  2.,  3.,  4.])
>>> lp1[1].npvalue()
array([ 5.,  6.,  7.,  8.,  9.])
>>> lp1[2].npvalue()
array([ 10.,  11.,  12.,  13.,  14.])

# Method 2: the embeddings should be arranged as columns
>>> lp2 = model.add_lookup_parameters((3, 5), init=dy.NumpyInitializer(array.T))
>>> lp2[0].npvalue()
array([ 0.,  1.,  2.,  3.,  4.])
>>> lp2[1].npvalue()
array([ 5.,  6.,  7.,  8.,  9.])
>>> lp2[2].npvalue()
array([ 10.,  11.,  12.,  13.,  14.])
```

chaitanyamalaviya:
Thanks a lot!

yoavg:
When were initializers added to add_lookup_parameters?

I recall that I left it out on purpose, because the init_from_array() method was sufficient.

The current behavior, requiring the the embeddings to be arranged in columns, looks like a buggy/counterintuitive behavior to me. It is also inconsistent with initializing regular parameters, when we do not require this transpose.

mitchellstern:
Evidently they were added in https://github.com/clab/dynet/commit/1ec1c82e9ebb296896e4457ae24269314fb3f8ff.

neubig:
In light of this: https://github.com/clab/dynet/pull/340

[...]

PR #894, opened 2017-09-13
Syntactic sugar for parameter initializers

pmichel31415:
Addresses #224 

- [x] `add_parameters`
- [x] tests
- [x] `add_lookup_parameters`
- [x] Put examples in the api tutorial

@neubig or @yoavg let me know if that sounds good, I'll wait for your approval before doing lookup parameters.


yoavg:
Looks fine to me as far as the API goes, though its a bit ugly to have `mean`, `scale` and `std` parameters, where a diff subset if considered depending on the init...

But if we are on the topic of initialization, I'm not sure the python API should follow the c++ init interface as it does now. I think the "from_numpy" initializers are sufficient for everything, and we can just rely on them, and maybe provide wrappers for generating the numpy arrays for the common ones (uniform, glorot, etc).

What are the reasons for keeping the c++ initializers? The only thing I can think of is efficiency for on-device init instead of copy, but parameters initialization is hardly a bottleneck op.

pmichel31415:
One issue might also be code redundancy but if you don't think that's an issue we can go with the numpy wrappers.

I agree that the many named arguments are ugly but I don't know of any clean way of overloading functions in cython.

neubig:
The reason for keeping the on-device initializers is because they are precisely the "wrapper for common ops," which will work for not only Python but also C++, Scala, or any other languages.

pmichel31415:
So what should I do for this?

neubig:
I'm in favor of the current API.

pmichel31415:
OK This should be good to merge once the tests pass.

This also fixes #971 (Using NumpyInit with lookup params throws an error, no weird transposition)

python/_dynet.pyx
```diff
@@ -1143,28 +1143,70 @@ cdef class ParameterCollection: # {{{
...
+            p = m.add_parameters((5,5), init='identity')                # Creates 5x5 identity matrix
+            p = m.add_parameters((5,5), init='saxe')                    # Creates 5x5 orthogonal matrix (NOT SUPPORTED YET)
+            p = m.add_parameters((3,5), init='glorot')                  # Creates 3x5 matrix with glorot init
+            p = m.add_parameters((3,5), init='he')                      # Creates 3x5 matrix with he init
+            p = m.add_parameters(np.zeros((3,5)))                       # Creates 3x5 matrix from a numpy array (size is inferred)
+            p = m.add_parameters((3,5), init=dy.PyInitializer())        # Any parameter initializer
         
         Args:
-            dim (tuple): Shape of the parameter
+            dim (tuple, np.ndarray): Shape of the parameter. If only a numpy array is passed the dimension is inferred
```
chunyang-wen: I think it is a little confusing that we specifially handle when it is np.ndarray for the `dim`. Users can use NumpyInitializer or `parameters_from_numpy`. I think dim can also be specified by np.array([2,3]).

python/_dynet.pyx
```diff
@@ -1143,28 +1143,70 @@ cdef class ParameterCollection: # {{{
...
         cdef CParameterInit *initializer
         cdef CDevice *dev
         cdef string _name = <string> name.encode("utf8")
+        cdef PyInitializer pyinit
         if init is None:
-            init = GlorotInitializer()
-        initializer = init.initializer
+            pyinit = GlorotInitializer()
+        else:
+            if isinstance(init, (int, float)):
```
chunyang-wen: It is better that we can support use python list/tuple to initialize parameters.

```python
init = [1,2] # or (1, 2)
a = m.add_parameters((1,2), init=init)
```

pmichel31415: I can fix this

python/_dynet.pyx
pmichel31415: I agree that the name is confusing. However there is no easy way to overload functions and I would really like to make it an option to just specify a numpy array and let the dimension be inferred. Using NumpyInitializer is not an option for LookupParameters because of reshaping.

chunyang-wen:
```python
model.add_parameters(arr.shape, init=arr)
```

I think it is better that we keep a **consistent interface**. If we reuse dim, it may be easier to initialize for parameters from numpy array. But I doubt whether it is suitable to make the compromise.

neubig:
Sorry review took so long. This looks good, but could you merge master?

chunyang-wen:
I think a rebase is better than a merge from master.

neubig:
Thanks!

PR #895, opened 2017-09-14
RNN-related enhancements to Scala bindings

shuheik:
Namely, enable control over updates of RNN parameter (#547), set dropout masks per sentence (#812) and add CompactVanillaLSTMBuilder (#729).

Also, fixed file path in the RnnLanguageModelBatch example.


neubig:
Thanks!

PR #902, opened 2017-09-17
examples: Use --hidden_size to set the size of hidden vectors

tetsuok:
In rnnlm/README.md, "-H" is used to specify the size of hidden vectors.
There is no such option, though.  Current command line parser silently ignores the non-existent options. This is unfortunate for beginners who want to start with C++ examples.
We should fix the parser so that it can report errors on the use of non-existent options.

Also, "-x" is used in README.md but I can't find such option either. What's the correct option for it?
"--bidirectionnal"?

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/902)


tetsuok:
@neubig Do you know what `-x` is used for?

neubig:
No.

tetsuok:
OK. Thanks.

tetsuok:
Figured out it by going back to commit 3605089a. "-x" was used to set the mode to train models.

PR #916, opened 2017-09-20
Python multi device

yoavg:
Multi-device support(https://github.com/clab/dynet/issues/92) in the python api. 

Currently there is no dedicated `Device` class, and we refer to devices by their string names.

This seem less cumbersome and more Pythonic to me. However, if we foresee a case where we would really want Device to be an object, we probably should add this now for backward compatibility.

The API is: 
- `dy.available_devices()` returns a list of device names. 
- `dy.to_device(expression, dev_name)` copies an expression across   devices.
- `param_colection.add_parameteres(..., device=name)`
- `param_colection.add_lookup_parameteres(..., device=name)` -
- `param_colection.parameteres_from_numpy(..., device=name)`
- `param_colection.lookup_parameteres_from_numpy(..., device=name)`

The devices mechanism documentation is in docstring of `available_devices()`

There is also a `DeviceInfo` class and `get_device_info(str)` stubs, which we could extend later. I think it could be nice to allow querying devices for their available memory, both on the Python and C++ apis.

[note: I didn't yet tested it on a GPU machine]

### DIFF
--- a/python/_dynet.pxd
+++ b/python/_dynet.pxd
@@ -223,6 +223,13 @@ cdef extern from "dynet/training.h" namespace "dynet":
     cdef cppclass CAdamTrainer "dynet::AdamTrainer" (CTrainer):
         CAdamTrainer(CModel& m, float alpha, float beta_1, float beta_2, float eps) # TODO removed lam, update docs
 
+cdef extern from "dynet/devices.h" namespace "dynet":
+    cdef cppclass CDevice "dynet::Device":
+        pass
+    cdef cppclass CDeviceManager "dynet::DeviceManager":
+        CDevice* get_global_device(string name)
+
+    CDeviceManager* c_get_device_manager "dynet::get_device_manager" () 
 
 cdef extern from "dynet/expr.h" namespace "dynet":
     cdef cppclass CExpression "dynet::Expression":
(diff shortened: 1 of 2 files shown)


yoavg:
Ah! I indeed forgot about those. Good catch!

Added device to all (I think) input expressions.

Also, now python raises a proper exception and does not segfault on bad device names.

xunzhang:
I just tested it and got errors below:
```
In [2]: x = dy.vecInput(2, "GPU:0")
Exception TypeError: 'instance exception may not have a separate value' in '_dynet.Dim' ignored
````

yoavg:
This should fix it.

xunzhang:
I have tested it and I think basically it is working now. I am gonna to merge this now and will send a pull request adding some tests.

jbuckman:
I've been unable to get this to work on a cloud Ubuntu 16.04 VM with K80s. 

The following command:

python xor-multidevice.py --dynet-devices CPU,GPU:0

...works as intended. Additionally, if I go through all of the locations in the code that say GPU:0 and change them to GPU:1, the following command works too:

python xor-multidevice.py --dynet-devices CPU,GPU:1

But I run into issues when I try to load multiple GPUs at once:

python xor-multidevice.py --dynet-devices CPU,GPU:0,GPU:1
[dynet] initializing CUDA
[dynet] Request for 2 specific GPUs ...
[dynet] Device Number: 0
[dynet]   Device name: Tesla K80
[dynet]   Memory Clock Rate (KHz): 2505000
[dynet]   Memory Bus Width (bits): 384
[dynet]   Peak Memory Bandwidth (GB/s): 240.48
[dynet]   Memory Free (GB): 11.901/11.9956
[dynet]
[dynet] Device Number: 1
[dynet]   Device name: Tesla K80
[dynet]   Memory Clock Rate (KHz): 2505000
[dynet]   Memory Bus Width (bits): 384
[dynet]   Peak Memory Bandwidth (GB/s): 240.48
[dynet]   Memory Free (GB): 11.9177/11.9956
[dynet]
[dynet] Device(s) selected: 1 0
[dynet] random seed: 146991311
Available devices ['GPU:1', 'GPU:0', 'CPU']
python: /home/buckman/dynet-base/eigen/unsupported/Eigen/CXX11/src/Tensor/TensorExecutor.h:262: static void Eigen::internal::TensorExecutor<Expression, Eigen::GpuDevice, Vectorizable>::run(const Expression&, const Eigen::GpuDevice&) [with Expression = const Eigen::TensorAssignOp<Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long int>, 0, Eigen::MakePointer>, const Eigen::TensorCwiseNullaryOp<Eigen::internal::scalar_constant_op<float>, const Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long int>, 0, Eigen::MakePointer> > >; bool Vectorizable = false]: Assertion `cudaGetLastError() == cudaSuccess' failed.
Aborted (core dumped)

I didn't use GPU:0 anywhere in the code; all references are explicitly using GPU:1. The only change is that I am now loading both devices.

xunzhang:
There might be some issues when specifying more GPUs. I will take a look and try to fix this. I think at least the case specifying more than one GPUs(and just use one) should work.

PR #919, opened 2017-09-21
Skip most Travis CI configs on PRs to reduce load

danielhers:


neubig:
Thanks for fixing this quickly! I think we should keep at least one Python 2 configuration (not sure whether Mac or Linux is more common), as Python 2 often breaks when we do something silly.

danielhers:
Added back one 2.7 setup.

neubig:
Sorry I wasn't clear... We should do this for both pip and manual install, in case there are Python 2 specific errors in setup.py that only fail for pip or manual install.

danielhers:
Right. Enabled pip too.

neubig:
Thanks!

PR #924, opened 2017-09-22
Update multi-device document

xunzhang:


doc/source/commandline.rst
```diff
@@ -28,6 +28,8 @@ options.
    `here <http://dynet.readthedocs.io/en/latest/minibatch.html>`_.
 -  ``--dynet-gpus NUMBER``: Specify how many GPUs you want to use, if
    DyNet is compiled with CUDA. Currently, only one GPU is supported.
+-  ``--dynet-gpu``: Specify whether to use GPU or not. Note that it is the only option
```
neubig: Do you mean "this is only an option for Python programs"? Is there a reason why we can't support it for C++ also?

xunzhang: I've no idea actually. The old version of DyNet only has this option in the Python layer. I just follow the previous rule(https://github.com/clab/dynet/blob/9606e2a3fdc31788b3739adfcfb498b1626645fd/python/dynet.py#L5).

xunzhang:
It seems this blocks the readthedocs build: https://readthedocs.org/projects/dynet/builds/6012492/ when building `python_ref`. Let me fix it.

pmichel31415:
I can take a look at readthedocs @xunzhang

pmichel31415:
I think it's just that you are using `autoclass` for a function (`to_device`).

I will fix directly on the master branch

xunzhang:
@pmichel31415 Oh right. Thanks a lot!

Issue #191, opened 2016-11-24
Better organized examples, and parity between C++ and Python


neubig:
At the moment, the examples that are implemented are kind-of random, and there is no parity between the Python and C++ examples. I think we should re-organize the example section to make sure that:

1. The examples cover all the functionality of DyNet and demonstrate interesting things that it can do.
2. All examples are simple and canonical. For example, they don't contain anything that would be unexpected based on the models they are implementing (https://github.com/clab/dynet/issues/141).
3. There is parity between the C++ and python examples, so neither language is left out.
4. There are not examples that are overly complicated and hard to maintain.

neubig:
I have a proposal for fixing this:

1. Restructure the examples so it is `examples/MODEL_NAME/LANGUAGE_NAME` (e.g. `examples/rnnlm/py`)
2. Create a `README.md` in `examples` that briefly describes each example, and explains which language it is available for.
3. Create a `README.md` in each `examples/MODEL_NAME` directory. All `README.md` files should explain how to a) get data with which to test the model, and b) run the example.

In addition, we should be proactive about adding new examples. Having examples is very useful for people just starting out with the toolkit.

zhisbug:
Will start looking at this issue soon.

neubig:
The examples are now better organized by #927, so I'll close this.

PR #927, opened 2017-09-22
Restructure examples

neubig:
This is a first step towards better organized examples (https://github.com/clab/dynet/issues/191).
See the README here for details:
> https://github.com/clab/dynet/tree/restructure-examples/examples

This is WIP because I need to fix this to make sure the cpp examples compile, and the Python examples run.


stefan-it:
@neubig The *Text Categorization* example in the `traincat` will currently not be compiled, as it still uses the `conv1d` operator (was removed with [this](https://github.com/clab/dynet/commit/29bbd6bcdc6854a0f3e379e738d06b68c7ee276f) commit). Is there any alternative available to get the example code running?

neubig:
This is compiling now, so I think this is ready to merge if tests pass (but they might fail because of the Conda/Travis combination bug?)

@stefan-it This would need to be re-written to use conv2d. You can find an example in Python here: https://github.com/neubig/nn4nlp2017-code/blob/master/05-cnn/cnn-class.py
I myself don't have time to fix the example now, but a PR would certainly be welcome!

PR #929, opened 2017-09-23
Some partial supports for READ-only inplaced operations.

zzsfornlp:
This PR adds certain partial enhancements for READ-only in-place operations, replacing another PR #840 which brings too much complexities and might bother the users.

The concerned Nodes are the ones which can safely reuse memories (no rewriting for forward values):
Reshape(forward/backward), Identity(forward/backward), NoBackprop(forward-only), FlipGradient(forward-only).

Some notes:
1. Since `NoBackprop` and `FlipGradient` need special gradient manipulation, they have their own backward memories.
2. For auto-batch mode, currently simply forbid batching for thoes nodes. Firstly the memory-allocation part of auto-batching seems a little complex for me, so I do not change much there in the fear of breaking sth; Secondly, in auto-batching mode, maybe memory is even less important.
3. (TODO) For the memory saving of WRITE nodes, we might need some better mechanism for them.

* In fact, I'm not quite sure about the dealing of the pointers in the ExecutionEngines when doing memory allocating or sharing, wonder if they are appropriate. (Although all the tests have been passed)


neubig:
Sorry it took so long to review this! It basically looks good, but I need to check autobatching support. I'll try to take a closer look later today.

neubig:
Actually, on a closer look it seems that it's probably OK to just disable batching multiple nodes when in-placing, and leave this for another PR. One question: there are several checks in `forward_dev_impl()` that are checking whether the operation is in-placed. Won't these checks always be true? I don't think we need the conditional statement there (and maybe just a comment that the operation is a no-op).

zzsfornlp:
Yes, currently this is always true because those read-only nodes are inplaced by default; but I wonder whether there are situations where inplaced operation should be prohibited? And thus still leaving part of the switches there. 
I'll mdify those correspondingly.

neubig:
Really sorry I've taken so long on checking this, but I definitely would like to incorporate this! The only reason why it's taking time is because I want to make sure we don't accidentally break anything, and in order to do so I'll have to do comprehensive tests.

zzsfornlp:
Sorry about closing this previously, it is just that it seems there are certain changes and conflicts in `exec.cc/dynet.cc`, and I have not been following those. I'm worried that there might be some inconsistencies. And yes, this should be tested comprehensively and taken care specially for it brings memory sharing among nodes.

neubig:
Hi @zzsfornlp , thanks for committing this! Unfortunately when cleaning the repository history of a large file to reduce download size it messed up this commit. Would you mind pulling the new master, then cherry-picking the relevant commits, and re-submitting a pull request? If you prefer, I could also do this myself, but it would probably result in github counting it as my PR instead of yours.

zzsfornlp:
Of course, I will repull the master branch and reopen a new PR (in a couple of days maybe), thus I'll close this one first.
Another thing is that sorry but I have not been following the new commits and changes between last commit and the current point. I'm not quite sure whether there will be any logical conflicts. Thus about this I might need some help. Thanks!

PR #932, opened 2017-09-26
AppVeyor badge

danielhers:


neubig:
It seems that AppVeyor is still failing?

danielhers:
That's odd. Nothing changed there since the last commit, which succeeded. Maybe it's a transient error and restarting the build will fix it (but I don't have permissions to do that).

Issue #933, opened 2017-09-27
DyNet no longer searches for open GPU


neubig:
DyNet used to search for an open GPU and pick the one with the most open memory. This was convenient behavior, but now apparently when setting the `--dynet-gpu` flag in Python it simply allocates to GPU0. It would be nice if we could get the old behavior back.

xunzhang:
Fixed by https://github.com/clab/dynet/pull/934. I have tested that and the latest code works.

PR #934, opened 2017-09-27
replace --dynet-gpu with --dynet-gpus 1 instead of --dynet-devices GPU:0

chunyang-wen:
#933 
In _dynet.pyx, I noticed that the default configuration is set.

> the python gpu switch, use GPU:0 by default

By removing the replacement procedure,  `dynet/init.cc` will choose appropriate GPU (with most available memory)


xunzhang:
Thanks!

PR #935, opened 2017-09-27
Add const_parameter to python docs

EggyLv999:
Function is missing documentation! Maybe this split should also be referred in Parameter.expr() docs?


neubig:
Thanks!

PR #946, opened 2017-09-29
Fix typo in tutorial-1-xor.ipynb

ZiyaoWei:
The tutorial says "a single hidden node", and one line later "a hidden layer of 8 nodes" (which seems to be the case).


xunzhang:
Thanks.

PR #947, opened 2017-09-29
Multi-gpu support for basic add_parameters function.

xunzhang:
Address comment under dynet-916(https://github.com/clab/dynet/pull/916#issuecomment-331745925). With this enhancement, I think basically DyNet can support the multi-GPU case. But subtle changes are needed later on to switch between GPU devices before invoking CUDA calls.

### DIFF
--- a/dynet/init.cc
+++ b/dynet/init.cc
@@ -214,6 +214,12 @@ void initialize(DynetParams& params) {
   Device *d = new Device_CPU(device_manager->num_devices(), params.mem_descriptor, params.shared_parameters);
   device_manager->add(d);
   default_device = device_manager->get(default_index);
+#if HAVE_CUDA
+  if (default_device->type == DeviceType::GPU) {
+    auto default_gpu_device = static_cast<Device_GPU *>(default_device);
+    CUDA_CHECK(cudaSetDevice(default_gpu_device->cuda_device_id));
+  }
+#endif
 
   // TODO these should be accessed through the relevant device and removed here
   kSCALAR_MINUSONE = default_device->kSCALAR_MINUSONE;
--- a/dynet/model.cc
+++ b/dynet/model.cc
@@ -54,6 +54,12 @@ ParameterStorage::ParameterStorage(const Dim& d, float scale, const std::string 
     : name(name), dim(d), updated(true), nonzero_grad(false), owner(nullptr), device(dev) {
   DYNET_ARG_CHECK(default_device != nullptr,
                   "Attempting to define parameters before initializing DyNet. Be sure to call dynet::initialize() before defining your model.");
+#if HAVE_CUDA
+  if (dev->type == DeviceType::GPU) {
+    auto gpu_dev = static_cast<Device_GPU *>(dev);
+    CUDA_CHECK(cudaSetDevice(gpu_dev->cuda_device_id));
+  }
+#endif
   values.d = g.d = d;
   values.device = g.device = device;
   device->allocate_tensor(DeviceMempool::PS, values);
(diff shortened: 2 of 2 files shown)


xunzhang:
cc @neubig @jbuckman

neubig:
Thanks! I'm not sure if it's a result of this PR, but the `xor-multidevice.py` example is failing on Python 3 for me on the following error:

      File "../examples/python/xor-multidevice.py", line 19, in <module>
        pW = m.add_parameters((HIDDEN_SIZE, 2)) # GPU:0 by default 
      File "_dynet.pyx", line 1117, in _dynet.ParameterCollection.add_parameters (_dynet.cpp:21971)
      File "_dynet.pyx", line 1142, in _dynet.ParameterCollection.add_parameters (_dynet.cpp:21818)
      File "stringsource", line 15, in string.from_py.__pyx_convert_string_from_py_std__in_string (_dynet.cpp:77077)
    TypeError: expected bytes, str found

I suspect that this is similar to this: https://github.com/clab/dynet/commit/5d44a8d948e425bd1504760a3c091e48c33a2067

@xunzhang do you have a moment to confirm if you can reproduce on Python 3?

jbuckman:
This was successful in fixing the error in the case where I load multiple GPUs but only use 1. However, when I load multiple GPUs and also use multiple GPUs, the error returns:

```
python: /home/buckman/dynet-base/eigen/unsupported/Eigen/CXX11/src/Tensor/TensorExecutor.h:262: static void Eigen::internal::TensorExecutor<Expression, Eigen::GpuDevice, Vectorizable>::run(const Expression&, const Eigen::GpuDevice&) [with Expression = const Eigen::TensorAssignOp<Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long int>, 0, Eigen::MakePointer>, const Eigen::TensorCwiseBinaryOp<Eigen::internal::scalar_sum_op<float, float>, const Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long int>, 0, Eigen::MakePointer>, const Eigen::TensorMap<Eigen::Tensor<float, 1, 0, long int>, 0, Eigen::MakePointer> > >; bool Vectorizable = false]: Assertion `cudaGetLastError() == cudaSuccess' failed.
Aborted (core dumped)
```

I ran xor-multidevice.py with parameters pW and pb located on GPU:0, instantiated them with dy.parameter, and then moved them to GPU:1 with dy.to_device, throwing the error.

xunzhang:
Yes. Actually it is more like a fix for add_parameters function under multi-gpu mode as the title mentioned. And we need more refactoring of the code to fully support multi-GPU. I will make further pull requests for supporting that and will cc you when that works eventually.

Send from my iPhone

jbuckman:
Ah, okay. I need multi-GPU fairly soon (next few weeks) to run some experiments, and I'd be happy to help with whatever refactoring is needed. Could you point me towards the parts of the code that need to be updated?

xunzhang:
For example, I think we need to add a customized cudaMemcpy what will switch to the right device id before the syscall. Other CUDA calls are also needed to verify executing to the correct device.

xunzhang:
@neubig what about the python3 bug you met. Have you found the root cause?
.

neubig:
@xunzhang I've fixed the Python3 bug.
I can help with this a little if you're busy (I'd like to finish this ASAP). I'm guessing that we mainly need to look at the non-Eigen CUDA calls?

xunzhang:
Currently under GPU mode, after this line(https://github.com/clab/dynet/blob/master/dynet/model.cc#L93) if we try to print out `values.tvec()`, it will cause a core dump. But I don't have time to fix that under next week.

PR #950, opened 2017-09-29
Native Python on Linux on Travis CI instead of Conda

danielhers:
MacOS doesn't have native Python on Travis CI, so we still need Miniconda, but the problem we had lately is only on Linux anyway.
This PR also includes some reorganization of the Travis CI scripts to simplify them.


danielhers:
Done.

PR #957, opened 2017-10-04
exception handling for model related API

pcyin:
if `name` argument has invalid character (e.g, `/`) python shell will crash.


neubig:
Thanks!

Issue #952, opened 2017-09-30
Multi-GPU support still not finished yet


neubig:
Apparently multi-GPU support is still not working completely. See https://github.com/clab/dynet/pull/947 and https://github.com/clab/dynet/pull/916#issuecomment-331745925

Hopefully we can finish up the remaining things soon.

xunzhang:
Resolved via https://github.com/clab/dynet/pull/967, https://github.com/clab/dynet/pull/994, and https://github.com/clab/dynet/pull/999.

neubig:
Awesome, thank you! I think we can think of doing a minor version release pretty soon.

xunzhang:
Sure, let's do that.

PR #967, opened 2017-10-06
Full multi-GPU support

neubig:
A start at addressing https://github.com/clab/dynet/issues/952 (which I'll do with @xunzhang )

I'm not sure the extent with which we'll have to ensure that we switch devices appropriately before performing CUDA operations.


xunzhang:
@neubig It works now! I think we can merge this now.
@jbuckman I am a little busy last week and this week. Now with code in this pull request, I think the basic `xor-multidevice.py` example could successfully run using multiple GPUs. But there are still several places we need to switch to the correct device(I will try to fix all of them next week). After merging this, please try your code again. If you still get issues, leave comments for me and I think I can do quick fix for you. Sorry to have you waiting for this feature a while.

dynet/model.cc
```diff
@@ -562,7 +562,7 @@ template void ParameterStorage::accumulate_grad_dev<Device_CPU>(Device_CPU & dev
 void ParameterStorage::accumulate_grad(const Tensor& d) {
   nonzero_grad = true;
   if (values.device->type == DeviceType::CPU) { accumulate_grad_dev(*(Device_CPU*)values.device, d); }
-  else if (values.device->type == DeviceType::GPU) { accumulate_grad_dev(*(Device_GPU*)values.device, d); }
+  else if (values.device->type == DeviceType::GPU) { cudaSetDevice(((Device_GPU*)values.device)->cuda_device_id); accumulate_grad_dev(*(Device_GPU*)values.device, d); }
```
neubig: This should probably be multiple lines.

dynet/nodes-affinetransform.cc
```diff
@@ -75,6 +75,11 @@ std::vector<int> AffineTransform::autobatch_concat(const ComputationGraph & cg)
 // much faster than using Eigen's tensor contractions (as of the writing)
 template<class MyDevice>
 void AffineTransform::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
+#if HAVE_CUDA
```
neubig: We shouldn't add a check to individual nodes, this will cause code bloat. Is the check added to `forward()` in `nodes-macros.h` not enough?

dynet/nodes-matrixmultiply.cc
```diff
@@ -77,42 +77,47 @@ void MatrixMultiply::backward_dev_impl(const MyDevice & dev,
                              Tensor& dEdxi) const {
   DYNET_ASSERT(i < 2, "Failed dimension check in MatrixMultiply::backward");
   int max_b = max(xs[0]->d.bd, xs[1]->d.bd);
+#if HAVE_CUDA
```
neubig: Similarly, we shouldn't add this to the nodes here.

dynet/training.cc
```diff
@@ -15,7 +15,7 @@
   template void MyTrainer::update_rule_dev<Device_CPU>(const Device_CPU & dev, real gscale, const std::vector<Tensor*> & values); \
   void MyTrainer::update_rule(real gscale, const std::vector<Tensor*> & values) { \
     if(values[0]->device->type == DeviceType::CPU) { update_rule_dev(*(Device_CPU*)values[0]->device,gscale,values); } \
-    else if(values[0]->device->type == DeviceType::GPU) { update_rule_dev(*(Device_GPU*)values[0]->device,gscale,values); } \
+    else if(values[0]->device->type == DeviceType::GPU) { cudaSetDevice(((Device_GPU*)values[0]->device)->cuda_device_id); update_rule_dev(*(Device_GPU*)values[0]->device,gscale,values); } \
```
neubig: Let's change this to multiple lines.

neubig:
Hmm, there is a 1s performance regression in this benchmark, probably due to the added `cudaSetDevice` calls upon every call to forward: https://gist.github.com/neubig/7d874d327e7d700bb609479f6cf359bb

We probably need to add a check that avoids this call if we're already on the correct device.

xunzhang:
Have you compared it to the latest master? I tried to do 200000 `cudaSetDevice` in `init.cc`, but it didn't slow down the original program.

neubig:
Sorry, I took another look, and I think my result was simply an artifact of the fact that I was counting initialization time. This looks good to go!

PR #968, opened 2017-10-08
Use reinterpret_cast to cast from 'const size_t' to 'float*'

tetsuok:
Currently, HEAD of DyNet doesn't compile when CUDA backend is used.
This fixes the build error introduced in 07381369d.

---
This change is [Reviewable](https://reviewable.io/reviews/clab/dynet/968)


tetsuok:
I'll close this as the fix was pushed already in commit b992b4032bc47fb0bb8a49d689d7b25f677346a3.

PR #977, opened 2017-10-12
Multidim2

msperber:
- fix an issue with broadcasting along batch dimension
- make it possible to overwrite normalizer in moment operations. This is helpful for computing moments over masked expressions, e.g. for implementing batch normalization.


msperber:
BTW overwriting the normalizer is a bit hacky and obviously open for discussion. I did include it in the PR because masking is an essential feature, and without this it was almost impossible to compute masked standard deviations. It might be worth thinking about more principled masking support in the future, which would go nicely with DyNet's convenient handling of minibatches.

neubig:
LGTM!

PR #979, opened 2017-10-12
An attempt at more numerically stable softmax

neubig:
Previously softmax was calculating things in log space unnecessarily. This removes an unnecessary log and hopefully makes things a little bit more stable.


neubig:
Two more fixes to this that make softmax/log softmax faster on CPU: 4x for softmax, 2.2x for log softmax.

PR #984, opened 2017-10-15
 Run Python unit tests on AppVeyor

danielhers:
The tests didn't work, maybe because the global Conda environment was used. Now creating a local one with the correct Python version.

Also enabled deployment of Windows wheels on tag creation, but note it will currently work only on my fork because I encrypted my PyPI password using my AppVeyor account - you can't encrypt data using another AppVeyor user's account. @neubig, if you want to enable this you'll need to change `TWINE_USERNAME: danielh` to `TWINE_USERNAME: neubig` and `secure: cpMn69tLzuhwO7EPhhiVwA==` to the value you'll get by [encrypting](https://ci.appveyor.com/tools/encrypt) your PyPI passwod.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,6 +9,9 @@ environment:
     BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib64-msvc-14.0
     MSVC: 1
     MINICONDA: "C:\\Miniconda3-x64"
+    TWINE_USERNAME: danielh
+    TWINE_PASSWORD:
+      secure: cpMn69tLzuhwO7EPhhiVwA==
   matrix:
 #    - PYTHON_VERSION: 3.5
 #      PYTHON_INSTALL: manual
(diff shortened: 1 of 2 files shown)


neubig:
Thanks! I don't have time to figure this out today, but because @danielhers because you've done a huge amount of work on the CI stuff I'll add you as a contributor. (As before, in most cases try to send PRs instead of committing stuff directly, so someone else can take a look at it before you finally commit.) Hopefully this will allow you to associate AppVeyor with your account and fix the problem. We could also potentially create an AppVeyor for the `clab` group if you think that'd be a better option.

danielhers:
Thanks. I won't abuse this. But I think I still don't have sufficient permissions because I don't see the Settings section in clab/dynet, where the Webhooks settings are, so I can't change the AppVeyor webhook to use my account rather than yours.

neubig:
Send me the URL (via email if it's sensitive) and I'll add it.

danielhers:
Done.

neubig:
And done.

danielhers:
Thanks. Changing the webhook caused a failure because I had hard-coded the build folder path, but this is now fixed.

Issue #982, opened 2017-10-15
Travis CI push trigger is disabled


danielhers:
Builds are enabled only for PRs and cron: https://travis-ci.org/clab/dynet/builds
So when looking at [commits](https://github.com/clab/dynet/commits/master) you can't see if one broke the build.
Is this intentional? If the reason is too many jobs then we can enable the full set only for cron (not push+cron as it is now).

neubig:
This is for two reasons:
1. pushes were triggering the full suite of tests.
2. whenever you send a PR it does a duplicate test, one for the push and one for the PR.

Both of these were wasting time on Travis. If both of these are resolved I think we can turn it back on, but if not I'm relatively happy with it remaining off.

danielhers:
#985 resolves both these issues.

pmichel31415:
I thin this one can be closed then?

danielhers:
I think the push trigger is still disabled; only PR and cron are enabled.

danielhers:
@neubig looks like Travis CI push trigger is still disabled. Just making sure this is intentional...

PR #985, opened 2017-10-15
Run full Travis CI test suite only on cron jobs

danielhers:
Related to #982.


danielhers:
The AppVeyor fail is unrelated and fixed by #984.

PR #986, opened 2017-10-15
Change model parameters to be stored as shared pointers

danielhers:
Deleting the ParameterCollection will no longer delete the Parameter objects. Instead, they will be deleted only when their last reference is deleted.

### DIFF
--- a/contrib/swig/dynet_swig.i
+++ b/contrib/swig/dynet_swig.i
@@ -121,8 +121,8 @@ namespace std {
   %template(LongVector)                   vector<long>;
   %template(StringVector)                 vector<std::string>;
   %template(ExpressionVector)             vector<dynet::Expression>;
-  %template(ParameterStorageVector)       vector<dynet::ParameterStorage*>;
-  %template(LookupParameterStorageVector) vector<dynet::LookupParameterStorage*>;
+  %template(ParameterStorageVector)       vector<shared_ptr<dynet::ParameterStorage>>;
+  %template(LookupParameterStorageVector) vector<shared_ptr<dynet::LookupParameterStorage>>;
   %template(ExpressionVectorVector)       vector<vector<dynet::Expression>>;
   %template(ParameterVector)              vector<dynet::Parameter>;
   %template(ParameterVectorVector)        vector<vector<dynet::Parameter>>;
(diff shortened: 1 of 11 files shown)


neubig:
I'm pretty sure this will not be a problem, but just in case it might be worth running a benchmark to make sure this has no speed implications. Unfortunately we don't have a good benchmark suite that can be run automatically, but something in the `neulab/dynet-benchmark` repository or https://gist.github.com/neubig/7d874d327e7d700bb609479f6cf359bb might be good.

danielhers:
I tried running the gist but I don't have `wsj.pkl`.
By the way, why don't you think these benchmarks are good for running automatically (e.g. once a week)?

xunzhang:
wsj.pkl(https://gist.githubusercontent.com/hal3/8c170c4400576eb8d0a8bd94ab231232/raw/76f57c198cb7afdc5122e413c2a3023ed024b643/wsj.pkl)

neubig:
I think it'd be great to run the benchmarks weekly, we just don't have it setup to be done automatically. If you could try to set something up that'd be greatly appreciated

chunyang-wen:
I think it is strange to remove all the namespace restrict: `std::`.

danielhers:
@chunyang-wen why are they needed when there's `using namespace std`?

chunyang-wen:
In order to write `shared_ptr` instead of `std::shared_ptr`,  namespace of std is imported. namespace is used to restrict name collisions. Normally, `use namespace std` should be restricted to a minimal case. Of course it is ok to add `use namespace std;` at the head of the program.

[Namespace of Cpp Style from Google](https://google.github.io/styleguide/cppguide.html#Namespaces)

image

All in all, it is just a suggestion.

danielhers:
Fair enough, I added it back.

Here is the result of the [wsj benchmark] (https://gist.github.com/danielhers/0fe8a922b91048618f054cf16915eef0) (after fixes for Python 3):

	% time python dynet-tagger.py
	[dynet] random seed: 2715362074
	# test on wsj subset
	13202.740951538086
	2609.627410888672
	1421.740089416504
	1276.03759765625
	1159.3768157958984
	1013.7634506225586
	951.3451042175293
	892.1206817626953
	838.1738700866699
	795.3973922729492
	0.797u 0.496s 0:00.84 152.3%    0+0k 0+0io 0pf+0w

Not sure how that compares to the 1.749s reported in the file. I'm running on Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz.

neubig:
I tested both this and the master branch, looks good!

PR #997, opened 2017-10-18
Sum dim

neubig:
The sum over dimensions operation wasn't in a nice form like the moment over dimensions operations. This PR fixes this.

### DIFF
--- a/doc/source/python_ref.rst
+++ b/doc/source/python_ref.rst
@@ -211,33 +211,46 @@ Arithmetic operations
 
 .. autofunction:: dynet.bmax
 
-.. autofunction:: dynet.transpose
-
-.. autofunction:: dynet.sum_cols
+Reduction/moment operations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. autofunction:: dynet.sum_elems
 
-.. autofunction:: dynet.mean_elems
-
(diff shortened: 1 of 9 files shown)


pmichel31415:
I think having to specify a boolean b (and to a lesser extent having to specify a list even for a single dimension) makes the interface more complicated and less intuitive.

Imo b should be defaulted to `False` at least

neubig:
I think it does default to false. And yes, it might be worth not having to specify a list. If so, we should fix the moment operations as well.

pmichel31415:
It still requires the user to specify `b` as a positional argument (in python, sorry if that wasn't clear), eg:

```python
dy.mean_dim(dy.zeros((3,3)), d=[1])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-3dea0b6fc593> in <module>()
----> 1 dy.mean_dim(dy.zeros((3,3)), d=[1])

/home/pmichel31415/libs/dynet/build/python/_dynet.pyx in _dynet.mean_dim()
   3238
   3239
-> 3240 cpdef Expression mean_dim(Expression x, list d, bool b):
   3241     """Mean along an arbitrary dimension
   3242

TypeError: mean_dim() takes exactly 3 positional arguments (2 given)
```

PR #998, opened 2017-10-18
Remove useless external dependency on easylogging

pmichel31415:
There was a dependency on easylogging via a git submodule for an example (which has been deleted since then I believe).

This PR removes the dependency.


neubig:
Thanks!

Issue #987, opened 2017-10-16
DyNet "Examples" Doc is broken


neubig:
Here:
http://dynet.readthedocs.io/en/latest/examples.html

I think it might be better to just completely remove this and make it a link to the examples directory in the GitHub.

pmichel31415:
Agreed, will do later this week

PR #1003, opened 2017-10-19
Fix examples

pmichel31415:
This replaces the stale example section in the doc with the README from the example folder.

You can see an example here: http://dynet.readthedocs.io/en/fix_examples/examples.html

Incidentally, this allows us to use markdown in the documentation, and in particular to link to md files from the repo directly. A small disadvantage is that links must be replaced with hyperlinks in order to work in readthedocs.

Addresses #987


neubig:
Thanks!

Issue #1002, opened 2017-10-19
dynet treats differently for numpy.array and python list


chunyang-wen:
```python
import dynet
import numpy as np
model = dynet.ParameterCollection()
pX = model.add_parameters((2))
pX.set_value(np.array([2,0]))

v = dynet.parameter(pX)
print v.value()

# pX.set_value([2,0])
```

Using python list will get error:

```python
Traceback (most recent call last):
  File "test_dynet.py", line 28, in <module>
    pX.set_value([2,0])
  File "_dynet.pyx", line 619, in _dynet.Parameters.set_value
    cpdef set_value(self, arr):
  File "_dynet.pyx", line 623, in _dynet.Parameters.set_value
    shape = arr.shape
AttributeError: 'list' object has no attribute 'shape'
```

It is really annoying.

PR #1006, opened 2017-10-21
fix parameter set_value

chunyang-wen:
Support both np.ndarray and python list used in the `parameter.set_value`

```python
import dynet

m = dynet.ParameterCollection()
pX = m.add_parameters((2,))
pY = m.add_parameters((2,))

pX.set_value([2,1])
p = dynet.parameter(pX)
print p.value()

pY.set_value(np.array([2,1]))
q = dynet.parameter(pY)
print q.value()
```

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -621,7 +621,7 @@ cdef class Parameters: # {{{
 
         """
         if not isinstance(arr, np.ndarray):
-            arr = np.array(arr)
+            arr = np.asarray(arr)
         shape = arr.shape
         if self.shape() != shape:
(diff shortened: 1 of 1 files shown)


chunyang-wen:
#1002

pmichel31415:
I'm not sure which is faster but `np.asarray` should do the same

chunyang-wen:
@pmichel31415 I have searched the doc. `np.array` default will make a copy, so under circumstance it will be slower. Replace it with `np.asarray'.

chunyang-wen:
@neubig Ping.

neubig:
Thanks!

PR #1007, opened 2017-10-22
Fix creation of wheels on Windows

danielhers:
- Removed `repository` from PyPI settings as it is no longer supported.
- Add `dynet.lib` as data file to wheel.

Wheels are now created successfully: https://pypi.python.org/pypi?:action=display&name=dyNET&version=2.0.1.post8


neubig:
Thanks!

danielhers:
@neubig By the way, the automatic creation of wheels upon tag creation in clab/dynet depends on the tag name being the numeric version. Is that OK, or are you going to add a `v` prefix, for example?

PR #1008, opened 2017-10-22
Fix scala bindings

shuheik:
Scala bindings can't be built with the latest master and this PR fixes it.
I've also changed the Scala version from 2.11.8 to 2.11.11.


neubig:
Thanks!

Issue #1010, opened 2017-10-24
Loading model with more than one parameter on Windows hangs forever


danielhers:
Using Python 3.6 on Windows 7 64bit, I'm saving a very small model but loading it never ends (I waited half an hour before killing it):

    import sys
    import dynet as dy
    m = dy.ParameterCollection()
    p = [m.add_parameters(1), m.add_parameters(1)]
    print("Saving model...", file=sys.stderr)
    dy.save("test_model", p)
    print("Loading model...", file=sys.stderr)
    dy.load("test_model", m)
    print("Done", file=sys.stderr)

If the model has just one parameter, it works, but adding the second parameter makes it hang.

danielhers:
@mattr1 maybe it's related to the infinite loop you mentioned in 2ce679b0b328b8b5e3c8c700f00fb804728a0701...

chunyang-wen:
Hi, would please test master branch of this repo?

https://github.com/chunyang-wen/dynet

I think the problem is caused by "\n" -- "\r\n" under windows. I have no windows machine.

mattr1:
I can take a look at this. I also thought the problem was \r\n vs. \n, but stripping the \r's from the test input file didn't fix the problem. I'll test your changes and see if that helps, but it still leaves a bit of a problem in that when you move files around, sometimes \n's are replaced with \r\n. For instance, the model files checked in to git might get \r's inserted in them if git decided it was a text file rather than binary file (not sure how git decides what the file type is). 

I think the problem is caused by the reader seeking forward in the file some set number of bytes (specified on a given line). I don't see any reason for this rather than just reading the line, except maybe an argument of efficiency. I think it's a little dangerous to have the model file look pretty much like a nice text file, with newlines and readable numbers, but then have byte seeking offsets integrated in it. If we change this to advance by reading a line then people could actually edit the file if they wanted.

chunyang-wen:
I just make the file as binary format when reading and writing. Then \n \r\n won't happen. And it is also readable I think.

chunyang-wen:
The reason to seek around is that we can load part of a saved model. Bytes shift is needed when skipping otherwise we have to read everything. I think current loader is a little inefficient because frequent creation and destruction of file stream. I will try to figure out if there exists any better way.

mattr1:
Okay, your changes worked in Windows, both with the test-io unit test, and the python script above. Can you do a pull request to merge them in?

[...]

PR #1013, opened 2017-10-24
Added C++ unit tests to Appveyor

mattr1:
As well as having it compile the C++ examples. The test-io fails (gets in an infinite loop), so temporarily is not part of the testing. This issue is captured here https://github.com/clab/dynet/issues/1010.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -53,6 +53,9 @@ after_build:
   - python ../../setup.py EIGEN3_INCLUDE_DIR=C:/projects/eigen build --build-dir=.. --skip-build bdist_wheel install
 
 test_script:
+  - cmd: cd build
+  - cmd: ctest -C %configuration% --output-on-failure --timeout 300
+  - cmd: cd ..
   - python -m unittest discover %APPVEYOR_BUILD_FOLDER%\tests\python -v
 
 deploy_script:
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -53,6 +53,12 @@ after_build:
   - python ../../setup.py EIGEN3_INCLUDE_DIR=C:/projects/eigen build --build-dir=.. --skip-build bdist_wheel install
 
 test_script:
+  # Cmake (C++) unit tests
+  - cmd: set PATH=%BOOST_LIBRARYDIR%;%PATH%
+  - cmd: cd %APPVEYOR_BUILD_FOLDER%\build
+  - cmd: ctest -C %configuration% --output-on-failure --timeout 300
+  # Python unit tests
+  - cmd: cd %APPVEYOR_BUILD_FOLDER%\build\python
```
danielhers: I think this last `cd` isn't really necessary

mattr1: I thought that originally too, but the python testing that runs on the following line seems to need to be run from that directory. Here is the AppVeyor run without the cd command: https://ci.appveyor.com/project/danielh/dynet-c3iuq/build/1.0.67 (commit https://github.com/clab/dynet/pull/1013/commits/2ce679b0b328b8b5e3c8c700f00fb804728a0701). I wasn't sure what causes the errors, I'm not familiar enough with python unit testing.

danielhers: Oh, right, it was picking up the `dynet` directory inside `build` because that was the current working directory. So I guess it is necessary to `cd` to a directory without a `dynet` in it :)

mattr1: Oh, funny. You're right, that must be it.

neubig:
Thanks!

PR #1014, opened 2017-10-25
FIX: TextFileSaver/Loader under windows

chunyang-wen:
Due to automatic conversion between "\n" and "\r\n", `load_parm` of `TextFileLoader` is not working due to the `byte_count` we need when skipping bytes unrelated.

This patch changes the format from default to `std::ios_base::binary`. It will prevent the conversion and the number of bytes we calculate will be correct between systems.


chunyang-wen:
#1010

chunyang-wen:
@mattr1

neubig:
Thanks!

Issue #885, opened 2017-09-11
Fix warning for compiling the generated `_dynet.cpp` file.


xunzhang:
After https://github.com/clab/dynet/pull/884 and https://github.com/clab/dynet/pull/883 have been merged, the only compiling warnings left are from the generated `_dynet.cpp` file. Maybe let's also fix them in some way later.

danielhers:
Looks like this is fixed.

PR #1015, opened 2017-10-25
FIX: compiler warnings

chunyang-wen:
Hopefully, this patch can remove remaining compiler warnings.
#885 

Remaining warning is in `linear_regression` test case. I have searched the doc. It is a known issue.


chunyang-wen:
@xunzhang

Issue #1025, opened 2017-10-26
No bounds checking in pickneglogsoftmax (Python, CPU)


bdusell:
Hello,

When `pickneglogsoftmax` or `pickneglogsoftmax_batch` is called with an invalid index, rather than throwing an error, it returns a garbage value. To illustrate:

```
>>> import dynet
[dynet] random seed: 2458451077
[dynet] allocating memory: 512MB
[dynet] memory allocation done.
>>> v = dynet.inputVector([1, 2, 3])
>>> v[0].value()
1.0
>>> v[100].value()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "_dynet.pyx", line 1666, in _dynet.Expression.__getitem__
IndexError: Index too large: 100 > 2
>>> dynet.pickneglogsoftmax(v, 0).value()
2.4076058864593506
>>> dynet.pickneglogsoftmax(v, 100).value()
3.4076058864593506
```

This can lead to serious bugs which are hard to detect. Sometimes the value returned is negative -- I only noticed my bug when my loss function was somehow dropping _below_ 0.

I'm running on CPU (I haven't tried reproducing this on GPU yet).

PR #1026, opened 2017-10-26
Boundary check in PickNegLogSoftmax

neubig:
Fixes https://github.com/clab/dynet/issues/1025


chunyang-wen:
I think it is user's responsibility to do a lot of checks. If we add too many checks, performance will surely be impacted. And I think there should be macros to open or close these kinds of check for performance improvement.

neubig:
There already is an option to turn off boundary checks. Add `-DDYNET_SKIP_ARG_CHECK` to your command line when running `cmake` in the manual install.

chunyang-wen:
Yes! That's what I want.

Issue #940, opened 2017-09-28
confusing APIs: sum_elems vs esum


pcyin:
I found the names of some APIs are confusing. For example, `sum_elems` vs `esum`. Any plan on revising them?

neubig:
`esum` is just because `sum` in Python is a reserved word I think. It could be `sum_expressions` or something, but that's awfully long to write.

In the long term, it'd be nice to rethink all of the API naming and make it more consistent, perhaps matching numpy or one of the recent attempts to standardize naming for neural network functions (with the old API names aliased to the new names).

pmichel31415:
Actually if we're going to follow the numpy conventions (which I think would be a very good idea) we can just have a `sum` function that behaves differently given the parameter:

* `dy.sum(list)` <=> `dy.sum_elems`
* `dy.sum(expression)` <=> `dy.sum_elems`
* `dy.sum(expression, dim=d)` <=> `dy.sum_dim`

Same for `mean_*` and `average`...

neubig:
@pmichel31415 I think we'll try to do this.

yoavg:
The original reason for 'esum' was to allow 'from dynet import *' without overriding the builtin sum function. I'm fine with changing it to sum. Re behaving differently for different arguments, this can potentially have some speed overhead (though, granted, sum is not that commonly used).

neubig:
@yoavg Speed overhead is a concern, as I'm essentially saying that we should do this for all componentwise operations. We should definitely profile before committing to this design.

PR #1028, opened 2017-10-27
Softmax across rows

msperber:
- Introduce a softmax operator across rows, rather than columns as in the standard softmax.
- Both the softmax() and softmax_rows() are improved to use scratch memory rather than auxiliary memory.

Rather than introducing a new operator, I first tried to add a "dimension" parameter to the softmax that defaults to d=1, but that introduced inconsistency because softmax is also supposed to work for a vector input (for which d=1 makes no sense), so I went with the new operator instead.

### DIFF
--- a/dynet/expr.cc
+++ b/dynet/expr.cc
@@ -85,6 +85,7 @@ Expression sparsemax(const Expression& x) { return Expression(x.pg, x.pg->add_fu
 Expression sparsemax_loss(const Expression& x, const vector<unsigned>& target_support) { return Expression(x.pg, x.pg->add_function<SparsemaxLoss>({x.i}, target_support)); }
 Expression sparsemax_loss(const Expression& x, const vector<unsigned>* ptarget_support) { return Expression(x.pg, x.pg->add_function<SparsemaxLoss>({x.i}, ptarget_support)); }
 Expression softmax(const Expression& x) { return Expression(x.pg, x.pg->add_function<Softmax>({x.i})); }
+Expression softmax_rows(const Expression& x) { return Expression(x.pg, x.pg->add_function<SoftmaxRows>({x.i})); }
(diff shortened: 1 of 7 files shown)


neubig:
Recently I've been trying to avoid "_rows" and instead going with "_dim" that allows you to specify an arbitrary dimension. Could you take a look at `logsumexp_dim`? I think it should be pretty simple to add the ability to specify a dimension to the `softmax` (and `logsoftmax`) functions. This has the added advantage of not adding more functions to the API.

msperber:
Sure, so we would basically have something like this?
- softmax_dim(e, d)
- softmax(e) -> softmax_dim(e,0) if e is a vector, or softmax_dim(e,1) if e is a matrix?

neubig:
I was just thinking of something like `softmax(e, d=0)`.

When we do our long-needed renaming of functions, I think we need to separate the componentwise operations (e.g. operations across expressions where each component is subject to that operation), and dimension-wise operations (operations that reduce or calculate along dimensions). `logsumexp` was already defined as a componentwise operation, so I needed to make `logsumexp_dim`. `softmax` on the other hand, is already a dimensionwise operation, and it doesn't have a componentwise version, so we can just add specifying the dimension to softmax.

neubig:
(see here: https://github.com/clab/dynet/issues/940). This will be a lot of work to re-design though, so this pull request, let's just make a minimal modification while being consistent with the current functions.

msperber:
Yes, that makes sense to me. But doing this for softmax would break existing code that uses softmax and force people to change their call from softmax(e) to softmax(e,d=1). If that's ok, I will push a commit.

neubig:
No, softmax is currently done over columns, which means we need to reduce over rows. `d` should be the dimension over which we are calculating the denominator.

msperber:
Ok got it, this should be ready then.

neubig:
Thanks!

PR #1029, opened 2017-10-27
Revert eigen stream

### LINKED ISSUE #974: Training with autobatching seems to be slower than before
I suspect that training with autobatching got a bit slower after some recent commit. 
I compared some examples (train_rnn-autobatch in examples, bilstm-tagger-withchar-bulk, treenn-bulk and rnnlm-batch in neulab/dynet-benchmark) between the latest (a69c515) and an old (a3782a1) commit, and found that old ones were 1.4 ~ 1.8 times faster in all of them. 
I chose a3782a1 for comparison because I remembered it was faster, and I don't think it is _the_ last commit without this speed issue.

neubig:
This is an attempt to fix #974, which was probably introduced by `d3ac2aefdb460939f41401006d56fff19d9667d4`. I was able to reproduce the problem with the following results:

`./bilstm-tagger-withchar-bulk --dynet-autobatch 1 20 128 50 32 0 32 2 600`
* Latest master: word_per_sec=4103.38
* a3782a1: word_per_sec=8279.58

I've started working by reverting the commit found by @shuheik and fixing some compile errors, but am getting segmentation faults:

* Latest master after `git revert d3ac2aefdb460939f41401006d56fff19d9667d4`: segmentation fault

I'm out of time now but will resume later. @QiJune or @xunzhang if you could help debug it'd be appreciated.


xunzhang:
I will spend some time looking at this soon.

neubig:
OK, it looks like the segmentation fault might have just been a mistake in my environment where I forgot to rebuild from scratch. I've switched to a different machine and here are the results:

* Before Eigen stream modification: word_per_sec=16736.5
* After Eigen stream modification: word_per_sec=7432.61
* Latest master: word_per_sec=4292.2
* This PR: word_per_sec=11800.2

So basically, this has fixed a large part of the problem, but there's also probably another bad regression somewhere. I'll continue to look for this in a second PR.

PR #1030, opened 2017-10-29
Adding the 'SILU / Swish' activation function

jkkummerfeld:
Adding a new activation function, proposed by Ramachandran, Zoph, and Le [https://arxiv.org/pdf/1710.05941.pdf](https://arxiv.org/pdf/1710.05941.pdf)

swish(x) = x . sigmoid(beta * x)

Given how simple this is, it could be implemented by users manually, but it is certainly more convenient to have it as a built-in dynet function.

Note:
- I only tried this on top of 2.0.1 because I find that the current version of master doesn't work on my machine.
- I wasn't sure how to run the tests (I couldn't see where the target for `make test` was)


stefan-it:
I think we should consider using the name "SILU" instead of Swish, because the Google paper (version 1) missed prior that proposed the same activation function, which can be found in [this](https://openreview.net/pdf?id=Bk0MRI5lg) paper. 

See full discusion [here](https://www.reddit.com/r/MachineLearning/comments/773epu/r_swish_a_selfgated_activation_function_google/).

jkkummerfeld:
Thanks for pointing to that discussion. I agree re: naming, and have updated the branch. In the documentation I did also write 'Swish' to make this easy to find by either name.

jkkummerfeld:
Hm, I'm not sure why github is treating that file as conflicting.

neubig:
Cool, looks good thanks! Looking at the Hendricks and Gimpel paper, it looks like it might also be nice to have GELUs implemented as well.

PR #1033, opened 2017-10-31
Leave out intermediate top directory from Eigen zip in Python setup

danielhers:
When extracting Eigen from zip, avoid the `eigen-eigen-...` top directory, and just put everything under `eigen/`. Otherwise running `setup.py` again skips downloading Eigen (because it already exists) but fails in building DyNet (because its contents are not what we expect, but that intermediate top directory).


neubig:
Thanks!

PR #1036, opened 2017-11-03
--dynet-autobatch parameter update

yanaiela:
seems like the auto-batch option is expecting a boolean parameter (0/1)
from here: https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L152
also because it throws a segmentation fault when passing something different.

### DIFF
--- a/doc/source/commandline.rst
+++ b/doc/source/commandline.rst
@@ -22,7 +22,7 @@ options.
    every update. This is similar to L2 regularization, but different in a
    couple ways, which are noted in detail in the "Unorthodox Design"
    section.
--  ``--dynet-autobatch NUMBER``: Turns on DyNet's automatic operation
+-  ``--dynet-autobatch BOOL``: Turns on DyNet's automatic operation (BOOL should be 0/1)
    batching capability. This makes it possible to speed up computation with
    a minimum of work. More information about this functionality can be found
    `here <http://dynet.readthedocs.io/en/latest/minibatch.html>`_.


xunzhang:
I prefer keeping `NUMBER` instead of `BOOL`. The reason is that the command line document is both for c++ and python ends, so `BOOL` might let user input `true/false` value which is not supported in c++ end(it's `INT` in c++). I think non-zero values should be treated as true in python.

yanaiela:
so maybe should clarify that it's supposed to be 0/1 (i got confused at first and tried to give it the size of the mini-batch - and got that exception)

neubig:
Also, "2" and "3" are actually supported, although not recommended. "2" is the ByDepth batching strategy in the autobatching paper, and "3" is a pseudo-strategy that tries "0", "1", and "2" for one training example, then continues with the best one. We should definitely improve the documentation, and also prevent a segfault on bad values though.

chunyang-wen:
@neubig I am interested in the paper you mentioned. Would you please provide more information?

xunzhang:
https://arxiv.org/abs/1705.07860

neubig:
Hi @chunyang-wen , thanks for committing this! Unfortunately when cleaning the repository history of a large file to reduce download size it messed up this commit. Would you mind pulling the new master, then cherry-picking the relevant commits, and re-submitting a pull request? If you prefer, I could also do this myself, but it would probably result in github counting it as my PR instead of yours.

chunyang-wen:
@neubig Do you mean this specific commit? If so, it is not my work.

neubig:
Sorry, yes. This should have been addressed to @yanaiela

neubig:
This documentation update needs to be re-based. I'm going to close this PR, but welcome another PR that performs this update on the most recent branch.

Issue #1034, opened 2017-11-02
pip install dynet not working on Mac


chunyang-wen:
Mac OS:10.12.5 (16F73)

```python
pip install dynet
```

```python
>>> import dynet
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/dynet.py", line 15, in <module>
    from _dynet import *
ImportError: dlopen(/usr/local/lib/python2.7/site-packages/_dynet.so, 2): Library not loaded: /private/var/folders/xf/qm8tkgkx5yjfgrhrmqxvsknr0000gn/T/pip-build-snn3Sx/dynet/build/py2.7-64bit/dynet/libdynet.dylib
  Referenced from: /usr/local/lib/python2.7/site-packages/_dynet.so
  Reason: image not found
```

neubig:
I was able to confirm this problem, @danielhers any ideas?

danielhers:
It's installing from source - since I wasn't able to build macOS wheels with the library in them yet, I didn't upload them. A workaround would be something like

  pip install --build-dir=~/dynet_build dynet
  export DYLD_LIBRARY_PATH=~/dynet_build/dynet/build/dynet/:$DYLD_LIBRARY_PATH

danielhers:
Please try again now. I think `2.0.1.post9` wheels should include `libdynet.dylib`.

chunyang-wen:
Hmm. It still does not work even through the file exists.

danielhers:
Right. Looks like the `rpath` wasn't set to find it though. Added `@loader_path/` to the `rpath` in  `2.0.1.post10` but I'm still not sure it worked.

chunyang-wen:
```shell
export DYLD_LIBRARY_PATH=/install/path/of/dynet/:$DYLD_LIBRARY_PATH
```
will solve the problem

danielhers:
Fixed in `2.0.1.post13` and verified. `DYLD_LIBRARY_PATH` is no longer required.

PR #1043, opened 2017-11-05
Deploy macOS wheels

### LINKED ISSUE #870: Precompiled wheels for macOS with linked dynamic library
Wheels can be built for macOS (#841), and the `delocate-wheel` tool is supposed to take `libdynet.dylib` and relink it into the wheel so that the `DYLD_LIBRARY_PATH` workaround (#146) doesn't have to be used. However, when using `delocate-wheel`, even though `libdynet.dylib` is found, it is not relinked into the wheel. The cause for this problem might be that the DyNet Python files are not in a package, but just at the top level of the installation (matthew-brett/delocate#15).

danielhers:
- Fixes #870 and #1034: create macOS wheel on Travis CI with the DyNet shared library and deploy to PyPI on tags.
- Unrelated fix for Travis CI installation retrying the `brew install`.
- Unrelated fix for AppVeyor failures.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,7 @@ environment:
     BOOST_ROOT: C:\Libraries\boost_1_59_0
     BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib64-msvc-14.0
     MSVC: 1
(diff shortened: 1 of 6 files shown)


neubig:
Thanks!

PR #1044, opened 2017-11-05
Constrained Softmax

andre-martins:
Implementation of a ConstrainedSoftmax node as described in this paper:

André F. T. Martins and Julia Kreutzer.
"Learning What’s Easy: Fully Differentiable Neural Easy-First Taggers."
Conference on Empirical Methods in Natural Language Processing (EMNLP'17), Copenhagen, Denmark, September 2017.
https://aclanthology.info/pdf/D/D17/D17-1036.pdf

The function implemented in this node is similar to Softmax, but it allows upper bound constraints on the resulting probabilities. It's differentiable both with respect to both the scores and the constraints. Currently CPU-only.


neubig:
Hi @andre-martins , thanks for committing this! Unfortunately when cleaning the repository history of a large file to reduce download size it messed up this commit. Would you mind pulling the new master, then cherry-picking the relevant commits, and re-submitting a pull request? If you prefer, I could also do this myself, but it would probably result in github counting it as my PR instead of yours.

andre-martins:
Sure, I can do that in the next few days.

2017-11-25 13:08 GMT+00:00 Graham Neubig <notifications@github.com>:

andre-martins:
Hi @neubig, I just cleaned this pull request and it's running tests now.

neubig:
Thanks @andre-martins ! But actually, if you look at the PR page there are still a bunch of leftover commits before your first one at e37caa8 (this is due to my need to clean the repository of a big file, which rewrote the entire history).

If you want to fix it yourself, I think the easiest way is to clone a new version of master, then use `git cherry-pick` to merge only your commits into the clean version of master. I can also help if you'd like, but it might end up erasing the history of your commits.

andre-martins:
Got it. Actually I think it's easier to add the file changes manually instead of doing `git cherry-pick`. I'm going to close this PR and open a new one, where I'll do this from the Unbabel fork.

PR #1048, opened 2017-11-06
Fix install name in macOS wheels with install_name_tool

danielhers:
[Resulting wheels](https://pypi.python.org/pypi/dyNET/2.0.1.post13) verified to work without `DYLD_LIBRARY_PATH`. Just do `pip install dynet`.
Removed `delocate` because it doesn't work in our case (matthew-brett/delocate#22).

### DIFF
--- a/.travis/build_macos_wheel.sh
+++ b/.travis/build_macos_wheel.sh
@@ -3,10 +3,6 @@ set -xe
 
 cd "$TRAVIS_BUILD_DIR"
 source activate "$PYVER"
-pip install pypandoc delocate
+pip install pypandoc
 python setup.py bdist_wheel
-mkdir -p dist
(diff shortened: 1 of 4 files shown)


neubig:
Sorry, I managed to break the git history when deleting a large file. Would you mind re-checking out master, copying the relevant files, and re-submitting a new PR? Sorry about the inconvenience.

danielhers:
If you wanted to undo that, you could just [`git reset --hard c57f9da5778b6921a08f544f70d550d3977c9b73`](https://github.com/clab/dynet/commit/c57f9da5778b6921a08f544f70d550d3977c9b73).

But no problem, I updated my master by doing `git reset --hard` to the current clab/master and then doing `git cherry-pick` to get my commits again.

Issue #758, opened 2017-08-03
Advanced Indexing


shuoyangd:
Hi All,

Per my question with @neubig earlier today, it would be really great if we can have support for [advanced indexing](https://docs.scipy.org/doc/numpy-1.8.1/reference/arrays.indexing.html#advanced-indexing) feature in DyNet.

PyTorch also started [this issue](https://github.com/pytorch/pytorch/issues/1080) a while ago and seems to be officially supporting it soon in v0.2.0. Although this might look much too fancy for someone, I do know some people (like me) are using this prevalently for numpy tensors and back in the era of theano. Supporting such features would create a more seamless experience when these folks are migrating to DyNet.

I hope someone who is familiar with the codebase can give some instructions towards how this could be integrated so people can work to add support for this feature.

Thanks!

pmichel31415:
I agree this would be a cool feature to have.

There are two places to modify to make this work:

1. the `__getitem__` method in the [cython source](https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L1424-L1485). This should be pretty easy to do: just interpret the slice argument and call the according dynet `pick*` node.

2. For more efficiency we might need to create some more picking nodes (e.g. pick from multiple dimensions, pick with step size, etc) in [there](https://github.com/clab/dynet/blob/master/dynet/nodes-select.h) and corresponding expressions.

In theory everything couls be implemented with 1., so if anyone wants to do it i suggest starting there. Then later on we can add relevant nodes in the C++ code to make it more efficient.

PR #1070, opened 2017-11-14
implement strided_select

msperber:
This implements a striding operator, including specification of offset / range. It can be used as dy.strided_select(strides, offsets, to) and works over multiple dimensions (including batch dimension) at once. I tried to follow numpy selector conventions, although this is not as powerful, e.g. no negative indices or strides are supported. It could be extended toward #758 in the future.


dynet/nodes-select.cc
```diff
@@ -332,4 +332,132 @@ void PickBatchElements::backward_dev_impl(const MyDevice & dev,
...
+  return ret;
+}
+
+#endif
+
+template<class MyDevice>
+void StridedSelect::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
+  Eigen::array<int, 5> offsets = {0, 0, 0, 0, 0};
+  Eigen::array<int, 5> extents = {(int)(xs[0]->d[0]), (int)(xs[0]->d.nd < 2 ? 1 : xs[0]->d[1]), (int)(xs[0]->d.nd < 3 ? 1 : xs[0]->d[2]), (int)(xs[0]->d.nd < 4 ? 1 : xs[0]->d[3]), (int)(xs[0]->d.bd)};
+  Eigen::array<int, 5> strides_arr({1,1,1,1,1});
```
neubig: I think this doesn't compile on windows. It should be:

    Eigen::array<int, 5> strides_arr = {1,1,1,1,1};

neubig:
Thanks! All the ternaries made the code a bit difficult to read, but it looks like the functionality is fine.

Issue #1072, opened 2017-11-16
possible bug of select_cols?


duyvuleo:
Hi,

I've had a weird test result on select_cols, e.g., 

```
import dynet as dy

mod = dy.Model()

pv = mod.add_lookup_parameters((20, 5))

ix_batch = dy.lookup_batch(pv, [1, 2 ,3])
iy_batch = dy.lookup_batch(pv, [4, 5, 6])
ixy_batch = dy.concatenate_cols([ix_batch, iy_batch])
print "ixy_batch.dim(): " + ixy_batch.dim()

ixy_col = dy.select_cols(ixy_batch, [0])
print "ixy_col.dim(): " + str(ixy_col.dim()

ixy_row = dy.select_rows(ixy_batch, [0])
print "ixy_row.dim(): " +str(ixy_row.dim())
```

Results:
ixy_batch.dim(): ((5, 2), 3)
ixy_col.dim(): ((5, 1), **1**)
ixy_row.dim(): ((1, 2), 3)

ixy_col.dim() should retain the batch size, right? perhaps a bug in select_cols? 

Thanks!

PR #1073, opened 2017-11-17
bug in select_cols

duyvuleo:
This PR is to fix a bug in dim for select_cols node.


duyvuleo:
refers to #1072

neubig:
Thanks!

PR #1075, opened 2017-11-17
Update Scala bindings

shuheik:
Add some Expression-related methods and Trainer.restart to Scala bindings.
Also add UnsignedVectorVector, as it's required as the argument for batched hinge_dim.


neubig:
Thanks!

PR #1076, opened 2017-11-18
After deployment, test pip installation from PyPI works

danielhers:
Deployment to PyPI is done on tag creation. Adding a build stage (new Travis feature) to test the deployed wheels work in a fresh environment, to verify there is no dependence on any remaining build outputs.


neubig:
Thanks!

Issue #1080, opened 2017-11-21
Use sdl linking for mkl


akoehn:
I noticed that dynet links directly against `mkl_intel_lp64 mkl_intel_thread mkl_core`, which somehow doesn't work with my use case (i.e. using the scala bindings and mkl).

I get the same error as described [in this intel forum post](https://software.intel.com/en-us/forums/intel-math-kernel-library/topic/733207) and the answer is to instead link against `mkl_rt`.

Linking against `mkl_rt` uses the [single dynamic library approach](https://software.intel.com/en-us/articles/a-new-linking-model-single-dynamic-library-mkl_rt-since-intel-mkl-103/), in which this library decides at runtime which "backend library" to use.

Changing this linking behavior can be achieved by switching the libraries in `CMakeLists.txt` lines 45 and 47.  I did this locally and it solves my issue.  However, I'm not sure enough of other implications to create a pull request directly.

neubig:
Could you create a pull request and we'll try it out ourselves? Thanks!

akoehn:
Done, see #1081

neubig:
Fixed by #1081

PR #1081, opened 2017-11-21
MKL: Link against mkl_rt instead of all libraries directly

akoehn:
Use the "single dynamic library" linking model of MKL, described here:
https://software.intel.com/en-us/articles/a-new-linking-model-single-dynamic-library-mkl_rt-since-intel-mkl-103/

See also #1080


neubig:
Thanks! I confirmed this works.

mattr1:
I can confirm this works on MSVC/Windows as well. I'm going to remove the MSVC conditional since it's now superfluous (yay!).

PR #1086, opened 2017-11-22
scala bindings: allow to set scala version using cmake

akoehn:
Scala 2.11 and 2.12 are binary incompatible; using this patch it is possible
to switch between scala versions using cmake variables instead of
editing the build.sbt file.

In addition, sbt settingKeys are used instead of plain values to
conform to the weird sbt way.

### DIFF
--- a/contrib/swig/CMakeLists.txt
+++ b/contrib/swig/CMakeLists.txt
@@ -14,6 +14,12 @@ include_directories("../../dynet")
 # Set java output dir
 set(CMAKE_SWIG_OUTDIR "${CMAKE_CURRENT_BINARY_DIR}/java/src/edu/cmu/dynet")
 
+# the exact scala version used for compilation
+set(SCALA_VERSION "2.11.11" CACHE STRING "Scala version to compile the swig wrapper for")
+
+# the binary compatibility, e.g. 2.11 or 2.12
+string(REGEX MATCH "[0-9]+\\.[0-9]+" SCALA_BIN_VERSION "${SCALA_VERSION}")
+
 # Set java package (+cuda flag, if appropriate)
 if(WITH_CUDA_BACKEND)
   set(CMAKE_SWIG_FLAGS -package edu.cmu.dynet.internal -DSWIG_USE_CUDA)
@@ -131,7 +137,7 @@ option(INCLUDE_SCALA "INCLUDE_SCALA" ON)
 if(INCLUDE_SCALA)
 
   # make doesn't know about the dynet_scala jar, so it doesn't clean it unless we tell it about it
-  set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES "dynet_swigJNI_scala.jar")
+  set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES "dynet_swigJNI_scala_${SCALA_BIN_VERSION}.jar")
 
   # Find sbt
   find_program(SBT sbt)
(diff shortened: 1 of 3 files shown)


neubig:
@shuheik do you have any comments about this? I'm not familiar with Scala, but if it looks OK we can go ahead and commit.

shuheik:
I tried this locally and I think two lines in CMakeList.txt need to be changed,

1.
`    "${CMAKE_SWIG_OUTDIR}/SWIGTYPE_p_std__vectorT_std__vectorT_unsigned_int_t_t.java"`
in https://github.com/akoehn/dynet/blob/ae13e73fb7fa413f9d3359630f803e36b94435b4/contrib/swig/CMakeLists.txt#L115 should be removed as in #1085.

2.
Shouldn't
`    COMMAND ${SBT} assembly -Dbuildpath=${CMAKE_CURRENT_BINARY_DIR} -Dscalaversion=${CMAKE_CURRENT_BINARY_DIR}`
in https://github.com/akoehn/dynet/blob/ae13e73fb7fa413f9d3359630f803e36b94435b4/contrib/swig/CMakeLists.txt#L150 be
`    COMMAND ${SBT} assembly -Dbuildpath=${CMAKE_CURRENT_BINARY_DIR} -Dscalaversion=${SCALA_VERSION}`
?

Actually I'm editing scala version in build.sbt manually now, and while I think it's good to be able to set it through cmake variables, there are some tradeoffs between the two approaches.
Especially, when I change the version in build.sbt manually and rerun make, only Scala-dependent tasks are triggered and swig-related files are automatically replaced. 
If I change the version through cmake variables and rerun make, some C++ codes are also re-compiled and I have to delete swig-related files generated under the different version to avoid conflicts.
Scala versions are not likely to be modified so often and maybe this overhead doesn't matter so much, though.

I would like to hear what other people think, especially @joelgrus.

akoehn:
Regarding 1., #1085 is already merged, so why also delete it in this patch? It would at most result in merge problems.

2.: You're right, fixed.  I have no clue how that got there.

Regarding the recompilation:  I think it is undesirable to always have to build from a dirty state, i.e. having to modify files checked into git.  I think most use cases for the bindings deal with exactly one scala version, so that the recompilation overhead does not manifest itself.

Can you tell me which swig-related files you have to manually delete to avoid conflicts?

shuheik:
Seems like deleting contrib/swig/dynet_swigJNI_scala_2.X.jar before re-make under different Scala version is enough.
If it's still there, I see errors like this.

 ```
[info] Including from cache: dynet_swigJNI_scala_2.12.jar
 ...
 java.lang.RuntimeException: deduplicate: different file contents found in the following:
 edu/cmu/dynet/AdadeltaTrainer$.class
 /work/shuhei-k/tools/tmp/dynet/build/contrib/swig/dynet_swigJNI_scala_2.12.jar:edu/cmu/dynet/AdadeltaTrainer$.class
 ...
 [error] (*:assembly) deduplicate: different file contents found in the following:
 [error] edu/cmu/dynet/AdadeltaTrainer$.class
 [error] /work/shuhei-k/tools/tmp/dynet/build/contrib/swig/dynet_swigJNI_scala_2.12.jar:edu/cmu/dynet/AdadeltaTrainer$.class
 ...
```

But I think it's acceptable to delete build/contrib directory before rebuilding the bindings.

>Regarding the recompilation: I think it is undesirable to always have to build from a dirty state, i.e. having to modify files checked into git. I think most use cases for the bindings deal with exactly one scala version, so that the recompilation overhead does not manifest itself.

I agree. My current way of dealing with Scala versions is ad-hoc and I'll have little trouble getting used to cmake-way.
@neubig I think this change is good to go.

Issue #498, opened 2017-04-29
Feature Request: Better memory profiling capability


neubig:
It would be nice if DyNet gave you some way to easily profile which nodes were using lots of memory in the computation graph.

neubig:
Fixed by https://github.com/clab/dynet/pull/1088

PR #1088, opened 2017-11-22
profile fwd/bwd runtimes and memory

msperber:
- rename autobatch_debug to profiling
- if set:
  - print fwd / bwd runtimes upon program termination, sorted by runtime
  - using print_graphviz now prints node and auxiliary memory consumption
  - if >=2: print nodes sorted by memory consumption after the graph

related to #498 and #470


neubig:
Awesome, thanks!

Issue #1059, opened 2017-11-09
Random variate generator improvements


redpony:
There are a few problems with the random variate generators.
1) They don't autobatch, although they should.
2) If you want random variates on the GPU, you have to create them on the CPU and copy them to the GPU.
3) The GaussianNoise function (which adds noise to an expression) also creates the random variates in a temporary buffer on the CPU, then copies it to the target device, then executes addition. It also does not autobatch.
4) Dropout doesn't autobatch, it generates on the CPU and then copies the dropout mask (which has to live in auxiliary memory) to the appropriate device.

Fixing autobatching should be straightforward (although it is possible that the autobatching code will have to be adapted to deal correctly with artiy-0 functions, but this should be trivial). Dealing with random variate generation in a device-appropriate way will be more effort. First, the RNG state needs to be initialized in a sensible way on each device. Second, there may not be existing kernels for generating noise and adding it to a vector (needed only for GaussianNoise) or for generating all the random variate types DyNet offers. However, most of these kernels should be straightforward to implement. Finally, there will need to be an API change to pass in the device you want to generate the RVs on.

Issue #438, opened 2017-04-12
TensorTools Randomize Functions could be CUDAfied


neubig:
Currently generating random numbers is done on the CPU and these random numbers are transferred to the GPU. It would be nice if we could generate random numbers directly on the GPU as it would presumably be much faster. (The GPU random seed would also need to be set with dynet-seed to ensure consistency across runs.)

neubig:
This can likely be done through Eigen tensor's random number generation capability. See `eigen/unsupported/test/cxx11_tensor_random.cc` and `eigen/unsupported/test/cxx11_tensor_random_cuda.cu` for details.

neubig:
Done by #1154.

PR #1094, opened 2017-11-24
[WIP] Use tensor random generator

### LINKED ISSUE #542: Dropout speed issues
Hi there!

I have been experiencing some issues using dropout. The thing is that it makes my system many times slower to train. Before dropout it took ~6h in GPU to run a full epoch. Now, it takes ~24h. That's a 4 fold increase. The only change in my code was do use dropout in the hidden layers of the net. Also, GPU usage is now way down, and a lot of time it actually reaches 0, which didn't happen before.

Any clues?

msperber:
Use tensor random generator instead of the standard one. This should fix speed issues on GPUs due to creating random numbers on CPU and then transferring to GPU, as was done previously.

I implemented this carefully, but it would be good if someone could review the code, as this involves core functionality and is hard to unit-test.

Some numbers on a CPU machine and a GPU machine:
```
e = dy.random_normal((1000,1000), batch_size=40)
e2 = dy.dropout(e, 0.5)
v = e2.value()

old, CPU: 8.45s
new, CPU: 8.08s
old, GPU: 6.33s
new, GPU: 3.48s
```
I suspect that most of the 3.48s were dynet overhead because in my particular use case dropout went from consuming 90% of computation time to <1%.

fixes #542, #438 
related to #1059 

- [X] Initial implementation using Tensor random generator
- [ ] fix strange behavior on some GPUs

### DIFF
--- a/dynet/nodes-dropout.cc
+++ b/dynet/nodes-dropout.cc
@@ -30,7 +30,10 @@ size_t Dropout::aux_storage_size() const {
 template<class MyDevice>
 void Dropout::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
   Tensor m(dim, (float*)aux_mem, fx.device, DeviceMempool::FXS);
-  TensorTools::randomize_bernoulli(m, (1.f-p), 1.f / (1.f-p));
+  std::uniform_int_distribution<> seed_dist(1, 2147483647);
+  Eigen::internal::UniformRandomGenerator<float> uni_rg(seed_dist(*rndeng));
+  m.tvec().device(*dev.edevice) = m.tvec().random(uni_rg);
+  m.tvec().device(*dev.edevice) = (m.tvec() < m.tvec().constant((1.f-p))).cast<float>() * 1.f / (1.f-p);
   fx.tvec().device(*dev.edevice) = xs[0]->tvec() * m.tvec();
 }
 
@@ -74,7 +77,11 @@ void DropoutDim::forward_dev_impl(const MyDevice & dev, const vector<const Tenso
   Dim mask_dim(dim);
   mask_dim.d[dimension]=1;
   Tensor m(mask_dim, (float*)aux_mem, fx.device, DeviceMempool::FXS);
-  TensorTools::randomize_bernoulli(m, (1.f-p), 1.f / (1.f-p));
+  std::uniform_int_distribution<> seed_dist(1, 2147483647);
+  Eigen::internal::UniformRandomGenerator<float> uni_rg(seed_dist(*rndeng));
+  m.tvec().device(*dev.edevice) = m.tvec().random(uni_rg);
+  m.tvec().device(*dev.edevice) = (m.tvec() < m.tvec().constant((1.f-p))).cast<float>() * 1.f / (1.f-p);
+
   Eigen::array<ptrdiff_t, 4> bcast = {1, 1, 1, 1}; bcast[dimension] = xs[0]->d[dimension];
   fx.tb<3>().device(*dev.edevice) = xs[0]->tb<3>() * m.tb<3>().broadcast(bcast);
 }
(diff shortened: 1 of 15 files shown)


dynet/nodes-dropout.cc
```diff
@@ -30,7 +30,10 @@ size_t Dropout::aux_storage_size() const {
 template<class MyDevice>
 void Dropout::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
   Tensor m(dim, (float*)aux_mem, fx.device, DeviceMempool::FXS);
-  TensorTools::randomize_bernoulli(m, (1.f-p), 1.f / (1.f-p));
+  std::uniform_int_distribution<> seed_dist(1, 2147483647);
+  Eigen::internal::UniformRandomGenerator<float> uni_rg(seed_dist(*rndeng));
+  m.tvec().device(*dev.edevice) = m.tvec().random(uni_rg);
+  m.tvec().device(*dev.edevice) = (m.tvec() < m.tvec().constant((1.f-p))).cast<float>() * 1.f / (1.f-p);
```
neubig: Do we need `*1.f` here?

dynet/nodes-dropout.cc
```diff
@@ -30,7 +30,10 @@ size_t Dropout::aux_storage_size() const {
 template<class MyDevice>
 void Dropout::forward_dev_impl(const MyDevice & dev, const vector<const Tensor*>& xs, Tensor& fx) const {
   Tensor m(dim, (float*)aux_mem, fx.device, DeviceMempool::FXS);
-  TensorTools::randomize_bernoulli(m, (1.f-p), 1.f / (1.f-p));
+  std::uniform_int_distribution<> seed_dist(1, 2147483647);
```
neubig: I'm not a huge fan of this code being copied all the time. If this is something that we'll want to use frequently, perhaps it can be made a global variable in the DyNet namespace or something?

Also, tangentially, there are some random number generation functions in `tensor.h`/`tensor.cc` that it might be reasonable to move to their own `rand.h`/`rand.cc` files, along with this one. What do you think?

neubig: Also, is there a reason why you didn't just implement all of this in TensorTools? It seems like that would work as well. If there's a reason that's no problem though.

msperber: That sounds good, I moved the random helpers to their own files and added a draw_random_seed() helper to avoid code duplication.

msperber: I didn't do that partly because the TensorTools are also used for parameter initialization which I didn't want to touch, and partly because I didn't know how to properly move over the device-related code.

dynet/nodes-dropout.cc
msperber: This is fixed now.

neubig:
OK, I tested this on CPU and GPU with the current master branch and this PR using the attached script. It looks like master CPU/GPU and PR CPU are consistent, but the PR on GPU gives weird results. Could you try running the script and seeing what's up?

[dynet-random.py.zip](https://github.com/clab/dynet/files/1510625/dynet-random.py.zip)

msperber:
Thanks for taking a look, that is strange indeed. random_normal() and random_uniform() look fine to me but bernoulli and dropout do not, which would hint at this line being broken on GPU:

```m.tvec().device(*dev.edevice) = (m.tvec() < m.tvec().constant((1.f-p))).cast<float>() / (1.f-p);```

The result of operator< is a Tensor<bool> which might potentially be a problem?

neubig:
OK, I'm getting very strange behavior even on RandomNormal. It looks like almost exactly half the time it's sampling the exact same value, while half the time it's sampling from the appropriate distribution. This looks like a relatively major bug in the upstream Eigen implementation. I'll try to reproduce it only in Eigen and see if they're interested in fixing it.

neubig:
FYI, we're having some discussions offline about how to implement this directly in cuRAND, which will presumably be more stable than relying on Eigen and potentially easier.

neubig:
Closed in favor of #1154.

PR #1095, opened 2017-11-24
Add forget bias choice for VanillaLSTM model

makyr90:
Based on this paper  http://proceedings.mlr.press/v37/jozefowicz15.pdf by Jozefowicz et al. adding a bias of 1 to the  LSTM's forget gate improves the performance. On the other hand Dozat et al.(https://web.stanford.edu/~tdozat/files/TDozat-CoNLL2017-Presentation.pdf) found that adding the aforementioned bias to the forget gate hurts the performance of their SOTA  POS tagging & Dependency parsing models. 
I have made some minor modifications in order to let the user to decide if forget bias should be used or not.


makyr90:
Ok ,that sounds better. I am going to change it

PR #1097, opened 2017-11-24
Add forget bias parameter(float) for VanillaLSTM model

makyr90:
Based on this paper http://proceedings.mlr.press/v37/jozefowicz15.pdf by Jozefowicz et al. adding a bias of 1 to the LSTM's forget gate improves the performance. On the other hand Dozat et al.(https://web.stanford.edu/~tdozat/files/TDozat-CoNLL2017-Presentation.pdf) found that adding the aforementioned bias to the forget gate hurts the performance of their SOTA POS tagging & Dependency parsing models.
I  made some minor modifications in order to let the user  decide on how much forget bias should be used(default 1.0). For no bias we just set forget_bias = 0.0 on VanilaLSTM initiallization.

### DIFF
--- a/dynet/lstm.cc
+++ b/dynet/lstm.cc
@@ -293,13 +293,13 @@ void CoupledLSTMBuilder::disable_dropout() {
 enum { _X2I, _H2I, _BI, _X2F, _H2F, _BF, _X2O, _H2O, _BO, _X2G, _H2G, _BG };
 enum { LN_GH, LN_BH, LN_GX, LN_BX, LN_GC, LN_BC};
 
-VanillaLSTMBuilder::VanillaLSTMBuilder() : has_initial_state(false), layers(0), input_dim(0), hid(0), dropout_rate_h(0), ln_lstm(false), forget_bias(true), dropout_masks_valid(false) { }
+VanillaLSTMBuilder::VanillaLSTMBuilder() : has_initial_state(false), layers(0), input_dim(0), hid(0), dropout_rate_h(0), ln_lstm(false), forget_bias(1.f), dropout_masks_valid(false) { }
 
 VanillaLSTMBuilder::VanillaLSTMBuilder(unsigned layers,
                                        unsigned input_dim,
                                        unsigned hidden_dim,
                                        ParameterCollection& model,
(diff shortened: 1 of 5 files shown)


dynet/lstm.cc
```diff
@@ -485,8 +485,7 @@ Expression VanillaLSTMBuilder::add_input_impl(int prev, const Expression& x) {
     i_aot = pick_range(tmp, hid * 2, hid * 3);
     i_agt = pick_range(tmp, hid * 3, hid * 4);
     Expression i_it = logistic(i_ait);
-    // TODO(odashi): Should the forget bias be a hyperparameter?
-    Expression i_ft = logistic(i_aft + 1.f);
+    Expression i_ft = logistic(i_aft + forget_bias);
```
neubig: There should be an `if` statement here to avoid the addition if `forget_bias` is zero. This will save a little bit of time and memory.

makyr90: Ok i will fix it asap

neubig: Thanks. Also, you don't need to close the PR every time you make changes. Just push to the branch and your new changes will be reflected here.

dynet/lstm.cc
```diff
@@ -485,8 +485,12 @@ Expression VanillaLSTMBuilder::add_input_impl(int prev, const Expression& x) {
     i_aot = pick_range(tmp, hid * 2, hid * 3);
     i_agt = pick_range(tmp, hid * 3, hid * 4);
     Expression i_it = logistic(i_ait);
-    // TODO(odashi): Should the forget bias be a hyperparameter?
-    Expression i_ft = logistic(i_aft + 1.f);
+    if (forget_bias > 0.0)
```
neubig: We could have a negative forget bias. I think this should just be `!= 0.0`

neubig:
This commit is now auto-formatting a huge amount of code in `_dynet.pyx`. Could you please limit your changes to things that are relevant to the commit?

makyr90:
Oh, i didn't notice it. Probably my text editor is the culprit.
 I will fix it.

neubig:
Thanks! Looks good.

Issue #1101, opened 2017-11-25
cannot install for python 3.6 with pip3


RobinChenRichmond:
When I tried:
sudo pip3 install git+https://github.com/clab/dynet#egg=dynet

Here's the error:
(normal output skipped)............................
INFO:root:Creating build directory /private/tmp/pip-build-h0xm3sj0/dynet/build/py3.6-64bit
    INFO:root:Fetching Eigen...
    error: Could not download Eigen from 'https://bitbucket.org/eigen/eigen/get/699b6595fc47.zip'
    
    ----------------------------------------
Command "/Library/Frameworks/Python.framework/Versions/3.6/bin/python3.6 -u -c "import setuptools, tokenize;__file__='/private/tmp/pip-build-h0xm3sj0/dynet/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-0laqqle5-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /private/tmp/pip-build-h0xm3sj0/dynet/

can someone help and check if it's the problem of setting for my python 3.6?

yanfeiyu:
I have a similar issue with 
pip install git+https://github.com/clab/dynet#egg=dynet

Here is the error:

INFO:root:Copying built extensions...
    running install_lib
    copying build/lib.macosx-10.7-x86_64-3.6/_dynet.cpython-36m-darwin.so -> /Users/Olivia/anaconda3/lib/python3.6/site-packages
    copying build/lib.macosx-10.7-x86_64-3.6/dynet.py -> /Users/Olivia/anaconda3/lib/python3.6/site-packages
    copying build/lib.macosx-10.7-x86_64-3.6/dynet_config.py -> /Users/Olivia/anaconda3/lib/python3.6/site-packages
    copying build/lib.macosx-10.7-x86_64-3.6/dynet_viz.py -> /Users/Olivia/anaconda3/lib/python3.6/site-packages
    byte-compiling /Users/Olivia/anaconda3/lib/python3.6/site-packages/dynet.py to dynet.cpython-36.pyc
    byte-compiling /Users/Olivia/anaconda3/lib/python3.6/site-packages/dynet_config.py to dynet_config.cpython-36.pyc
[...]

PR #1102, opened 2017-11-26
Fix problem reported in #1101: Permission denied

danielhers:
libdynet.dylib needs to be copied to the wheel directory in bdist_wheel, but in install it should be copied to the loader_path instead.


neubig:
Thanks!

PR #1103, opened 2017-11-27
Some partial supports for READ-only inplaced operations (Re-based)

zzsfornlp:
This is a re-based PR of #929 basing on the current master branching, sorry that I have not been following the new commits and changes between that PR and the current point. Thus about this I might need some help for checking if there will be any logical conflicts. Thanks!

Here is some notes about this PR:
1. Since NoBackprop and FlipGradient need special gradient manipulation, they have their own backward memories.
2. For auto-batch mode, currently simply forbid batching for thoes nodes. Firstly the memory-allocation part of auto-batching seems a little complex for me, so I do not change much there in the fear of breaking sth; Secondly, in auto-batching mode, maybe memory is even less important.
3. (TODO) For the memory saving of WRITE nodes, we might need some better mechanism for them.


neubig:
OK, I finally got around to taking a look. First, thanks again for contributing this!

I did a few changes to ensure that things don't break with autobatching. The biggest one is that the behavior of the nodes themselves are not modified, but forward or backward is just skipped in the executor. This is useful because now the executor is solely in charge of handling inplacing properly, and if inplacing is not supported things will work as they did before.

I'm going to do a few more tests and also make sure that the memory profiling code is working properly, then probably merge this.

zzsfornlp:
Thanks a lot for your help!

neubig:
#1156 was merged!

PR #1107, opened 2017-11-28
Improve speed for cwise operations

msperber:
This is a first attempt at fixing speed issues introduced by the broadcasting feature, mainly introducing better checks for whether broadcast operations are needed and bypassing if not.


neubig:
Just to check: is this WIP still?

msperber:
It's ready to merge and should improve things; I marked it as WIP because it would need confirmation of whether this completely resolves the speed issues or not quite.

neubig:
OK, thanks! My interpretation of "WIP" on a PR is "not ready to be merged yet", which is why I hadn't been looking at it.

neubig:
OK, I did a bunch of refactoring, which I think is useful in itself but didn't actually improve the speed that much.

Then finally I realized the problem was simple: csum didn't have an autobatching profile, so it wasn't being autobatched. Now that I fixed that we're back to the original speed!

@msperber if you could take a quick look that'd be nice, and if it looks OK I'll merge.

msperber:
Yes, looks fine!

redpony:
This most recent change seems to have triggered a major performance
regression on CPU for autobatching a simple LSTM LM.

With the change:
[lr=0.001 clips=21 updates=30] loss=8.5153, words per second=3079.8651
[lr=0.001 clips=31 updates=31] loss=6.9493, words per second=3123.0852
[lr=0.001 clips=31 updates=31] loss=6.8555, words per second=3163.4031
[lr=0.001 clips=31 updates=31] loss=6.6663, words per second=3174.6266
[lr=0.001 clips=31 updates=31] loss=6.6453, words per second=3182.0366
[lr=0.001 clips=31 updates=31] loss=6.6166, words per second=3179.5335

Immediately prior to the merge, I get about 5000 words per second.

Unfortunately my desktop computer is suddenly unaccessible from home, so I
can't send you the performance differences on GPU or repro code. But
perhaps we want to roll this back until we investigate?

neubig:
This was concerning so, I confirmed the speed on CPU/GPU on 3 different tasks:
* https://github.com/clab/dynet/blob/master/examples/autobatch/train_rnn-autobatch.cc
* https://github.com/neulab/dynet-benchmark/blob/master/dynet-cpp/treenn-bulk.cc
* https://github.com/neulab/dynet-benchmark/blob/master/dynet-cpp/bilstm-tagger-bulk.cc

And all of them were anywhere from a little faster to a lot faster (10-40%) on both CPU/GPU after the merge. I agree that your regression is concerning and we should figure out the reason ASAP, but given that it's not universal I don't think we need to revert. If you're able to send code that can reproduce I'll be happy to investigate.

redpony:
I've checked in an RNN LM example in examples/rnnlm that shows the contrast
very clearly with the default settings (running on the Mikolov PTB data).

Before yesterday's checkin:
$ python lstmlm-auto.py  --dynet-autobatch 1
[dynet] random seed: 442972807
[dynet] using autobatching
startup time: 0.16513919830322266
[lr=0.001 clips=22 updates=30] loss=8.4710, words per second=4960.6473
[lr=0.001 clips=31 updates=31] loss=6.9619, words per second=4971.5865
[lr=0.001 clips=31 updates=31] loss=6.8444, words per second=4977.6649
[lr=0.001 clips=31 updates=31] loss=6.7307, words per second=4978.4304
[lr=0.001 clips=31 updates=31] loss=6.6279, words per second=4976.5928

With HEAD:master:
$ python lstmlm-auto.py  --dynet-autobatch 1
[dynet] random seed: 326981018
[dynet] using autobatching
RUN WITH AND WITHOUT --dynet-autobatch 1
startup time: 0.17082881927490234
[lr=0.001 clips=21 updates=30] loss=8.5626, words per second=3075.0398
[lr=0.001 clips=31 updates=31] loss=7.0087, words per second=3150.2093
[lr=0.001 clips=31 updates=31] loss=6.8206, words per second=3174.2403
[lr=0.001 clips=31 updates=31] loss=6.7509, words per second=3178.5864
[lr=0.001 clips=31 updates=31] loss=6.6351, words per second=3181.7043

I can confirm that train_rnn-autobatch doesn't have a performance
regression (in fact, it gets slightly faster). I'll leave it up to you if
you think this is a serious problem, but this does underscore the necessity
of having more systematic performance tests.

neubig:
OK, I understand the problem at least. Basically before, there was a bug that caused us not to autobatch together the sum operation, which I fixed resulting in most of the other models to get faster. However, in the particular case of the RNNLM code that you posted, this interacts with our prioritization heuristic, and causes the final matrix multiply + softmax to not be batched across time steps. This is annoying, but not a bug, and the correct way to fix it will be by coming up with a better prioritization method.

And yes, I think having a better suite of tests is a good idea. The DyNet benchmark repository now is automatically testing the speed periodically, so we can probably start by working with that.

Issue #1090, opened 2017-11-23
logistic function inconsistent negative outputs


supakjk:
dynet.logistic is supposed to return values in (0,1). Because of numeric issues, [0,1] should be ok. However, it seems that it sometimes return tiny negative values when the input values are large negative values.

I am using Python 2.7.14, dynet 2.0.1, and Eigen 3.3.4. The following code shows weird output:

```
import dynet as dy
import numpy as np

a = [0.0 for x in xrange(30)]
a += [-17.0502529144, -17.0516262054, -17.0403594971, -17.0335273743, -17.0677127838]
a = dy.inputTensor(a)

b = dy.logistic(a)
c = 0.5 + 0.5*dy.tanh(a*0.5)

print b.value()[-5:]
print c.value()[-5:]
```

Outputs:
```
[-5.960464477539063e-08, -5.960464477539063e-08, 3.976178675202391e-08, 4.0034372261743556e-08, 3.86889098535903e-08]
[-5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08]
```

If I change the 4th line as  `a = [0.0 for x in xrange(100)]`,
The outputs are:
```
[-5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08, 3.86889098535903e-08]
[-5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08, -5.960464477539063e-08]
```

So, logistic results seem not consistent. I thought the implementation of logistic function is indeed `0.5 + 0.5*tanh(input*0.5)`, but not always.

neubig:
Confirmed. The logistic function should always be between 0 and 1, so we should fix this ASAP.

neubig:
Wait, actually in my environment I was not able to confirm this for the logistic function itself. I got the following:

    [5.960464477539063e-08, 2.9802322387695312e-08, 2.9802322387695312e-08, 2.9802322387695312e-08, 2.9802322387695312e-08]
    [5.960464477539063e-08, 2.9802322387695312e-08, -5.960464477539063e-08, 5.960464477539063e-08, 0.0]

The second one is indeed numerically unstable, but naive implementations are. We'll need to look at this more closely.

supakjk:
The 1st line of the output changes as I try different xrange value at `a = [0.0 for x in xrange(30)]`

[...]

PR #1122, opened 2017-12-04
Stable logistic

pmichel31415:
Stabilize logistic function computation (so that there is no negative output)

Addresses #1090

### DIFF
--- a/dynet/functors.h
+++ b/dynet/functors.h
@@ -243,7 +243,13 @@ struct FSoftSignBackward {
 
 struct FLogisticSigmoid {
   DYNET_DEVICE_FUNC inline float operator()(float x) const {
-    return 0.5 + 0.5 * tanh(x * 0.5);
+    if (x >= 0){
+      float exp_minus_x = expf(-x);
+      return 1.f / (1.f + exp_minus_x);
+    else{
+      float exp_x = expf(x);
+      return exp_x / (1.f + exp_x);
+    }
   }
 };
 
--- a/dynet/functors.h
+++ b/dynet/functors.h
@@ -246,7 +246,7 @@ struct FLogisticSigmoid {
     if (x >= 0){
       float exp_minus_x = expf(-x);
       return 1.f / (1.f + exp_minus_x);
-    else{
+    }else{
       float exp_x = expf(x);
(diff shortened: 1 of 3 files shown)


redpony:
This isn’t going to be SIMD friendly- you should check the speed on CPU for
large batches.

redpony:
Actually, better to check the generated assembly to see if the vectorised
ops are getting generated.

neubig:
Taking a look at this now, and indeed this causes a significant speed regression (0m1.565s -> 0m16.911s) using the following code.

    import dynet as dy
    
    for _ in range(10):
      dy.renew_cg()
      x = dy.zeros((10000000,))
      for i in range(20):
        x = dy.logistic(x)
      y = dy.sum_elems(x)
      print(y.value())

@pmichel31415 can we make a SIMD-friendly version by just adding `max(0, val)` to the SIMD functors?

pmichel31415:
Thanks for running those tests! I didn't want to use a rectifier initially because in the use case described in #1090 (and in my own code) the intent is to use `log(sigmoid(x))` which might end up returning NaN.

I'll implement this nonetheless (maybe as `1/(e+e^{-x})` instead), but I wont have time to do it his week unfortunately.

neubig:
We now have a bunch of new examples of functors. Is it worth revisiting this PR?

redpony:
I think the right way to implement this is to do two custom ops, one for
the x<0 case and the other for the x>=0 case, then you the select operation
together with those two. I looked into implementing select and the two
operators together in one op and using the packet operations myself, but
the way Eigen is designed makes this difficult to do (basically, you need
to have access to the vector size of the packet type, which could have been
a template parameter but isn't).

I assume this will be faster than the non-packet version (if if you're
doing both the -x and x cases every time), but this would need to be
empirically looked at.

pmichel31415:
Thanks for the tips, I'll definitely do that. Chris, is there a documentation/reference manual online for simd operations in Eigen?

I'll probably close this PR and open a new one with the simd implementation

redpony:
Unfortunately no, you just have to read the code. Look for a file called
“packet_math.h” or something like that. The functions are called things
like pmul, padd, pdiv, pexp, etc.

Issue #1057, opened 2017-11-09
CUBLAS related error


DevSinghSachan:
My code in dynet is getting the following error:
epoch:0.000673/01       train_loss:9.9675       lr:1.7469281488047272e-07
epoch:0.001345/01       train_loss:9.9751       lr:3.4938562976094545e-07
epoch:0.002018/01       train_loss:9.9434       lr:5.240784162197087e-07
epoch:0.002690/01       train_loss:9.9434       lr:6.987712595218909e-07
epoch:0.003363/01       train_loss:9.9434       lr:8.734640459806542e-07
epoch:0.004035/01       train_loss:9.9434       lr:1.0481568324394175e-06
epoch:0.004708/01       train_loss:9.9383       lr:1.2228496188981808e-06
CUBLAS failure in cublasSgemm(dev.cublas_handle, CUBLAS_OP_N, CUBLAS_OP_N, y.d.rows(), y.d.cols(), l.d.cols(), dev.kSCALAR_ONE, l.batch_ptr(b), l.d.rows(), r.batch_ptr(b), r.d.rows(), acc_scalar, y.batch_ptr(b), y.d.rows())
13
terminate called after throwing an instance of 'dynet::cuda_exception'
  what():  cublasSgemm(dev.cublas_handle, CUBLAS_OP_N, CUBLAS_OP_N, y.d.rows(), y.d.cols(), l.d.cols(), dev.kSCALAR_ONE, l.batch_ptr(b), l.d.rows(), r.batch_ptr(b), r.d.rows(), acc_scalar, y.batch_ptr(b), y.d.rows())

This happens on updating dynet to the latest version.
On one of my older version of the installed dynet, the error is not coming.

Thanks,

chunyang-wen:
would please provide the information from
```shell
nvidia-smi
```

DevSinghSachan:
Hi Chunyang,

Here is the output of `nvidia-smi` of both the systems I tried running the code on:

1 .)
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 367.44                 Driver Version: 367.44                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
[...]

PR #1127, opened 2017-12-08
Fixed pernicious bug in autobatching

neubig:
OK, it took me literally all day to find a bug that can be fixed in 10 characters of code.

Basically, parallel memcpy (used in autobatching) requires memory to be transferred to the GPU, but this was being done in an asynchronous fashion, and as a result the memory wasn't necessarily arriving at the GPU before the memcpy kernel started executing.

This probably fixes https://github.com/clab/dynet/issues/1057

### DIFF
--- a/dynet/exec.cc
+++ b/dynet/exec.cc
@@ -322,10 +322,10 @@ void BatchedExecutionEngine::combine_tensors(
     float** srcs = static_cast<float**>(basemem);
     float** trgs = static_cast<float**>(basemem) + TRG;
     std::size_t* lens = static_cast<std::size_t*>(basemem) + LEN;
-    CUDA_CHECK(cudaMemcpyAsync(basemem,
-                               &(locs)[0],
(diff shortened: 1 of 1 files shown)


redpony:
Does this actually affect things? If these are executed on the same stream,
the kernel shouldn’t start executing until after the copy completes. Is the
story that this forced synch is expensive?

neubig:
I'm not actually sure whether the copy and the kernel are being executed on the same stream (probably not). Probably the best thing to do is enforce that the copy uses the same stream as the kernel, but I wasn't immediately sure how to do this.

And yes, I'm aware that the forced sync might be expensive, I haven't merged because I haven't had time to figure out the speed ramifications yet.

neubig:
OK, so CudaMemcpy caused a performance degradation, but specifying a stream for CudaMemcpyAsync seems to be working for now. I'm not super-confident in this fix though, so after merging I'll continue to monitor for reported problems.

PR #1129, opened 2017-12-09
Added constrained softmax operator.

andre-martins:
Implementation of a ConstrainedSoftmax node as described in this paper:

André F. T. Martins and Julia Kreutzer.
"Learning What’s Easy: Fully Differentiable Neural Easy-First Taggers."
Conference on Empirical Methods in Natural Language Processing (EMNLP'17), Copenhagen, Denmark, September 2017.
https://aclanthology.info/pdf/D/D17/D17-1036.pdf

The function implemented in this node is similar to Softmax, but it allows upper bound constraints on the resulting probabilities. It's differentiable both with respect to both the scores and the constraints. Currently CPU-only.


neubig:
Thanks, looks good!

PR #1131, opened 2017-12-09
Cuda malloc msg

jeticg:
I took the liberty of adding some additional information in the event of `cudaErrorMemoryAllocation`  failure. I personally find it a bit more helpful.


neubig:
Thanks! Could you also add the device number so we can tell which GPU ran out?

jeticg:
Done, now in Python, it would look something like this:
ohne titel

neubig:
Awesome, thank you!

PR #1133, opened 2017-12-14
TreeLSTM additions

akoehn:
- make Nary- and Unidirectional treelstm non-abstract
 - add nodes in arbitrary order if needed
 - swig bindings
 - documentation (fairly minimal)


neubig:
Thanks!

PR #1135, opened 2017-12-14
scala: Add implicit conversions from seq to vector

akoehn:
To non-scala people: These definition allow to pass e.g. a Seq[Int] to a method
expecting an IntVector and and the compiler will automatically call the conversion
method constructing an IntVector from the sequence.
 - pro: less Code to write, easier interaction with scala-native data structures
 - con: conversion cost is hidden istead of explicit

This is a matter of style and maybe others don't like implicit conversions?
@shuheik and @joelgrus might have an opinion.

For me it is quite handy.  Alternatively, I would add it to a separate package
(e.g. edu.cmu.dynet.implicitConversions) which can be imported if needed.

### DIFF
--- a/contrib/swig/src/main/scala/edu/cmu/dynet/RNNBuilder.scala
+++ b/contrib/swig/src/main/scala/edu/cmu/dynet/RNNBuilder.scala
@@ -23,7 +23,7 @@ abstract class RnnBuilder(private[dynet] val _builder: internal.RNNBuilder) {
     new Expression(expr)
   }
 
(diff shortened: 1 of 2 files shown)


joelgrus:
Seems reasonable to me, it looks like I was using helpers like this in the unit tests already 😀

https://github.com/clab/dynet/blob/master/contrib/swig/src/test/scala/edu/cmu/dynet/ExpressionSpec.scala#L11

shuheik:
Looks good to me, too.

PR #1136, opened 2017-12-14
Trivial fix to make doxygen url work

akoehn:


akoehn:
I opted to create separate pull requests for different parts of the code even though some are quite small.  Is that okay?

pmichel31415:
Sure, better safe than sorry. Thanks for the fix

Issue #1137, opened 2017-12-15
Displaying memory pool information


jeticg:
As I understand dynet maintains 4 memory pools for forward pass, backward pass, parameters, and scratch. But sometimes individual models may vary in construction, and thus making memory demand for each pool potentially different. As models grow big, since the GPU memory is relatively limited, it would be nice to know when dynet is "out of memory", which pool exactly is using more memory than the others, so that one can adjust the size for each memory pool manually.

xunzhang:
I'll add this.

jeticg:
Thanks! But I tried the above pull request, it does not seem to address the issue for Python.

xunzhang:
You are right. The GPU malloc issue is caught by `CUDA_CHECK` already, I need to add the profiling function there too. See my coming pull request, please.

For the CPU memory allocation, the `__mm_malloc` will not return an empty pointer since Linux uses the optimistic memory allocation strategy, it will break when DyNet uses the actual out-of-memory address (e.g. `memset`). But I think it's a different issue though.

jeticg:
Resolved by #1141. Closing.

PR #1139, opened 2017-12-15
Add pool memory info for profiling.

xunzhang:
Address https://github.com/clab/dynet/issues/1137.


neubig:
Thanks, this is nice. But it actually doesn't address #1137, which would like the "out of memory" error to note which pool was out of memory. I think this is easy to add, so could you add this as well? Thanks!

xunzhang:
Sure.

PR #1140, opened 2017-12-15
Moved complex structures page from dynet.io to documentation

aterenin:
This is a PR associated with the latest update to the DyNet website. It moves the "Complex Structures" page from `dynet.io` to the documentation.

### DIFF
--- a/doc/source/complex.rst
+++ b/doc/source/complex.rst
@@ -1,0 +1,46 @@
+Complex Structures
+==================
+
+DyNet works for your complex neural networks
+--------------------------------------------
+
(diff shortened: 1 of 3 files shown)


doc/source/complex.rst
```diff
@@ -0,0 +1,46 @@
+Complex Structures
```
neubig: Maybe rename "Projects Using DyNet" and rename the file appropriately? The wording here seems a bit strange at the end of the documentation where the readers would presumably already know about DyNet's strong points. Other than that looks good.

PR #1146, opened 2017-12-21
Run Travis CI on release tags

danielhers:
Release 2.0.2 didn't trigger a deployment to PyPI because it didn't trigger a build at all, due to restriction to run builds only for `master`. Now they'll run for tags like "2.0.2" too.

For this one, I manually triggered a build here: https://travis-ci.org/danielhers/dynet/builds/319905996


neubig:
Thanks!

PR #1149, opened 2017-12-22
Fix cudnn ops after Tensor refactor in PR #1091.

mattr1:
Updating the cudnn-ops to match the newly refactored Tensor class (refactored in PR #1091).


neubig:
Thank you!

Issue #1147, opened 2017-12-21
Swapping out `--dynet-gpu` for `--dynet-gpus 1` automatically is confusing


neubig:
In the code below, the `--dynet-gpu` argument is replaced with `--dynet-gpus 1`.
https://github.com/clab/dynet/blob/master/python/_dynet.pyx#L100

This is confusing if you have a downstream python library that uses `argparse`, because suddenly the `--dynet-gpus` argument appears out of nowhere. It'd be nice if we could just keep the `--dynet-gpu` argument and deal with it appropriately (in C++ maybe?). CC @xunzhang

xunzhang:
I think it should be implemented in python. I'll do a quick enhancement for this.

PR #1152, opened 2017-12-22
Enhanced implmentation of --dynet-gpu option in python end

xunzhang:
Address https://github.com/clab/dynet/issues/1147

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -97,7 +97,7 @@ cdef class DynetParams: # {{{
             shared_parameters([type]): [description] (default: None)
         """
         cpu_use = False
-        sys_argv = sys.argv
+        sys_argv = list(sys.argv)
         if '--dynet-gpu' in sys_argv:
             sys_argv.remove('--dynet-gpu')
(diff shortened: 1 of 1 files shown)


xunzhang:
cc @neubig

python/_dynet.pyx
```diff
@@ -97,20 +97,21 @@ cdef class DynetParams: # {{{
...
         """
         cpu_use = False
-        if '--dynet-gpu' in sys.argv:
-            sys.argv.remove('--dynet-gpu')
-            sys.argv.append('--dynet-gpus')
-            sys.argv.append('1')
-        elif not ('--dynet-gpus' in sys.argv or
-                  '--dynet-devices' in sys.argv or
-                  '--dynet-viz' in sys.argv):
+        sys_argv = sys.argv
```
neubig: Just out of curiosity, did you test this? I thought you would have to do `sys_argv = list(sys.argv)` to make a copy of the list, but I might be wrong.

xunzhang: ohm you are right!

PR #1154, opened 2017-12-22
cuRAND support

neubig:
With help from @aterenin, I was able to get cuRAND support working. Things are looking good! Generating large numbers of random numbers on GPU is easily 60 times faster, and the results look correct.

This PR:
* Supersedes https://github.com/clab/dynet/pull/1094
* Fixes https://github.com/clab/dynet/issues/542 https://github.com/clab/dynet/issues/438
* Mostly fixes https://github.com/clab/dynet/issues/1059


aterenin:
Looks good. Only remaining comment is that you should call curandDestroyGenerator in an appropriate destructor.

PR #1155, opened 2017-12-23
sdist fix

danielhers:
The version needs to be fixed in `setup.py` before creating the sdist so it can be uploaded to the right place in PyPI. Also, create the sdist in a separate Travis job rather than in each of the jobs.


neubig:
Thanks!

PR #1158, opened 2017-12-23
customable flip gradient option: C++ and python support

yanaiela:
regarding #318 and #321 
I added to the ```flip_gradient``` function the lambda parameter from Ganin & Lempitsky article - 
makes the reversed task objective customizable


pmichel31415:
I think this is good. Maybe we should have a separate "weight gradient" expression too

neubig:
Actually, I think @pmichel31415 's suggestion is a better one. Why don't we remove this from `flip_gradient`, and just add a `scale_gradient` operation that allows you to scale the gradient by an arbitrary scalar. Then you can call `scale_gradient(x, -lambda)` to get the same effect.

yoavg:
How about renaming the current flip_gradient to scale_gradient, and have flip_gradient be implemented as Expression flip_gradient(x) { return scale_gradient(x, -1); }  ?

neubig:
@yoavg Yeah, sounds good.

PR #1159, opened 2017-12-25
Implementing dropout for SimpleRNNBuilder

tohtsky:
Though not practical, I needed dropout functionality for SimpleRNN.
Unfortunately, with relatively large dropout ratio and long sequence, this lead to the explosion of gradient (due to the scaling factor in dropout?).

### DIFF
--- a/dynet/rnn.h
+++ b/dynet/rnn.h
@@ -171,14 +171,14 @@ struct RNNBuilder {
    *
    * \param d Dropout rate
    */
-  void set_dropout(float d) { dropout_rate = d; }
+  virtual void set_dropout(float d) { dropout_rate = d; }
 
   /**
    *
    * \brief Disable Dropout
    * \details In general, you should disable dropout at test time
    */
-  void disable_dropout() { dropout_rate = 0; }
+  virtual void disable_dropout() { dropout_rate = 0; }
 
   /**
    *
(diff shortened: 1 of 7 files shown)


dynet/rnn.h
```diff
@@ -166,10 +166,19 @@ struct RNNBuilder {
   }
 
   /**
-   *
-   * \brief Set Dropout
-   *
    * \param d Dropout rate
+   * \details The dropout implemented here is the variational dropout with tied weights introduced in [Gal, 2016](http://papers.nips.cc/paper/6241-a-theoretically-grounded-application-of-dropout-in-recurrent-neural-networks)
```
pmichel31415: I'm assuming you mean tied masks and not tied weights here?

tohtsky: You are right, there is no tied / un-tied variants for SimpleRNN, as there's single gate in SimpleRNN.

tohtsky:
Thanks for the comment.
So I've made several changes : 

1. Moved the document for set_dropout from that in base RNNBuilder to SimpleRNNBuilder (just a mistake, sorry!)
1. Enabled different dropout ratios for internal connections and input weights.
1. Corrected some documents.
1. Added Cython declarations and set_dropout & set_dropout_masks methods

pmichel31415:
Sorry for taking so long to get back to you. Do you think you could fix #1174 in this PR as well?

tohtsky:
Sure, no problem, I've added virtual specifiers for set_dropout(d) and disable_dropout() functions.
I believe this will automatically correct ``set_dropout`` in the python API.

By the way, it seems the dropout strategy for ``GRUBuilder`` is still that of Zaremba et. al.
Can I also work on changing its dropout strategy to variational dropout in this PR?

pmichel31415:
This PR is already pretty big so I'll go ahead and merge it, can you submit the GRU changes separately?

PR #1176, opened 2018-01-10
Added `Trainer.set_learning_rate()` #695

yuvalpinter:
Allow Trainer update when using cython versions < 0.24
@yoavg


pmichel31415:
Thanks! The tests seem to be failing because bitbucket was down so I think this can be merged safely

PR #1180, opened 2018-01-11
Save/load with generators in Python API

danielhers:
This allows saving memory or even showing a progress bar while loading the parameters (using `tqdm`, for example). This helps when waiting for a model to be saved/loaded, as it gives an idea for how far along it has got.
`load` has the same behavior as before. Added `load_generator`, which yields parameters rather than returning a list.
`save` didn't need to change at all, since it already supported generators. I only generalized the documentation.
Also used `with` syntactic sugar for opening files, and added informative messages to assertions.

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -649,7 +649,8 @@ cdef class Parameters: # {{{
             arr = np.asarray(arr)
         shape = arr.shape
         if self.shape() != shape:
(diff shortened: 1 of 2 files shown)


danielhers:
Build failing due to df20bc09cdbfacc577e3378f1d863241b20230c6.

PR #1186, opened 2018-01-16
Transformer example in C++

duyvuleo:
Hi,

I would like to add a C++ implementation example for recent Transformer architecture (Attention is All You Need). It would be good if DyNet has it. In fact, we can build a complete sequence-to-sequence system using transformer architecture with this simplified code. 

I also maintain another repository (https://github.com/duyvuleo/Transformer-DyNet) which has how-to-use examples and results on some datasets. 

Hope you accept it. Thanks!

--
Cheers,
Vu


neubig:
Thanks! This will be nice to have.

PR #1191, opened 2018-01-17
Fixed errors on mac by removing "auto" calls

neubig:
Added "auto" calls were breaking at runtime on Mac. This PR removes these.


neubig:
Sorry, apparently I misread the test. This was not just mac, but linux as well (which makes more sense).

neubig:
I've confirmed this works on mac, but Travis is taking forever, so I'll merge without tests passing.

PR #1194, opened 2018-01-19
Faster inputTensor()

### LINKED ISSUE #1143: non-slow version of inputTensor
Currently the `inputTensor` function is slow because it swaps the memory layout from numpy's row-major order to DyNet's column-major order. We are thinking of adding `inputTensorTranspose` that inputs a tensor and transposes its dimensions at the same time. This will allow for inputting numpy tensors without swapping the dimensions, which will probably be important when inputting data in this way.

msperber:
This is an initial attempt to fix #1143, following discussion with @neubig and @yoavg . By bypassing a few intermediate calls, I was able to reduce the time needed to input a 100000000-sized vector as follows:

CPU:
125.5sec -> 12.3sec
GPU:
13.5sec -> 6.5sec

The Expression returned by the new version no longer supports set(), so it cannot be used with static graphs. The question here would be how we handle this API-wise. As of this current implementation, we distinguish between the two cases via ```dy.inputExpr(np_arr)``` for dynamic graphs, and ```dy.inputExpr(np_arr,reusable_expr=True)``` for static graphs.

To-Do list:

- [ ] establish API
- [x] check if further speed-ups are possible
- [x] if needed, implement inputTensorTranspose()
- [x] practice test

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -1995,6 +1995,22 @@ cdef class _inputExpression(Expression):
 def scalarInput(float s, device=""):
     return _cg.inputValue(s, device)
 
+cdef class _immutableVecInputExpression(Expression):
+    cdef vector[float] val
+    def __cinit__(self, ComputationGraph g, vector[float] val, dim=None,batch_size=1,device=""):
+        self.val = val
+        if dim is None: dim = self.val.size()
+        self.cg_version = g.version()
+        cdef CExpression e
+        cdef CDevice* dev
+        if str(device) != "":
+            dev = c_str2dev(device)
+            e = c_input(self.cgp()[0], Dim(dim,batch_size=batch_size), &self.val, dev)
+        else:
+            e = c_input(self.cgp()[0], Dim(dim,batch_size=batch_size), &self.val)
+        self.vindex = e.i
+        g._inputs.append(self)
+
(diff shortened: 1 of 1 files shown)


msperber:
This would be ready for a review now @neubig @yoavg . I tested it with a model that takes speech inputs and get better speed and reasonable outputs. I also tried the static-graph example from the documentation to make sure I didn't break anything there. I tried to keep API changes at a minimum, but this PR will break existing code that uses static graphs and inputTensor (in order to unbreak, people will have to add ```reusable_expr=True``` as argument). Alternatively, we could make True the default, but that will limit the speed-up experienced by users unless they add ```reusable_expr=False``` in their code.

Changes are summarized as follows:
- avoid some unnecessary intermediate calls
- for flat inputs, skip call to np.flatten()
- for non-reusable inputs, directly use float-vectors rather than the rather slow FloatVectorValue container
- best speed is achieved by the new inputTensorTranspose() operator that takes in a flat numpy array and desired dimensions and will directly pass it through in column-major format.
- deactivate bound checks

As before, we make sure things are not garbage-collected prematurely by referencing the stored values in the returned Expression which is itself held in the computation graph. I have a feeling that more speed-up might be possible, but this is about as far as I got without spending too much time digging into Cython memory management.

neubig:
Thanks, LGTM!

Issue #1199, opened 2018-01-22
Implementation of a straight-through operator ?


FilippoC:
Hello,

is there an easy way to implement a straight-through operator in dynet?
i.e.:
- Forward: argmax
- Backward: softmax loss

Thanks.

pmichel31415:
FYI this is not doable with the current set of operations but I'm working on it

FilippoC:
@pmichel31415  well, the easy way is to manually compute the argmax (i.e. outside dynet), create a new input with the argmax and then add (softmax - block_gradient(softmax)), but its kinda annoying.

neubig:
Fixed by #1208

PR #1208, opened 2018-01-26
Argmax and straight-through estimator

pmichel31415:
Addresses #1199 

Only the argmax is done for now. Not sure this is very optimized (I'm basically looping over the batch dimension), I couldn't find a way to do complicated indexing in Eigen. Suggestions are welcome.

TODO:

- [x] Straight-through estimator
- [x] Python bindings


pmichel31415:
Should be ready to merge once the tests are finished

neubig:
Is this finished? It looks like we're still missing some documentation for the argument `d`. I'll take a look when it's ready.

pmichel31415:
OK I changed the API according to our email discussion.

I removed the "d" argument because it's not supported yet. I left it in the node because we might need it if we ever want to implement argmax along arbitrary dimensions.

PR #1215, opened 2018-02-01
MSVC fixes

mattr1:
With these, the Appveyor unit tests work again. Things that break the CUDA compilation don't break Appveyor because it's only testing the CPU compilation. The Transformer project needed a few small tweaks to get it to compile. The curand change is to fix the python compile (not sure why it didn't fail in appveyor).


neubig:
Thanks!

PR #1216, opened 2018-02-03
autobatch for conv2d

draplater:
The current version lacks support for conv2d autobatch, so I try to add it. I just reference the autobatch implementation of other nodes and don't sure what I modification is correct.

### DIFF
--- a/dynet/nodes-conv2d.cc
+++ b/dynet/nodes-conv2d.cc
@@ -72,6 +72,8 @@ Dim Conv2D::dim_forward(const vector<Dim>& xs) const {
 
 int Conv2D::autobatch_sig(const ComputationGraph & cg, SigMap &sm) const {
   Sig s(nt::conv2d);
+  // Note that autobatching will only occur when inputs are of batch size one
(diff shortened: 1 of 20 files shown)


dynet/nodes-conv2d.cc
```diff
@@ -70,6 +70,26 @@ Dim Conv2D::dim_forward(const vector<Dim>& xs) const {
   return Dim(output_shape, bs);
 }
 
+int Conv2D::autobatch_sig(const ComputationGraph & cg, SigMap &sm) const {
+  Sig s(nt::conv2d);
+  if(dim.bd == 1) {
```
neubig: Note that this will not batch convolutions where the filter has batch dimension 1, but the input has batch dimension more than one. This seems like a case that we ideally should support, so could you either support this, or add a TODO here noting that this isn't supported yet, but should be?

neubig:
I made the changes myself, and also made a tangential change of allowing CuDNN to be found when it exists in the cuda toolkit directory without specifically specifying CUDNN_ROOT.

This should be ready to merge pending integration tests passing.

Issue #1217, opened 2018-02-04
Equivalent of torch.nn.functional.logsigmoid in dynet


Aditi138:
Is there an equivalent function for torch.nn.functional.logsigmoid in dynet?
I tried the operation with dy.log(dy.sigmoid()) but that leads to nan's

duyvuleo:
I think there is no log_sigmoid at the moment but you may do the equivalent
thing to avoid nan:

dy.log(dy.sigmoid(x) + sigma) where sigma = 10^-6 or smaller.

Not sure this is a right way.

Sent from Note 3

On Feb 4, 2018 12:17 PM, "Aditi" <notifications@github.com> wrote:

Is there an equivalent function for torch.nn.functional.logsigmoid in dynet?
I tried the operation with dy.log(dy.sigmoid()) but that leads to nan's

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/clab/dynet/issues/1217>, or mute the thread

Aditi138:
Thanks! This helped.

neubig:
This would be worth having as an actual function, so I'll leave this open.

neubig:
Fixed by #1224

PR #1224, opened 2018-02-07
Log sigmoid

pmichel31415:
Addresses #1217

### DIFF
--- a/doc/source/python_ref.rst
+++ b/doc/source/python_ref.rst
@@ -189,6 +189,8 @@ Arithmetic operations
 
 .. autofunction:: dynet.log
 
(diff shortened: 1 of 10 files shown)


neubig:
`simd-functors` was using 64-bit precision functions, which was tripping up CUDA. I fixed at `log1p` and `exp` to use their 32-bit versions, which seems to have fixed compile errors (and also reduced warnings?)

pmichel31415:
Thanks!

supakjk:
I'd like to know how return (1 - expf(t)) * d; is derived in scalar_log_sigmoid_backward_op. Could you elaborate it? It seems not stable when t is very large.

PR #1227, opened 2018-02-07
Fix rnn state init

### LINKED ISSUE #1223: buggy python RNN.initial_state(vecs)
This function (as well as the one initializing a state from raw vectors) is buggy:

https://github.com/clab/dynet/blob/850c4c87f966a9f9f96e06f0ad12a2b20bd95ebb/python/_dynet.pyx#L4652

the `vecs != None` case should be outside the cg version check.

yoavg:
This should fix #1223 . I tested it locally but I'll appreciate another set of eyes on this.

(this PR includes also an unrelated change by mistake, but I'm not sure how to remove it, so some git help will be appreciated.)

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -71,6 +71,13 @@ if(PYTHON)
   add_dependencies(pydynet pydynet_precopy)
   add_dependencies(pydynet_precopy dynet)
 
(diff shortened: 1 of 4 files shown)


yoavg:
(I'll create a clean one)

PR #1228, opened 2018-02-07
fix initializing RNN state with given expressions

yoavg:
This should solve #1223 , but I'll appreciate another set of eyes on this.

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -4644,21 +4644,23 @@ cdef class _RNNBuilder: # {{{
         Args:
             vecs (list): Initial hidden state for each layer as a list of :code:`dynet.Expression` s  (default: {None})
             update (bool): trainer updates internal parameters (default: {True})
+                           NOTE: subsequent calls without calling dynet.renew_cg() will not change the `update` behavior.
         
         Returns:
             :code:`dynet.RNNState` used to feed inputs/transduces sequences, etc...
             dynet.RNNState
         """
+        # if we didn't initialize for this CG yet, create a new _init_state
         if self.cg_version != _cg.version():
             self.new_graph(update)
-            if vecs is not None:
-                self.start_new_sequence(vecs)
-            else:
-                self.start_new_sequence()
+            self.start_new_sequence()
             self._init_state = RNNState(self, -1)
+        # if we have vecs, return a new state based on the initial state
+        if vecs is not None:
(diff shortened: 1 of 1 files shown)


neubig:
Looks good! Quick question though, what happens if you do:

    state_a = rnn.initial_state(a)
    state_b = rnn.initial_state(b)
    state_c = state_a.add_input(c)

Is `state_c` the result of calculating "a, c" or "b, c"? If I understand how Python handles references, I feel like it might be "b, c", which would be counter-intuitive. If it's "a, c" then I think this is fine to merge.

yoavg:
Why do you think it will be `b,c`? (which reference do you think is being shared?)

neubig:
It seems that both `state_a` and `state_b` will be pointing to `self._init_state` in the `RNNBuilder`. When you call `set_s()` when generating `state_b`, I think this might overwrite the expressions in `self._init_state`, which would affect `state_a` as well.

yoavg:
`set_s()` adds a new state, it doesn't replace the existing one (also in cpp). 

anyhow, does this convince?
```
In [2]: m = dy.Model()

In [3]: b = dy.LSTMBuilder(1, 2, 2, m)

In [4]: x1 = dy.inputTensor([1,1])

In [5]: x2 = dy.inputTensor([2,2])

In [6]: x3 = dy.inputTensor([3,3])

In [7]: b1 = b.initial_state()

In [8]: b2 = b.initial_state([x1,x1])

In [9]: b2.add_input(x1).output().npvalue()
Out[9]: array([ 0.09968442,  0.11609479])

In [10]: b2.add_input(x1).output().npvalue()
Out[10]: array([ 0.09968442,  0.11609479])

In [11]: b3 = b.initial_state([x2,x2])

In [12]: b3.add_input(x1).output().npvalue()
Out[12]: array([ 0.04428694,  0.31293687])

In [13]: b2.add_input(x1).output().npvalue()
Out[13]: array([ 0.09968442,  0.11609479])

In [14]: b3.add_input(x1).prev() == b2
Out[14]: False

In [15]: b3.add_input(x1).prev() == b3
Out[15]: True

In [16]: b2.add_input(x1).prev() == b3
Out[16]: False

In [17]: b2.add_input(x1).prev() == b2
Out[17]: True
```

neubig:
Thanks! LGTM then.

PR #1229, opened 2018-02-07
add __gitversion__ info to python build

yoavg:
This should allow:
```
import dynet
print dynet.__gitversion__
```
to get the exact version this module is based on.

It also fixes a build problem on my local machine, where automatic files prevented copytree from working.


neubig:
Good idea, thanks!

PR #1233, opened 2018-02-09
[WIP] simplify Parameters in python

yoavg:
This change makes adding parameters to the graph redundant in python.

Instead, each parameter can participate directly in computations, and will be automatically re-added to a graph when it is being used (this is handled in the `.c()` method, and the calls to `._iexpr()` ).

The API should be backward-compatible to before, although using `dy.parameter(x)` is not longer needed.

Would be nice if someone besides me gave it some stress-testing before merging this in.

To be clear, this PR allows:
```
import dynet as dy
m = dy.Model()
x = m.add_parameters((10,10))
y = m.add_parameters((10,10))
# x and y can be used as expressions
z=x+y

dy.renew_cg()
# z is now stale, but x and y can be used as is.
e=x+y
```

beyond the saving in typing overhead, it should also provide some efficiency benefits by removing the python calls to `dy.parameter(x)`, which unfortunately do take non-negligible time..


yoavg:
I can’t figure out why appveyor fails (mostly because I have no idea what it is, and the error seem to be windows related (?)). Advise anyone?

danielhers:
I restarted the build and it passed.

yoavg:
ping?

neubig:
ACL... (Will try to take a look on Friday)

redpony:
Next week. Sorry.

neubig:
LGTM. The only difficulty is if we incorporate something supporting multiple computation graphs, this functionality will potentially be more difficult to maintain because we won't know which graph to add things to. But this is immediately useful, and multiple computation graphs is something that we _might_ want in the future, so I think we can just go ahead with this.

yoavg:
So... how do we proceed from here?

Merge as is? Merge in next version?

It currently has a deprecation message when using the dy.parameters() function, should we keep/change this behavior?

neubig:
I think @redpony was going to check next week. I agree that a deprecation message, but not immediately causing programs to crash, is the right way to go. When looking at the deprecation message I did notice two things though: (1) it might be good to say explicitly that the "dy.parameter()" function is deprecated, (2) also "deprecated" is misspelled.

yoavg:
Will fix the message.

To be clear, the question was not "should we have a message or crash" but rather "should we have a message or not have a message". In any case crashing is not an option, and the dy.parameter function will keep working (as a nop) for the time being.

redpony:
Hi all, sorry for the delay. I looked through the change and I think
everything is fine- making parameters inherit from Expression is a clever
solution.

Definitely fix the spelling in the deprecation warning and perhaps add a
comment saying that subsequent versions will remove the old API.

I ask that we add a python-design.txt to keep a short, high-level list of
these sorts of major divergences.

Very excited about this!

-Chris

redpony:
Also: I would suggest that we put the code in place to force errors now,
even if we just warn. If this is a simple single line change to start
throwing exceptions, then people who rely on automated tests to figure out
whether their code is compliant will have an easier time with the migration.

yoavg:
Updated the deprecation warning. Not sure what caused AppVeyor to fail again.

redpony:
Is this ready to merge now?

neubig:
I think it's OK if Yoav thinks so.

yoavg:
I think its good to go.

redpony:
I've merged this.

PR #1236, opened 2018-02-10
[WIP] Python wheel for any commit

danielhers:
This PR adds a unique identifier to the Python wheel (binary distribution) created for any commit on Travis CI and AppVeyor. The identifier is "git." + the commit hash. When uploaded to a cloud storage, this will allow easily testing any commit or PR without having to build DyNet yourself - just install from the wheel in a few seconds.
The problem is that PyPI (even TestPyPI) doesn't allow uploading distributions with "PEP 440 local versions", that is, with the "git." + commit suffix, as it's only intended for release distributions, so we can't use it to store the builds.
Can someone suggest another cloud storage to store these? I can configure Travis CI and AppVeyor to upload there, but I don't have the resources to maintain the storage.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,6 +12,8 @@ environment:
     TWINE_USERNAME: danielh
     TWINE_PASSWORD:
       secure: cpMn69tLzuhwO7EPhhiVwA==
+    TWINE_PASSWORD_TEST:
+      secure: PgevW/oQiBa2F4u+FciQBA==
   matrix:
     - PYTHON_VERSION: 3.5
     - PYTHON_VERSION: 3.6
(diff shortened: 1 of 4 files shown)


neubig:
I could potentially set up cloud storage if this is something that we really need, but somehow I feel it's not. Is this functionality that anyone really wants? Is it something that would be fixed with more frequent minor version releases (which is something that we can do and I'm hoping to do anyway).

danielhers:
Well, I wanted to try out #1233 and didn't feel like going through the
clone and compile process, so it would have helped then. But maybe it's not
a very common need.

Op vr 23 feb. 2018 16:01 schreef Graham Neubig <notifications@github.com>:

neubig:
Yeah, I think this is another moving part that we'd have to maintain, and probably more work than it's worth, so I'm going to close this. Sorry about not accepting the PR this time, and thanks always for your help maintaining the build/install process!

danielhers:
Fair enough, thanks for the hard work :)

Op vr 23 feb. 2018 18:29 schreef Graham Neubig <notifications@github.com>:

pmichel31415:
Hey sorry for unearthing this thread but @danielhers do you know if there is a way of doing something like this where we only store the latest commit? Where something like 

```bash
pip install dynet-master
```

Would get you a precompiled version of the latest commit.

This would potentially solve the storing issue since we'd only be storing one version and updating it periodically. LEt me know if you think that's possible to do with pypi, I might take a shot at it

danielhers:
Sure. Just replace the project name in `setup.py` (like the version is replaced [here](https://github.com/clab/dynet/blob/master/.travis/fix_version.sh)) to "dynet-master" in the build script, and the version to some incrementing number (as there is no tag). Also, change the [deployment in .travis.yml](https://github.com/clab/dynet/blob/master/.travis.yml#L234) to run on every commit to mastet, not just tags (remove `tags: true`). But make sure to deploy to the `dynet` project only on tagged commits.

PR #1242, opened 2018-02-12
Exposing weight-decay-lambda through parameter collection.

yoavg:
Adding access to the weight-decay lambda value (both setting and getting) from python and cpp.

The second commit also fixes the issue in #1243 .


neubig:
Thanks!

PR #1250, opened 2018-02-16
Fix using CPU with CUDA compiled dynet when no GPU is available

### LINKED ISSUE #1248: Error when running dynet compiled with cuda on cpu
I compiled dynet with CUDA but when I try to use it on CPU (ie `--dynet-gpus 0` and no GPU visible to my program) I get the following error:

```
CURAND failure in curandCreateGenerator(&curandeng, CURAND_RNG_PSEUDO_PHILOX4_32_10)
```

However when a GPU is visible to my program running on CPU works.

This is a problem because I shouldn't need a GPU to run dynet on CPU.

I'm using the latest master

pmichel31415:
I think this fixes #1248 (it works on my machine at least)


neubig:
Thanks!

PR #1270, opened 2018-03-02
Fixing equation mistake in documentation

jkkummerfeld:
The C++ documentation for the L2 norm used the same equation as for the squared norm (in other words, it was missing the square root). This puts it in.


pmichel31415:
Thank you!

Issue #1289, opened 2018-03-09
"Competitive Example" (5): Variational Auto-encoders


neubig:
Part of: https://github.com/clab/dynet/issues/1284

We should have an example of variational auto-encoders. This could be either for:

1. Something like the PyTorch example: https://github.com/pytorch/examples/tree/master/vae
2. An implementation of Bowman et al.: https://arxiv.org/abs/1511.06349

I tend to prefer the latter, but the former would also potentially be OK.

pw2393:
Pull Request #1307: 1. done 2. to start

pw2393:
Hi! Got a question regarding the implementation of Bowman et al.: https://arxiv.org/abs/1511.06349:

I noticed in the [line 106](https://github.com/clab/dynet/blob/master/examples/variational-autoencoder/train_rnnlm-aevb.cc#L106) to [line 108](https://github.com/clab/dynet/blob/master/examples/variational-autoencoder/train_rnnlm-aevb.cc#L106) of `train_rnnlm-aevb.cc`, that the input of decoder RNN is the original `sent`. **HOWEVER**, should it only decodes the new hidden state? In other words, with the start-of-sentence token as initial input, the decoding RNN generates (i.e. reconstructs) the sentence step by step (**where the output of last step is the input of next step**) till the end-of-sentence token, right? 

![screen shot 2018-03-22 at 4 58 29 pm](https://user-images.githubusercontent.com/21228317/37798668-cd5677c2-2df3-11e8-9441-a2942b5d546c.png)

Just wanna make sure if I am not misunderstanding something here.

PR #1307, opened 2018-03-14
vae examples and benchmarks vs pytorch (#1289)

pw2393:
A vae example and benchmark vs pytorch

### DIFF
--- a/examples/variational-autoencoder/basic-image-recon/README.md
+++ b/examples/variational-autoencoder/basic-image-recon/README.md
@@ -1,0 +1,37 @@
+# Basic VAE Example: MNIST Images
+
+This example is a DyNet version of the same example as [a PyTorch one](https://github.com/pytorch/examples/tree/master/vae).
+Benchmarks are attached below.
+
+> This is an improved implementation of the paper Stochastic Gradient VB and the Variational Auto-Encoder by Kingma and Welling. It uses ReLUs and the adam optimizer, instead of sigmoids and adagrad. These changes make the network converge much faster.
+
+### 1. Prepare Data
+
+```sh
+mkdir data
+wget -O - http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz | gunzip > data/train-images-idx3-ubyte
+wget -O - http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz | gunzip > data/train-labels-idx1-ubyte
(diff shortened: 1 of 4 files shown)


pmichel31415:
That's very cool! From what I understand dynet is slower than pytorch on CPU with macOS, do you have any intuition why? maybe related to the Eigen version or something?

xunzhang:
Eigen version is a possibility. Another possibility is compiling options.

neubig:
Thanks, this is good! I've merged this because I think it'd be useful now as-is.

1. It seems that the loss is slightly worse in all reported results. Do you think this is a real difference, i.e. is it consistent across multiple runs? If so, we should probably figure out why. It might be due to parameter initialization.
2. We can open another issue to examine why DyNet is slower on the Mac.

pw2393:
Thanks!

1. Excuse me about the confusing 4-decimal test loss reported. It should have been `97.3±0.X` (of which the randomness comes from the parameter initialization and the prior), for both DyNet and PyTorch after 10 epochs of training. (During multiple runs, one's test loss could be lower than the other one's, while sometimes in reverse. The DyNet final reconstructed-or-sampled images are as good as those of PyTorch. )

2. The future benchmarks could also possibly offer some hints for it (the mac slow-down).

pw2393:
**UPDATE: ABOUT THE MAC OS**

The original CPU benchmarks on MAC OS were based on the current PyPI DyNet package. The environment was built via `pip install dynet`.

However, if I build DyNet manually with 
> https://bitbucket.org/eigen/eigen at `2355b22`
> https://github.com/clab/dynet.git at master d69bdc1

the results now become (the bolded):

| OS | Device | Framework | Speed |
| --- | --- | --- | --- |
| MAC OS 10.13.3 | 2.3 GHz Intel Core i5 | PyTorch | ~ 12.5±0.3 s per epoch |
| MAC OS 10.13.3 | 2.3 GHz Intel Core i5 | DyNet (PyPI) | ~ 19.8±0.3 s per epoch |
| MAC OS 10.13.3 | 2.3 GHz Intel Core i5 | DyNet (**Master**) | ~ **9.9±0.3 s per epoch** | 

On Ubuntu, no matter how I install DyNet (via pypi or github master), the results remain the same.

zhisbug:
@pw2393 How did you install Pytorch? This is probably not a fair claim if you install pytorch directly using their binaries.

pw2393:
screen shot 2018-03-26 at 3 53 09 pm

zhisbug:
Is it possible to compile PyTorch from source and see how it performs on Mac?

Issue #1309, opened 2018-03-15
huber_distance() does not work properly


zhechenyan:
I tried to use huber_distance to calculate the distance between two input vectors but the result is not quite as expected.
```
import dynet as dy

ex = dy.vecInput(1)
ex.set([0])
ey = dy.vecInput(1)
ey.set([5])

e = dy.huber_distance(ex, ey, c=1.345)
print(e.value())
```

The output result is 11.6409749985.

Actually, the right way t calculate, according to the [API: huber distance](http://dynet.readthedocs.io/en/latest/operations.html?highlight=huber#_CPPv3N5dynet14huber_distanceERK10ExpressionRK10Expressionf), Because 5-0=5>c=1.345, the output is 1.345*5-0.5*1.345^2=5.8204875 but somehow, the actually is 2 times the desired output.

In another case:
```
import dynet as dy

ex = dy.vecInput(1)
ex.set([0])
ey = dy.vecInput(1)
ey.set([1])

e = dy.huber_distance(ex, ey, c=1.345)
print(e.value())
```

The output result is 1.0.

Actually the desired output should be 0.5*(1-0)^2 = 0.5.

Also, in the [API: huber distance](http://dynet.readthedocs.io/en/latest/operations.html?highlight=huber#_CPPv3N5dynet14huber_distanceERK10ExpressionRK10Expressionf), the `lr` is somehow confusing when `|y-f(x)|<=c`.

xunzhang:
Yeah, DyNet forgets to multiply this by 0.5. https://github.com/clab/dynet/blob/master/dynet/nodes-similarities.cc#L97. If the distance is only used for optimization, that's fine. But I prefer we fixed that.

PR #1311, opened 2018-03-16
Fix HuberDistance::backward_dev_impl and correct log sigmoid doc.

xunzhang:
Fix backward of https://github.com/clab/dynet/issues/1309.


xunzhang:
cc @zhechenyan

Issue #1263, opened 2018-02-28
TopK for Tensor?


zzsfornlp:
Faster TopK functions for Tensor can be useful for processes like beam search. Correct me if I'm wrong, currently I have not found efficient way to do this directly on GPU in dynet.

It seems that tensorflow and pytorch adopt two different methods for this (but seemingly not with Eigen), which can be good references or resources:
Tensorflow (heap-based):
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/topk_op_gpu.cu.cc
PyTorch (radix-selection):
https://github.com/pytorch/pytorch/blob/master/aten/src/THC/THCTensorTopK.cuh

Will it be ok to adapt them into dynet? (Putting the relevant codes in `third_party` with corresponding comments about lisence and sources.)
Maybe something like:
`pair<IndexTensor, Tensor> TensorTools::topk(const Tensor& v, unsigned dim = 0, unsigned k = 1);`
If ok, I would like to work on this.

neubig:
Yes, it's OK to adopt code form other Apache licensed projects to DyNet. If the license is something other than Apache, we'll need to be careful.

We welcome a pull request on this.

PR #1328, opened 2018-03-27
Adding TensorTools::topk.

zzsfornlp:
This PR corresponds to Issue #1263 and add CPU/GPU topk function (only topk, no sorting) for Tensor, adopting the underlying implementations from Tensorflow(Heap-based) and PyTorch(Selection-based).

* Evaluations with different settings on these two GPU-strategies and CPU-version can be found at [HERE](https://github.com/zzsfornlp/misc/blob/master/topkv2/run.log).
* Currently, the default strategy is a simple one based on the K-value: using Tensorflow-version when K<=16, otherwise using PyTorch-version.
* The CPU-version simply uses std::nth_element and loops over the batches, which can be extended with multi-thread methods.
* Also, notice that the PyTorch implementation includes several GPU assembly code, and I only test it on K40.
* Finally, about the **license**, I include the licenses of both Tensorflow and PyTorch at the beginning of corresponding files (`third_party/topk_*.h`), wonder whether this is ok?

About the API, it is similar to `TensorTools::argmax`, but returns a pair of (Tensor, IndexTensor) instead.
`std::pair<Tensor, IndexTensor> topk(const Tensor& v, unsigned dim = 0, unsigned num = 1);`

Here is an example using topk:

    # test topk
    import dynet as dy
    import numpy as np
    VOCAB_SIZE = 1000
    BATCH_SIZE = 20
    K = 5
    v = [float(np.random.randint(0, 100000)) for i in range(VOCAB_SIZE*BATCH_SIZE)]
    t = dy.inputVector(v)
    t2 = dy.reshape(t, (VOCAB_SIZE,), BATCH_SIZE)
    t2_t = t2.tensor_value()
    
    # return val/idx
    val, idx = t2_t.topk(0, K)
    
    # testing
    calc = idx.as_numpy().T
    for i in range(BATCH_SIZE):
        gold = np.argsort(v[i*VOCAB_SIZE:(i+1)*VOCAB_SIZE])[-K:]
        if(set(gold) != set(calc[i])):
            print("gold", gold, [v[i*VOCAB_SIZE:(i+1)*VOCAB_SIZE][x] for x in gold])
            print("calc", calc[i], val.as_numpy().T[i])
        assert set(gold) == set(calc[i]), "bad value"

### DIFF
--- a/dynet/tensor.cc
+++ b/dynet/tensor.cc
@@ -15,6 +15,7 @@
 #include "dynet/gpu-ops.h"
 #include "dynet/cuda.h"
 #endif
+#include "third_party/topk.h"
 
 using namespace std;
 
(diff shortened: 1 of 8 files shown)


neubig:
Thanks, glad to have this! I think adding two implementations might be overkill though (adding more code means potentially adding more edge/corner cases). I had some trouble parsing the results, but do you think it's necessary to have the two implementations separated? Could you summarize the test results a little bit in a more succinct form? Thanks!

zzsfornlp:
According to my observations, I think an important factor is the K value. It seems that the selection-based method (pytorch) is not quite sensitive to K, and can deal with relatively large K. While in a large output setting (>10k), the heap-based method (tensorflow) is faster on smaller Ks (like K<=10), but becomes slower and possibly gets GPU out-of-shared-memory-errors on larger Ks. 

I think this accords to the totally different nature of the two k-max algorithms and maybe there are no one perfect solution for all situations.

Sorry that the test log is kind-of in a mess, here I extract some of the results as in a typical text-generation-setting:
* on one K-40 GPU, Value-Type=float, Index-Type=long, Batch-size=80, Output-size=30k, Inner-stride=1, displaying calculation time in seconds accumulated for 100 runs

 <table style="width:100%">
  <tr>
    <th>K</th>
    <th>CPU (nth_element)</th>
    <th>GPU-Heap (tensorflow)</th>
    <th>GPU-Selection (pytorch)</th>
  </tr>
  <tr>
    <td>1</td>
    <td>3.89452</td>
    <td><b>0.078358</b></td>
    <td>0.258311</td>
  </tr>
  <tr>
    <td>4</td>
    <td>4.08983</td>
    <td><b>0.190607</b></td>
    <td>0.292881</td>
  </tr>
  <tr>
    <td>16</td>
    <td>3.93389</td>
    <td>0.581771</td>
    <td><b>0.302342</b></td>
  </tr>
  <tr>
    <td>32</td>
    <td>3.93769</td>
    <td>1.27852</td>
    <td><b>0.30257</b></td>
  </tr>
  <tr>
    <td>64</td>
    <td>3.89152</td>
    <td>2.86543</td>
    <td><b>0.308461</b></td>
  </tr>
</table> 

I'm not sure which one is the generally better strategy, because this depends on the settings of output-sizes & Ks & others. Maybe for typical applications, we might not need that large K-value and the Selection-based method can be dropped for simplicity? But according to the testing results, if calculating on a larger K, the Heap-based method performs much worse.

neubig:
OK, thanks! In that case I think it's OK to add both.

supakjk:
It'd be great if there is an option to return a sorted top-k.

neubig:
@supakjk Could you open another issue for this? Thanks!

@zzsfornlp I agree that this would be good, do you think it'd be possible?

Issue #1288, opened 2018-03-09
"Competitive Example" (4): Tree LSTMs


neubig:
Part of: https://github.com/clab/dynet/issues/1284

Currently we have an example of training TreeLSTMs here:
> https://github.com/neulab/dynet-benchmark/blob/master/dynet-py/treenn.py
> https://github.com/neulab/dynet-benchmark/blob/master/dynet-cpp/treenn.cc

This should be verified to be competitive with the reported results in the original paper and moved to the competitive examples directory.

zhiyong1997:
Hi, all. I'd like to deal with this.

zhiyong1997:
I have modified the original treenn code, and spent some time tuning hyper-parameters. My code can be accessed by https://github.com/zhiyong1997/dynet/tree/master/examples/treelstm.
I want to make a throughput comparison with TF fold. However, I find that dropout in dynet is very slow. I'm aware of #1154 , but it seems that I still meet with the problem. Without dropout, one epoch takes ~11 seconds, while with dropout, one epoch takes ~18 seconds. Is it a normal case?

neubig:
Could you send a PR (labeled [WIP]) so we can test this easily and discuss there.

xunzhang:
Resolved via https://github.com/clab/dynet/pull/1330. Thanks for the contribution @zhiyong1997 and @zhisbug !

PR #1330, opened 2018-03-28
Tree-LSTM example

zhisbug:
#1288 
@neubig at your convenience please take a look at the dropout implementation, which seems to be extremely slow.

Reported by @zhiyong1997 : Without dropout, one epoch takes ~11 seconds on GPU, but when one dropout is added (only to the layer before logits calculation), the time becomes ~18 seconds. By contrast, when three different dropouts are added in TF fold at the same time, the time only turns from ~13s to ~15s, which is ~30s for dynet.


neubig:
This is because the dropout node is not being autobatched because it doesn't have an autobatching profile. I'll work on adding one today.

zhisbug:
Ready to be merged. @xunzhang will manually squash&merge in order to keep the authorship of commits.
Thanks @zhiyong1997 for the great contribution.

xunzhang:
Squashed and merged into master https://github.com/clab/dynet/commit/2bbc0b965febd5422e0dbbb68dc96da1e3f8dfdd and https://github.com/clab/dynet/commit/0eddde2f9a47e3a9fd98287d4590b5b1391a479b

xunzhang:
@zhiyong1997 please also fix the Codacy Quality check in the future contribution.

PR #1331, opened 2018-03-28
Swig fixes

akoehn:
swig bindings didn't work due to changes in the header files, esp. dynet::detail was not visible to swig anymore.

Also update one example to not use update_epoch anymore.


neubig:
Thank you!

PR #1338, opened 2018-04-04
fixed time and word per sec reported in validation step

YovaKem:
Previously dev_time was not computed at all, and the time and word per sec values computed during training were repeated in the report of the validation step. This also resulted in inaccuracy wrt to the calculations of train_time and word_per_sec during training after the first time validation is performed, as the dev time was not being subtracted from the train time.


neubig:
Thanks!

Issue #1348, opened 2018-04-17
Expose set_random_seed


talbaumel:
Hi, any chance you can expose `set_random_seed` so I can reset it without `dyparams.init()`?
I believe it will be useful (at least for me).

Thanks for the great library!

pmichel31415:
Done in #1351

PR #1351, opened 2018-04-17
Resetting the random seed after initialization

### LINKED ISSUE #1350: CURAND Not Initialized Properly w/o CUDNN?
It looks like CURAND might not be initialized properly if CUDNN is not present:
https://github.com/clab/dynet/blob/master/dynet/devices.cc#L97-L102 

Maybe this is related to this?
https://github.com/clab/dynet/issues/1274

@xunzhang is this a bug?

pmichel31415:
Introduces `dy.reset_random_seed(seed)` to reset the random seed after initialization #1348 

Coincidentally fixes #1350

### DIFF
--- a/doc/source/python_ref.rst
+++ b/doc/source/python_ref.rst
@@ -19,6 +19,8 @@ Initialization functions
 
 .. autofunction:: dynet.init_from_params
 
+.. autofunction:: dynet.reset_random_seed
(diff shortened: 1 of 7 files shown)


pmichel31415:
Not sure why the appveyor tests are failing...

Issue #1354, opened 2018-04-21
Faster batched matrix multiply


neubig:
Currently, matrix multiply where both the left and right sides are batched is simply done with a for loop:
https://github.com/clab/dynet/blob/master/dynet/matrix-multiply.h#L30

This is probably sub-optimal, especially on GPU. Apparently since CUDA 8.0, there is a batched version of matrix multiply:
https://devblogs.nvidia.com/cublas-strided-batched-matrix-multiply/

I think it should be easy to add/test this and should give significant speed improvements!

neubig:
#1355 adds this for CUDA, but it could also be added for MKL on CPU:
https://software.intel.com/en-us/articles/introducing-batch-gemm-operations

PR #1355, opened 2018-04-21
Added batched matrix multiplies on CUDA

neubig:
Addresses https://github.com/clab/dynet/issues/1354 on GPU, but not on CPU


neubig:
Actually, tests seem to be failing, so I'll mark this as WIP.

neubig:
Should be fixed now, it was a problem with broadcasting.
Surprisingly, this doesn't seem to make a large difference for speed with an attention-based MT model, but it might give larger improvements elsewhere.

Issue #1339, opened 2018-04-06
move python examples to python3


akoehn:
Right now, some examples are written in Python 2, some in 3.  Most scripts are missing a shebang.

I'll update the examples to Python 3 syntax and perform some cleanups unless someone stops me :-)

neubig:
This would be appreciated! It would be even better if you could maintain compatibility with Python 2 (for most cases it shouldn't be hard), but if you have to make a tradeoff I'd say go with 3, as 2 will be deprecated soon.

akoehn:
For the jupyter notebooks it's not that easy to verify the compatibility to Python 2 (without boring manual labor) but the only real incompatibility is the `print` statement afaict.

PR #1359, opened 2018-04-25
Enhance mnist-autobatch: automatic downloads, python3, shebang

akoehn:
--download will download the examples automatically.   No need to open a browser anymore.

Relates to #1339.


neubig:
Thanks!

PR #1362, opened 2018-04-25
One hot input operation

pmichel31415:
Adds a `one_hot` operation to create one hot vectors/batches of one hot vectors.


neubig:
Nice idea, thanks!

PR #1363, opened 2018-04-27
Slicing improvements

msperber:
- In Python, allows using numpy-like syntax to select ranges and strides across multiple axes, e.g. ```my_expr[:,1:4,::2]``` will now work. Under the hood, this uses the strided_select operator.
- as a special case, when using this to select a range of batch elements, and this range starts with the first batch element, the operation is marked as in-placed. Ideally, this should also work when the range does not start at the first batch element, but that would require some adjusting some pointers so I'll leave it for a future PR.

### DIFF
--- a/tests/python/test.py
+++ b/tests/python/test.py
@@ -461,7 +461,7 @@ class TestOperations(unittest.TestCase):
 
 class TestSlicing(unittest.TestCase):
 
-    def test_layer_norm(self):
+    def test_slicing(self):
         dy.renew_cg()
         data = np.random.random((10,10,10))
         self.assertTrue(np.allclose(dy.inputTensor(data)[:1,:2,:3].npvalue(), data[:1,:2,:3]))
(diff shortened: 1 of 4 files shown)


neubig:
Awesome, this will be very useful! Could you add a test for this?

Also maybe check this:
https://app.codacy.com/app/xunzhang/dynet/pullRequest?prid=1590987

msperber:
Sure, I've added some tests.

tests/python/test.py
```diff
@@ -459,6 +459,20 @@ def test_layer_norm(self):
         self.assertTrue(np.allclose(y.npvalue(), y_np_value))
 
 
+class TestSlicing(unittest.TestCase):
+
+    def test_layer_norm(self):
```
neubig: The name of this function is wrong, but other than that this is great! I'm happy to merge once this is fixed.

msperber:
Thanks for catching this, should be fixed now.

PR #1370, opened 2018-05-08
Typo in example code for saving/loading

jkkummerfeld:
The code was reloading the model into the same object, rather than the new one. Also tweaked the language slightly


neubig:
Thanks!

PR #1388, opened 2018-06-04
"Competitive Example" (2.5): Sequence-to-sequence

gpengzhi:
Here is a comparison between Dynet and PyTorch on the seq2seq translator example used in PyTorch tutorial (https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html).


xunzhang:
there are something wrong with this pull request, I'll help fix this tmr.

shuheik:
Just a question. Why _dynet.py and _pytorch.py use different learning rates?

gpengzhi:
When I tried the learning rate the PyTorch example used (0.01), the performance is very bad. I then choose a reasonable good learning rate for the Dynet example. When I selected the learning rate 0.2 in PyTorch example, it achieved very bad performance.

shuheik:
@gpengzhi Thank you for the explanation.

PR #1391, opened 2018-06-05
"Competitive Example" (2.5): Sequence-to-sequence

gpengzhi:
Here is a comparison between Dynet and PyTorch on the seq2seq translator example used in PyTorch tutorial (https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html).


gpengzhi:
I think I find a bug in my code. I will fix it and submit the commit very soon.

neubig:
Thanks! Is this fixed and ready for review?

gpengzhi:
@neubig Yes. The bug is fixed.

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,268 @@
+# Seq2seq Translator Benchmarks
+
+Here is the comparison between Dynet and PyTorch on the [seq2seq translator example](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html).
+
+The data we used is a set of many thousands of English to French translation pairs. Download the data from [here](https://download.pytorch.org/tutorial/data.zip) and extract it to the current directory.
+
+## Usage (Dynet)
+
+The architecture of the dynet model `seq2seq_dynet.py` is the same as that in [PyTorch Example](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html). We here implement the attention mechanism in the model.
```
neubig: dynet -> Dynet

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+The architecture of the dynet model is shown as follows. 
+
+```python
```
neubig: I don't think this should be included in `README.md`, because if we make changes to the actual code they might get out of sync. Let's keep the code in `.py` files.

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+    def initHidden(self):
+        return [dy.zeros(self.hidden_dim)]
+```
+
+Install the GPU version of Dynet according to the instructions on the [official website](http://dynet.readthedocs.io/en/latest/python.html#installing-a-cutting-edge-and-or-gpu-version).
+
+Then, run the training:
+
+<pre>
```
neubig: Instead of using `<pre>` tags, can you just use 4 spaces of indent?

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,268 @@
...
+python seq2seq_pytorch.py
+</pre>
+
+## Performance
+
+We run our codes on a desktop with NVIDIA TITAN X. We here have D stands for Dynet and P stands for PyTorch.
+
+| Time (D) | Iteration (D) | Loss (D) | Time (P) | Iteration (P) | Loss (P)|
+| --- | --- | --- | --- | --- | --- |
+| 0m 28s | 5000 5% | 3.2687 | 1m 30s | 5000 5% | 2.8794 |
```
neubig: The loss of DyNet and PyTorch are quite different, even at the start. Maybe this is due to different initialization of the parameters? Could you try to match the parameter initialization strategies between the two toolkits and see if they come more in line? Because PyTorch is better, you could try to match the PyTorch initialization strategy with DyNet.

examples/sequence-to-sequence/seq2seq_translator/README.md
gpengzhi: Modified.

examples/sequence-to-sequence/seq2seq_translator/README.md
gpengzhi: Deleted the architecture part in README.md.

examples/sequence-to-sequence/seq2seq_translator/README.md
gpengzhi: Modified.

examples/sequence-to-sequence/seq2seq_translator/README.md
gpengzhi: 3.2687 and 2.8794 are actually the loss after 5000 iterations for both Dynet and Pytorch. I listed the loss at the start of the training procedure for both Dynet example and PyTorch example. The initial loss of DyNet is 7.9808, and the initial loss of PyTorch is 7.9615.

pmichel31415:
Hi @gpengzhi , thanks for the PR! Do you think you'll have time to fix the lint errors?

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,151 @@
...
+| 5m 17s | 55000 55% | 1.2299 | 14m 5s | 55000 55% | 0.8312 |
+| 5m 46s | 60000 60% | 1.1067 | 15m 17s | 60000 60% | 0.7879 |
+| 6m 15s | 65000 65% | 1.0442 | 16m 48s | 65000 65% | 0.7188 |
+| 6m 44s | 70000 70% | 0.9789 | 18m 6s | 70000 70% | 0.6532 |
+| 7m 13s | 75000 75% | 0.8694 | 19m 18s | 75000 75% | 0.6273 |
+| 7m 43s | 80000 80% | 0.8219 | 20m 34s | 80000 80% | 0.6021 |
+| 8m 12s | 85000 85% | 0.7621 | 21m 44s | 85000 85% | 0.5210 |
+| 8m 41s | 90000 90% | 0.7453 | 22m 55s | 90000 90% | 0.5054 |
+| 9m 10s | 95000 95% | 0.6795 | 24m 9s | 95000 95% | 0.4417 |
+| 9m 39s | 100000 100% | 0.6442 | 25m 24s | 100000 100% | 0.4297 |
```
pmichel31415: This big difference is still weird to me. Can we make these numbers similar so that dynet is competitive with pytorch?

For instance I've noticed that there are some differences in the optimization procedure (dynet uses learning rate 0.2 vs pytorch has learning rate 0.01 if I'm not mistaken).

Also do you have any intuition on why there is such a big speed difference? Do you think it is due to your setup or the differences in implementation?

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,151 @@
+# Seq2seq Translator Benchmarks
+
+Here is the comparison between Dynet and PyTorch on the [seq2seq translator example](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html).
+
+The data we used is a set of many thousands of English to French translation pairs. Download the data from [here](https://download.pytorch.org/tutorial/data.zip) and extract it to the current directory.
```
pmichel31415: Can you credit the original source of the data like in the pytorch tutorial: https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html#loading-data-files

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,151 @@
...
+| 8m 12s | 85000 85% | 0.7621 | 21m 44s | 85000 85% | 0.5210 |
+| 8m 41s | 90000 90% | 0.7453 | 22m 55s | 90000 90% | 0.5054 |
+| 9m 10s | 95000 95% | 0.6795 | 24m 9s | 95000 95% | 0.4417 |
+| 9m 39s | 100000 100% | 0.6442 | 25m 24s | 100000 100% | 0.4297 |
+
+We then show some evaluation results as follows.
+
+Format:
+
+<pre>
```
pmichel31415: `<pre>` -> ` ``` `

examples/sequence-to-sequence/seq2seq_translator/README.md
```diff
@@ -0,0 +1,151 @@
...
+= i m happy you liked it .
+< i m happy you liked it . <EOS>
+
+> ils peuvent chanter .
+= they re able to sing .
+< they re able to sing . <EOS>
+
+> vous etes tellement belle dans cette robe !
+= you re so beautiful in that dress .
+< you re so beautiful in that dress . <EOS>
```
pmichel31415: Why not have the same examples for pytorch and dynet? For comparison.

examples/sequence-to-sequence/seq2seq_translator/seq2seq_dynet.py
```diff
@@ -0,0 +1,367 @@
...
+import math
+r = random.SystemRandom()
+
+# Data Preparation
+
+SOS_token = 0
+EOS_token = 1
+
+
+class Lang(object):
```
pmichel31415: Can you add a small docstring for each class/method/function? So that it is easier for people who want to reuse the code?

examples/sequence-to-sequence/seq2seq_translator/seq2seq_dynet.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def __init__(self, name):
+        self.name = name
+        self.word2index = {}
+        self.word2count = {}
+        self.index2word = {0: "SOS", 1: "EOS"}
+        self.n_words = 2
+
+    def addSentence(self, sentence):
```
pmichel31415: Can you use snake_case for function/method/variable names (but not class names) to be consistent with the dynet API. You should probably do it for the pytorch example as well.

examples/sequence-to-sequence/seq2seq_translator/seq2seq_pytorch.py
```diff
@@ -0,0 +1,378 @@
...
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+r = random.SystemRandom()
+
+# Data Preparation
+
+SOS_token = 0
+EOS_token = 1
+
+
+class Lang:
```
pmichel31415: Can you put this class and all the other utilities that are not dependent on the framework in a separate file like `utils.py`? To avoid code duplication

examples/sequence-to-sequence/seq2seq_translator/seq2seq_dynet.py
```diff
@@ -0,0 +1,367 @@
...
+
+from __future__ import unicode_literals, print_function, division
+import io
+import unicodedata
+import re
+import random
+import dynet as dy
+import time
+import math
+r = random.SystemRandom()
```
pmichel31415: Can you fix the random seed (for reproducibility). Please also fix the dynet/pytorch random seeds as well

examples/sequence-to-sequence/seq2seq_translator/seq2seq_pytorch.py
```diff
@@ -0,0 +1,378 @@
...
+            decoder_input = topi.squeeze().detach()
+
+        return decoded_words, decoder_attentions[:di + 1]
+
+
+def evaluateRandomly(encoder, decoder, n=10):
+
+    for _ in range(n):
+        pair = r.choice(pairs)
+        print('>', pair[0])
```
pmichel31415: This is a very nitpicky comment but can you use `"` or `'` consistently everywhere for strings? A simple search and replace should fix it. I'm partial to `"`.

PR #1420, opened 2018-07-18
Fix glorot initialization for convolutional kernels

blester125:
Other frameworks calculate `m` and `n` for glorot initialization as `in * (h * w)` and `out * (h * w)`. Dynet just calculates them as the sum of all the dimensions. This causes conv parameters to be initialized in a much larger range than other frame works.

For an example with `out = 128, in = 300, h = 1, w = 5` (common for a 1d conv) the glorot bounds as calculated by pytorch are &plusmn;0.052948795 while the bounds for dynet are &plusmn;0.16627992689609528

This pr calculates glorot for conv parameters in the same way as pytorch.


pmichel31415:
LGTM!

pmichel31415:
Actually on second thoughts could you add documentation (both in c++ and python) to explain this particular behavior in the case of 3d tensors?

PR #1437, opened 2018-08-31
SparseLSTM

### LINKED ISSUE #1434: Sparse Vanilla LSTM
Hi @neubig,

I'm trying to implement a WaveRNN model, for a TTS-Project I'm working on (https://tiberiu44.github.io/TTS-Cube/). In order to make it work in realtime, I've already created a C++ backend, which is compatible with the model format stored by DyNet. This is to avoid overhead during the sampling procedure at runtime. I now need to go for the weight pruning technique described here: https://arxiv.org/pdf/1704.05119.pdf. For what I've read, there is no supporting function from this already in place.

Looking over the DyNet code, it seems pretty straight-forward to do this. I just need to create two mask vectors in the VanillaLSTM class for `p_x2i` and `p_h2i`, which I need to add to the `local_model`. Then I just use component-wise multiplication in the forward pass. Just to be sure, these parameters will be saved automatically in the model, right. I don't need to do anything else. 

I have two questions:
1) How do I modify the python bindings to add a new method for setting the pruning percentage;
2) Do you think this will be useful to others (should I create a branch and a pull request, once I'm finished), or do you consider it adds to much overhead to be worthwhile.

tiberiu44:
Hi,

I added a new `SparseLSTMBuilder`. It basically implements the weight pruning technique described in the WaveRNN paper (https://arxiv.org/pdf/1802.08435.pdf).

It adds a new method called `set_sparisty(float percent)`, which modifies the binary masks that are applied on the internal parameters. 

Closes #1434

### DIFF
--- a/dynet/lstm.h
+++ b/dynet/lstm.h
@@ -258,6 +258,12 @@ struct VanillaLSTMBuilder : public RNNBuilder {
    * \details This is equivalent to `set_dropout(0)` or `set_dropout(0,0,0)`
    *
    */
+   
+   void set_sparsity(float percent);
+   /**
+   * \brief Select the percent of weights that will be masked, based on their magnitude
+   */
+   
   virtual void disable_dropout() override;
   /**
    * \brief Set dropout masks at the beginning of a sequence for a specific batch size
--- a/dynet/lstm.h
+++ b/dynet/lstm.h
@@ -321,18 +321,10 @@ typedef VanillaLSTMBuilder LSTMBuilder;
 /**
  * \ingroup rnnbuilders
  * @brief SparseLSTM allows the creation of a "sparse" LSTM, ie with decoupled input and forget gates and no peephole connections
- * @details This cell runs according to the following dynamics :
+ * @details During training the sparsity of the LSTM has to be increased incrementally. 
+ * Sparsity is controlled using the set_sparsity method. This works by sorting all the weights based on their magnitude and applying mask on the top x-percent weight with the lowest magnitude.
+ * More details on the process can be found in [Narang et al., 2017](https://arxiv.org/pdf/1704.05119.pdf). The rest of the implementation is identical to VanillaLSTM
  *
- * \f$
- * \begin{split}
-    i_t & =\sigma(W_{ix}x_t+W_{ih}h_{t-1}+b_i)\\
-    f_t & = \sigma(W_{fx}x_t+W_{fh}h_{t-1}+b_f+1)\\
-    o_t & = \sigma(W_{ox}x_t+W_{oh}h_{t-1}+b_o)\\
-    \tilde{c_t} & = \tanh(W_{cx}x_t+W_{ch}h_{t-1}+b_c)\\
-    c_t & = c_{t-1}\circ f_t + \tilde{c_t}\circ i_t\\
-    h_t & = \tanh(c_t)\circ o_t\\
-   \end{split}
- * \f$
  */
 struct SparseLSTMBuilder : public RNNBuilder {
   /**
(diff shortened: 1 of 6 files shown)


pmichel31415:
Looks like compilation is failing on travis, can you take a look?

tiberiu44:
Yes, I started preparing a Linux box. Everything seems to work on OSX.

tiberiu44:
Hi @pmichel31415,
It seems that the pull request now passes all checks. Sorry for the delay. It took me some time to get the right settings. On Debian it worked with the latest g++ and also with the oldest installable version (5.0).

dynet/lstm.h
```diff
@@ -317,6 +318,157 @@ struct VanillaLSTMBuilder : public RNNBuilder {
...
+  /**
+   * \brief Set all dropout rates to 0
+   * \details This is equivalent to `set_dropout(0)` or `set_dropout(0,0,0)`
+   *
+   */
+   
+   void set_sparsity(float percent);
+   /**
+   * \brief Select the percent of weights that will be masked, based on their magnitude
+   */
```
pmichel31415: I think the docstrings for the two methods here (set_sparsity and disable_dropout) are interchanged.

dynet/lstm.h
```diff
@@ -317,6 +318,157 @@ struct VanillaLSTMBuilder : public RNNBuilder {
...
+  /**
+   * \brief Set dropout masks at the beginning of a sequence for a specific batch size
+   * \details If this function is not called on batched input, the same mask will be applied across
+   * all batch elements. Use this to apply different masks to each batch element
+   *
+   * \param batch_size Batch size
+   */
+  void set_dropout_masks(unsigned batch_size = 1);
+  /**
+   * \brief Get parameters in VanillaLSTMBuilder
```
pmichel31415: VanillaLSTM -> SparseLSTM

python/_dynet.pyx
```diff
@@ -5401,6 +5400,176 @@ cdef class VanillaLSTMBuilder(_RNNBuilder): # {{{
...
+
+    Args:
+        layers (int): Number of layers
+        input_dim (int): Dimension of the input
+        hidden_dim (int): Dimension of the recurrent units
+        model (dynet.ParameterCollection): ParameterCollection to hold the parameters
+        ln_lstm (bool): Whether to use layer normalization
+        forget_bias (float): value to use as forget gate bias(default 1.0)
+    """
+    cdef CSparseLSTMBuilder* thisvanillaptr
```
pmichel31415: This variable should probably be renamed for readability

tiberiu44:
I think I need some help. I don't understand why the gradient check failed. I just added two variables for mask vectors and did a `cmult` with the internal LSTM parameters. Initially, all the values are set to 1.0f. Also, the masks are constant parameters. 

When the masks are set to 1.0, the operations should be identical to the VanillaLSTM. I checked this manually when I was testing the code. Training the SparseLSTM with sparsity set to 0, generated identical output to that of the VanillaLSTM (in terms of loss and weights).

pmichel31415:
OK so my educated guess is that check_grad doesn't handle const_parameters correctly and expects the gradient of the masks to be non-zero.

I don't think there is a straight-forward solution to this apart from fixing check-grad.cc which is a bit out of the scope of this PR so for now I suggest we remove the c++ test and add a disclaimer somewhere in the documentation that this is experimental/untested

tiberiu44:
Hi @pmichel31415 ,
Is everything ok with this pull request? I think I covered all the requested changes.

pmichel31415:
Oops, yes I'll merge, thanks

PR #1460, opened 2018-09-25
SparseLSTM memory optimization

tiberiu44:
Instead of creating const_parameters for sparse masks in add_input, the current implementation directly applies the masks the CG is renewed and uses the masked weights directly. Also, checking to see if the gradient check still fails, so I modified test-rnn.


pmichel31415:
LGTM! Thanks :)

Issue #1144, opened 2017-12-20
Plain C bindings?


neubig:
It might be nice to have plain-C bindings for DyNet. This is not trivial to do because DyNet uses C++ style quite heavily, but for creating things like interfaces to other languages, this might be useful. See:

https://github.com/chantera/dynet-rs/issues/1

xbainbain:
I am working on that now. I created a temporary repository ([dynet-c](https://github.com/xbainbain/dynet-c)) for that but it's still in early stage. A pull request will be made to the main repository when basic functions are fully implemented.

PR #1462, opened 2018-09-30
Add C-APIs.

chantera:
This PR adds C-APIs to use DyNet from other languages such as Ruby, Go, Rust, etc.
To build C-APIs, you will need to specify "-DENABLE_C=ON" when running cmake.


chantera:
Related to #1144.

pmichel31415:
Wow, this is an awesome contribution!

I have a few general suggestions:

- Can you add a readme to ` contrib/c` with some details on how to build the bindings, how to use them for other languages... Nothing necessarily too detailed but to give an idea of what to do to someone who wants eg. ruby bindings. Also a quick note on how to add new features to the bindings. (example for swig bindings: https://github.com/clab/dynet/blob/master/contrib/swig/README.md)
- Add a section to https://github.com/clab/dynet/blob/master/doc/source/other_languages.rst to link to ` contrib/c` so people are aware of its existence.
- [Optional] tests on travis. You'd need to modify `.travis.yml` to compile with ENABLE_C and run the unit tests. This might require a bit of playing around with the travis environment but at least it will let us know when we break something.

chantera:
Thank you for your comments.
I added README and documentation.
For travis, I added `-DENABLE_C` option to the build command.
I also wrote usage in README.
I will add an xor example if users want it, but for now I think it is not necessary since no one uses the raw C APIs rather than bindings from other languages.

pmichel31415:
Thank you so much!

Issue #1465, opened 2018-10-05
On-device Parameter Operations


alexandres:
I would like to implement Averaged SGD in dynet as done in https://github.com/pytorch/pytorch/blob/master/torch/optim/asgd.py 

The issue is implementing https://github.com/pytorch/pytorch/blob/master/torch/optim/asgd.py#L75:

`state['ax'].add_(p.data.sub(state['ax']).mul(state['mu']))`

`state['ax']` and `p.data` are an on-GPU Tensors, `add_` is an in-place operation.

My implementation in dynet is using  numpy arrays:

`state['ax'] += (p.as_array() - state['ax']) * state['mu']`

`state['ax']` is an np-array and `p` is a Parameter.

This works but is VERY slow. I would like to do this operation on the GPU using as little memory as possible (in-place `add_`). 

If I define `state['ax']` as a Parameter, I think I can do the subtraction and multiplication using the computational graph, but how would I do the in-place addition? 

Thanks!

FilippoC:
I have an implementation of the exponential moving average of parameters, I can try to put it online if you want. It's implemented in the Trainer class so it can be used with any optimizer (SGD, momentum SGD, ...) and its on device. It will just need a python binding.

alexandres:
Yes, that would be great!

PR #1466, opened 2018-10-05
Exponential Moving Average

FilippoC:
Hi,

I am working on an implementation of the Exponential Moving Average of parameters that can be used with any optimizer. It (will) work on CPU&GPU.

A previous pull-request was rejected #385. However, this feature seems to be of interest for users, see #73 and #1465. I implemented everything in the Trainer class, therefore the code is simple and located in a single place. I hope that this can be merged when finished. :-) 

TODO:
- test everything
- as of now, the EMA is computed for every parameters of the model. This introduce extra work for non-learned parameters (for example fixed embeddings). I don't know how to fix that yet, but I will look into that
- An example would be nice, probably a tagger where we can show that the EMA gets (hopefully) better results in the PTB
- update save/populate methods

### DIFF
--- a/dynet/training.cc
+++ b/dynet/training.cc
@@ -388,14 +388,14 @@ void Trainer::populate(std::istream& is, real lr)
         else if(ema->device->type == DeviceType::GPU)
         {
             cudaSetDevice(((Device_GPU*) ema->device)->cuda_device_id);
-            update_rule_rule_dev(*(Device_GPU*) ema->device, ema, p);
+            update_ema_rule_dev(*(Device_GPU*) ema->device, ema, p);
         }
         else
             throw std::runtime_error("Bad device in MyTrainer::update_ema_rule");
     }
 
     extern template void Trainer::swap_params_to_ema_rule_dev<Device_GPU>(const Device_GPU& dev, bool bias_correction, bool save_weights, Tensor* p, Tensor* mem, Tensor* ema);
-    template void Trainer::swap_params_to_ema_rule_dev<Device_CPU>(const Device_GPU& dev, bool bias_correction, bool save_weights, Tensor* p, Tensor* mem, Tensor* ema);
+    template void Trainer::swap_params_to_ema_rule_dev<Device_CPU>(const Device_CPU& dev, bool bias_correction, bool save_weights, Tensor* p, Tensor* mem, Tensor* ema);
     void Trainer::swap_params_to_ema_rule(bool bias_correction, bool save_weights, Tensor* p, Tensor* mem, Tensor* ema)
     {
         if(ema->device->type == DeviceType::CPU)
@@ -411,12 +411,12 @@ void Trainer::populate(std::istream& is, real lr)
 
 
     extern template void Trainer::swap_params_to_weights_rule_dev<Device_GPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
-    template void Trainer::swap_params_to_weights_rule_dev<Device_CPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
-    void Trainer::swap_params_to_weights_rule(const Device_GPU& dev, Tensor* p, Tensor* mem)
+    template void Trainer::swap_params_to_weights_rule_dev<Device_CPU>(const Device_CPU& dev, Tensor* p, Tensor* mem);
+    void Trainer::swap_params_to_weights_rule(Tensor* p, Tensor* mem)
     {
         if(p->device->type == DeviceType::CPU)
-            swap_params_to_weights_rule_dev(*(Device_CPU*)ema->device, p, mem);
-        else if(ema->device->type == DeviceType::GPU)
+            swap_params_to_weights_rule_dev(*(Device_CPU*)p->device, p, mem);
+        else if(p->device->type == DeviceType::GPU)
         {
             cudaSetDevice(((Device_GPU*) p->device)->cuda_device_id);
             swap_params_to_weights_rule_dev(*(Device_GPU*)p->device, p, mem);
@@ -556,6 +556,7 @@ void Trainer::swap_params_to_weights()
         return;
     if (!ema_params_saved)
         DYNET_RUNTIME_ERR("Weights have not been save.")
+    ema_params_swapped = false;
 
     const auto& params = model->parameters_list();
     const auto& lparams = model->lookup_parameters_list();
--- a/dynet/training.cc
+++ b/dynet/training.cc
@@ -196,11 +196,9 @@ void Trainer::update() {
 
   // Allocate if necessary
   if(aux_allocated < params.size()) {
-    allocate_shadow_parameters(*model, ema_p.size(), ema_p);
     aux_allocated = alloc_impl();
   }
   if(aux_allocated_lookup < lparams.size()) {
-    allocate_shadow_lookup_parameters(*model, ema_lp.size(), ema_lp);
     aux_allocated_lookup = alloc_lookup_impl();
   }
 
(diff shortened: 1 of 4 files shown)


neubig:
Thanks! I think this looks like a better solution, as it limits the extra complexity.

alexandres:
I think this is a good solution for EMA, but in my case (#1465) I would like a SMA as per [Polyak 92](http://www.meyn.ece.ufl.edu/archive/spm_files/Courses/ECE555-2011/555media/poljud92.pdf).

Can this be modified to accommodate that as well?

In #1465 I was thinking more of creating in-place operations on `Parameters` which would allow SMA (and EMA), being more flexible than this proposal (I believe). Not an expert on dynet so I could very well be wrong! 

You could then have a `ParameterCollection` just for the averages of `Parameters` of interest. Finally, one would need a way to copy the averages to the main `ParameterCollection` so evaluation can be run.

neubig:
@alexandres yes, in-place operations on parameters would be a really nice thing to have. @pmichel31415 and I discussed this before and it's on our list of things to do. This is a bit of a bigger change though, so I think for the time being having these EMA trainers implemented is nice.

Currently tests are failing though, so we'll need to find the reason.

FilippoC:
@neubig yes, I did not expect the test to run on this first commit. I'll fix everything in the days.

@alexandres Do you have the update rules somewhere ? I am a bit lazy to through the paper of Polyak. Is this the same as the averaged perceptron?

alexandres:
@FilippoC here is the PyTorch implementation: https://github.com/pytorch/pytorch/blob/master/torch/optim/asgd.py

FilippoC:
@alexandres Unfortunately reading code from an API that I don't know without type declaration is a little bit difficult. If you have a textual explanation with proper math, it would be easier. :-) 
Anyway, I will try to check that, but I cannot promise anything.

alexandres:
@FilippoC No worries! The update equation from Polyak is simply:

image

and for a running sum (what we need here)

image

pmichel31415:
Hi @FilippoC CI seems to be failing due to compilation errors for manual cuda install, do you have time to take a look?

FilippoC:
@pmichel31415  Unfortunately I am very busy this week. I will try to finish this next week, I am sorry for the delay.
(one of my concerns at the moment with this pull request is that I was not able to have a setting where it improve results on a simple task!)

dynet/training.cc
```diff
@@ -279,7 +370,213 @@ void Trainer::populate(std::istream& is, real lr)
...
+    extern template void Trainer::update_ema_rule_dev<Device_GPU>(const Device_GPU& dev, Tensor* ema, Tensor* p);
+    template void Trainer::update_ema_rule_dev<Device_CPU>(const Device_CPU& dev, Tensor* ema, Tensor* p);
+    void Trainer::update_ema_rule(Tensor* ema, Tensor* p)
+    {
+        if(ema->device->type == DeviceType::CPU)
+            update_ema_rule_dev(*(Device_CPU*)ema->device, ema, p);
+        else if(ema->device->type == DeviceType::GPU)
+        {
+            cudaSetDevice(((Device_GPU*) ema->device)->cuda_device_id);
+            update_rule_rule_dev(*(Device_GPU*) ema->device, ema, p);
```
alexandres: Shouldn't this be `update_ema_rule_dev`?

dynet/training.cc
```diff
@@ -279,7 +370,213 @@ void Trainer::populate(std::istream& is, real lr)
...
+            swap_params_to_ema_rule_dev(*(Device_GPU*)ema->device, bias_correction, save_weights, p, mem, ema);
+        }
+        else
+            throw std::runtime_error("Bad device in MyTrainer::swap_params_to_ema_rule");
+    }
+
+
+    extern template void Trainer::swap_params_to_weights_rule_dev<Device_GPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
+    template void Trainer::swap_params_to_weights_rule_dev<Device_CPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
+    void Trainer::swap_params_to_weights_rule(const Device_GPU& dev, Tensor* p, Tensor* mem)
```
alexandres: Should be `void Trainer::swap_params_to_weights_rule(Tensor* p, Tensor* mem)`.

dynet/training.cc
```diff
@@ -279,7 +370,213 @@ void Trainer::populate(std::istream& is, real lr)
...
+            cudaSetDevice(((Device_GPU*) ema->device)->cuda_device_id);
+            swap_params_to_ema_rule_dev(*(Device_GPU*)ema->device, bias_correction, save_weights, p, mem, ema);
+        }
+        else
+            throw std::runtime_error("Bad device in MyTrainer::swap_params_to_ema_rule");
+    }
+
+
+    extern template void Trainer::swap_params_to_weights_rule_dev<Device_GPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
+    template void Trainer::swap_params_to_weights_rule_dev<Device_CPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
```
alexandres: Should be `const Device_CPU& dev`.

dynet/training.cc
```diff
@@ -279,7 +370,213 @@ void Trainer::populate(std::istream& is, real lr)
...
+    void Trainer::swap_params_to_weights_rule(const Device_GPU& dev, Tensor* p, Tensor* mem)
+    {
+        if(p->device->type == DeviceType::CPU)
+            swap_params_to_weights_rule_dev(*(Device_CPU*)ema->device, p, mem);
+        else if(ema->device->type == DeviceType::GPU)
```
alexandres: Should be `else if(p->device->type == DeviceType::GPU)`.

dynet/training.cc
```diff
@@ -279,7 +370,213 @@ void Trainer::populate(std::istream& is, real lr)
...
+            throw std::runtime_error("Bad device in MyTrainer::swap_params_to_ema_rule");
+    }
+
+
+    extern template void Trainer::swap_params_to_weights_rule_dev<Device_GPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
+    template void Trainer::swap_params_to_weights_rule_dev<Device_CPU>(const Device_GPU& dev, Tensor* p, Tensor* mem);
+    void Trainer::swap_params_to_weights_rule(const Device_GPU& dev, Tensor* p, Tensor* mem)
+    {
+        if(p->device->type == DeviceType::CPU)
+            swap_params_to_weights_rule_dev(*(Device_CPU*)ema->device, p, mem);
```
alexandres: Should be `swap_params_to_weights_rule_dev(*(Device_CPU*)p->device, p, mem);`

dynet/training.cc
```diff
@@ -279,7 +370,213 @@ void Trainer::populate(std::istream& is, real lr)
...
+        Tensor& ema = ema_lp[i].all_h;
+
+        swap_params_to_ema_rule(bias_correction, save_weights, &weights, &mem, &ema);
+    }
+}
 
+void Trainer::swap_params_to_weights()
+{
+    if (!ema_params_swapped)
+        return;
```
alexandres: @FilippoC Aside from the compilation issues above, I think the reason for why you aren't getting results is the lack of `ema_params_swapped = false;` here to undo the `ema_params_swapped = true;` on line 515. 

I made this fix and the code seems to working perfectly! :) 

Thanks a bunch for this. Using CMA as I wished was absolutely trivial. I simply changed the update rule to `tvec(*ema).device(*dev.edevice) = (real(ema_updates) * tvec(*ema) + tvec(*p)) / (real(ema_updates)+1);`.

FilippoC:
@alexandres Thank you very much! I am currently out of work, hopefully I'll fix this monday when I will be back to work!

neubig:
@FilippoC I see this was changed back to "WIP". Is it ready for a check? Or do things still need to be modified?

FilippoC:
@neubig Sorry I miss your message. I want to add an example where resutls can be easily be compared. I'll try to do it this afternoon.

FilippoC:
@neubig I made some test with the tagger, seems to work. I also updated the tagger example with command line options.

Here is the dev accuracy w.r.t. to the epoch:
moving_avg

neubig:
Thanks! Sorry it took so long to merge.

PR #1472, opened 2018-10-08
Throws exception when `default_device` is null.

chantera:
In the current DyNet, `Device::get_global_device` returns `nullptr` if `dynet::initialize` is never called before.
It would be better if this method throws an exception when `default_device` is null.


pmichel31415:
Thanks!

PR #1483, opened 2018-10-24
Matmul code and speed improvements

pmichel31415:
This builds on #1355 and rewrite the matmul and affinetransform nodes to reduce code duplication (now all the matmul/matmul+transpose logic lives in `matrix-multiply.h` only).

This should also enable cublas batched matmul in the backward pass, which hopefully will make things like dot attention faster.

TODO:

- [x] Confirm speed gain
- [x] figure out why there is an ""assert false" in `matrix-multiply.h` (cf #1355)
- ~[ ] Add matmul + transpose as an operation (add the option to transpose in the matmul node)~ <- maybe for a future PR but this should help further with attention (both memory and speed wise)

### DIFF
--- a/dynet/matrix-multiply.h
+++ b/dynet/matrix-multiply.h
@@ -112,7 +112,6 @@ inline void MatrixMultiplyTranspAcc(const dynet::Device_GPU & dev, const dynet::
           r.v, r.d.rows(),
           dev.kSCALAR_ONE, y.v, y.d.rows()));
   } else {
-    DYNET_ARG_CHECK(false, "MatrixMultiplyTranspAcc");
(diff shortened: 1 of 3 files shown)


FilippoC:
FYI, I just tried this with my implementation of "A Decomposable Attention Model for Natural Language Inference" (I remove the line "DYNET_ARG_CHECK(false, "MatrixMultiplyTranspAcc");"). My implementation uses a static graph and batches of size 64.
- Before: ~188 seconds per epoch
- Now: ~105 seconds per epoch
This awesome because the Pytorch implementation of a colleague of the exact same network takes ~108 seconds per epoch.

Having a matmul + transpose operation would be awesome!

PR #1488, opened 2018-10-27
add a len function to Expressions, returns length of first dim

blester125:
This PR adds a `__len__` function to the python Expression class.

This allows for an expression that has shape `((T, H), B)` to be passed to functions like `BiRNNBuilder.transduce` and it will work the same as a list of length `T` of expression of shape `((H,), B)`.

`__len__` returns the length of the first dimension which is the same behavior of numpy. It is also matches the behavior of `for x in xs: ...`

### DIFF
--- a/python/_dynet.pyx
+++ b/python/_dynet.pyx
@@ -620,6 +620,11 @@ cdef class Expression: #{{{
         return c_dim_as_dim(d)
         # return (d.size(), d.rows(), d.cols(), d.batch_elems())
 
+    def __len__(self):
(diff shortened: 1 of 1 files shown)


redpony:
I can’t approve this change since the length of a (batch) of tensors could
reasonably be defined in many ways.

neubig:
Yes, I agree... The length of the first dimension is a pretty arbitrary choice here, and I'd like to avoid such arbitrary choices and make them explicit.

PR #1511, opened 2019-01-04
add note about dedault gradient clipping behavior

john-hewitt:
The default behavior of clipping gradients to 5 (which seems to hold in Python optimizers; I tried a few) [isn't obvious](https://twitter.com/haldaume3/status/925823753405325316); making this documented here may lead to improved reproducibility.


neubig:
Thanks!

PR #1540, opened 2019-04-28
Initial commit for dynet wrapper for c# - dynetsharp

shaltielshmid:


neubig:
Thanks! I haven't run this myself, but it looks good in general. I'll merge, so let's see if we can get any beta testers to try it out.

PR #1541, opened 2019-05-03
Added function to dump the current values and/or gradients of nodes i…

armatthews:
Added function to dump the current values and/or gradients of nodes in the computation graph either to stdout or to a file.

### DIFF
--- a/dynet/dynet.cc
+++ b/dynet/dynet.cc
@@ -1,4 +1,5 @@
 #include <iomanip>
+#include <fstream>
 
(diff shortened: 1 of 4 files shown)


FilippoC:
It would probably be better if you just take a ref to a ostream object as argument instead of this weird interface with a filename that can be empty.

neubig:
Thanks! I'll merge this now, but I think we can make a modification like @FilippoC suggested later as it becomes necessary. (welcome a PR)

PR #1542, opened 2019-05-05
Dump

armatthews:
Implementation of Tensor::is_valid for CUDA. This allows the new ComputationGraph::dump() function to be used with GPUs.


neubig:
Hmm, this is going to be really slow obviously, but better than nothing, especially because it's mainly for debugging. I think at least adding a comment to the doc noting that this is not optimized for GPU might be a good idea.

Issue #1536, opened 2019-04-08
Overflow error in metadata when saving large lookupParameter


kwalcock:
I have saved a large model with the Windows version of dynet.  The file contains this line:
```
#LookupParameter# /all/_0 {300,2196017} 1950947009 ZERO_GRAD
```
The math doesn't work out.  The large number should be 10540881601.  If this long value (quad word) is converted into an int (double word), the result is the value in the file, 1950947009.  Something overflows here, unexpectedly.

The problem is in io.cc around line 129.  Perhaps the datastream cannot handle the size_t properly.

neubig:
Hmm, this is a bad bug. If someone has access to a Windows machine it'd be great to test/fix it ASAP.

kwalcock:
FWIW we get the same numbers for the Mac version.  Perhaps debugging will be easier there.  The value is redundant and I haven't noticed it causing problems, but maybe I'm not very observant.

#LookupParameter# /all/_4 {300,2196017} 1950947009 ZERO_GRAD

kwalcock:
It does cause problems.  If I write a large lookupParameter to disk, read it back in, and write it to disk again, the two files on disk differ when the calculated sizes become large enough to overflow.  The second file can contain all zeros.  For some sizes, the test program also seems to crash.

kwalcock:
The bug for a lookup parameter is in line 129 in io.cc:

```cpp
size_t strsize = p.all_dim.size() * (FLOAT32_PRECISION + 8) + 1;
```

The problem is that p.all_dim.size() returns an unsigned int. and FLOAT32_PRECISION is also an int.

The calculation will result in an int even though the hoped for value is size_t, which is something like unsigned __int64.

One way to fix it is to cast the first value to the desired size_t as such:

```cpp
size_t strsize = (size_t) p.all_dim.size() * (FLOAT32_PRECISION + 8) + 1;
```

The (non-lookup) parameter has the same problem at line 115 and there are likely more instances elsewhere in the code.

neubig:
Fixed by #1550

PR #1550, opened 2019-05-16
Prevent overflow in calculation of strsize for [lookup]parameters

kwalcock:
The integer calculation otherwise overflows before conversion to a size_t, resulting in bad values being written to serialized models.


kwalcock:
This is meant to address issue #1536.

neubig:
Great, thanks a lot!

PR #1551, opened 2019-05-16
Make sure shared library looks like one

kwalcock:
Windows (10) will not do a System.load on a file that is not named like a loadable file.

### DIFF
--- a/contrib/swig/dynet_swig.i
+++ b/contrib/swig/dynet_swig.i
@@ -11,7 +11,7 @@
 %pragma(java) jniclasscode=%{
     static {
         try {
-            File tempFile = File.createTempFile("dynet", "");
+            File tempFile = File.createTempFile("dynet", ".dll");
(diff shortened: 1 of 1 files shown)


neubig:
Does this work on non-windows machines?

kwalcock:
Yes, neither MacOS nor Linux seem to care what the file is called, at least not the versions we tested, which were the current Mac version and Ubuntu 14, 16, and 18.

kwalcock:
And Ubuntu 19.

neubig:
Great, thanks!

submar:
> Great, thanks!

I used dynet recently and when I use  "model.load()" function, it got error " AttributeError: '_dynet.ParameterCollection' object has no attribute 'load'"   . Can I ask what happened?  
My version is the latest.
Thanks a lot!

kwalcock:
@submar, can you relate your question to this particular commit?  Is there any connection?

PR #1553, opened 2019-05-19
GPU support for C# + added device specification for numeric ecpression functions.

shaltielshmid:


neubig:
It looks like there are compile errors on linux. Could you check? Maybe to add an import of "assert"?

shaltielshmid:
Fixed

neubig:
Great, thanks! LGTM.

PR #1557, opened 2019-05-21
Add Rust bindings

chantera:
This PR adds the Rust bindings to use DyNet in Rust language.


neubig:
Great! I just browsed quickly, but I'm assuming they're functional. Thanks a lot for this contribution.

PR #1563, opened 2019-06-04
Fix sign in description of `log_sigmoid`

bskaggs:
Documentation appears wrong; this fixes the sign in the description of `log_sigmoid`


bskaggs:
Note actual implementation here: https://github.com/clab/dynet/blob/e5a1116adb60d4fa47439b74094e5df83f6dee7c/dynet/simd-functors.h#L147

neubig:
Thanks!

PR #1564, opened 2019-06-06
[typo-fix] In expr.h, tuntime_error -> runtime_error

harumo11:
Hello! developer of dynet.
Thanks for great library.

I found tiny typo in comment of expr.h.
So, I made this pull request.

**before**
Throws a tuntime_error exception if no computation graph is available

**after**
Throws a runtime_error exception if no computation graph is available


neubig:
Thanks!

PR #1565, opened 2019-06-06
Prediction is added to train_xor.cc

harumo11:
Hello, developer of dynet.
Thanks for great library.

Python XOR example has prediction function.
But C++ one doesn't have that function.

So, I added it in order to show how to predict using learned ComputationGraph.

**Concerns**
A file name is `train_xor` although it will have prediction function.
Should I make `predict_xor.cc` file instead of this pull request?

### DIFF
--- a/examples/xor/train_xor.cc
+++ b/examples/xor/train_xor.cc
@@ -65,6 +65,24 @@ int main(int argc, char** argv) {
     cerr << "E = " << loss << endl;
   }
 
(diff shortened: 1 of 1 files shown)


harumo11:
Hello, @neubig 

I could not figure out reasons that the CI test resulted in failure.
Could you take a look the result of CI test.

I know you are super busy. so I'll be waiting for you.
I'm glad if you give me some hints.

Thanks.

neubig:
Thanks!

PR #1576, opened 2019-07-26
Make sure ParameterStorageBase also uses shared pointer for Java interface

kwalcock:
Its subclasses are marked in dynet_swig.i as using shared pointers:

```
%shared_ptr(dynet::ParameterStorage)
%shared_ptr(dynet::LookupParameterStorage)
```
SWIG warns that the base class "is not similarly marked as a smart pointer".  The line below quiets the warning and also prevents crahes (because of double deletes?).
```
%shared_ptr(dynet::ParameterStorageBase)
```
An issue was filed previously and this will soon link to it.


kwalcock:
This addresses issue #1502.

kwalcock:
This is failing because of a python configuration issue unrelated to this PR.  The configuration should be fixed.

PR #1583, opened 2019-08-15
Batched matrix multiplication for CPU

### LINKED ISSUE #1354: Faster batched matrix multiply

artidoro:
Fixes #1354.

This PR adds batched matrix multiplication for CPU using MKL `sgemm_batch` (which is described [here](https://software.intel.com/en-us/mkl-developer-reference-c-cblas-gemm-batch)).

I ran a simple smoke test for batched matrix multiplication both with an without MKL and there was a performance improvement.

Related to #1355 which adds batched matrix multiplication for GPU.

### DIFF
--- a/dynet/matrix-multiply.h
+++ b/dynet/matrix-multiply.h
@@ -54,10 +54,38 @@ inline void MatrixMultiply(const Device_CPU & dev, const Tensor& l, const Tensor
       colbatch_matrix(y).noalias() += mat(l) * colbatch_matrix(r);
 
   } else {
-    // Otherwise, loop over the batches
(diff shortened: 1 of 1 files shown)


artidoro:
I think the build error is unrelated to my changes.

PR #1593, opened 2019-10-14
Add indices checking for parameters in model.cc

Dim131:
Currently, there is no index checking in gradient_l2_norm_dev and sometimes it could access elements beyond the end of the vector (when k1 = params.size() or k2 = lookup_params.size()).

To fix this, we introduce a simple bounds check.
 params.size() > k1
 lookup_params.size() > k2


kwalcock:
I was about to make a PR to fix this exact same problem.  It would have saved me a lot of time had this been merged a year ago.

PR #1614, opened 2020-07-06
Fix serious memory leak in code using CUDNN functionality

cydur:
When running convolutions with CUDNN then I lost about 1GB memory per hour.
I did the following to first get dynet to work with CUDA 10.2 and CUDNN 8.0.1 and then fix the main problem:

- CMakeLists.txt: Added RELWITHDEBINFO build target needed to get some performance even when debugging. Better options for RELEASE (-g removed) and DEBUG (-Og -g).
- cudnn-ops.cu: if CUDNN version is >= 8 then replace removed CUDNN API functions by new ones introduced in V7.
- devices.cc, nodes-conv2d.*, nodes-maxpooling2d.*: Ensure allocated memory is freed on destruction of objects.


neubig:
Many thanks, this is great!

PR #1627, opened 2021-01-17
Free memory for the correct variable

kwalcock:
Otherwise, memory of project_scratch is leaking and gradient_norm_scratch use will cause problems.


neubig:
Thank you!

kwalcock:
Likewise!

PR #1629, opened 2021-01-27
Protect Scala Expression from being deleted via its ExpressionVector

kwalcock:
If an ExpressionVector is garbage collected, the C++ memory for the vector and its expressions will be freed in C++.  If any Scala Expressions that were derived from it with apply(n) are still alive, their C++ counterparts will have disappeared out from under them.  The Scala Expression, when used, will try to access its C++ counterpart and cause the C++ code to crash and take down the JVM with it.


kwalcock:
Someone needs to give Travis a kick.

PR #1648, opened 2022-03-10
Fix compiling on M1 Macs

### LINKED ISSUE #1644: How does it run on an M1 Mac?
I suppose that a C++ build for an Intel Mac can run using Rosetta.  Is anyone able to verify?

Can DyNet be readily built to run natively on an M1 Mac?

In order to be called from Java, which might start out looking like an M1 application, will it work to use the Intel build or would one use the Intel Java in order to not confuse things?

Thanks for the hints.  I don't have a machine yet to try it on.

jonarcherii:
Clang on M1 does not support `-march=native`, so instead let's explicitly use `-mcpu=apple-m1`, this was all that was needed to have a mac python build.

Fixes #1644.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,10 @@ endif()
 ######## Platform-specific options
 if(WIN32)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNOMINMAX")   # Disable min/max macros in windef.h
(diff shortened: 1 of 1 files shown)


neubig:
Thank you!

ai-nikolai:
@neubig thanks for the package support. Is there any update on making dynet available on arm-based Mac?

Thanks.

ai-nikolai:
@JonArcherII could you elaborate on your answer. thank you.

Issue #1582, opened 2019-08-15
Sample cod Error on Gpu: RuntimeError: cublasSgemm(dev.cublas_handle,


yang182:
import numpy as np
import dynet_config
## Declare GPU as the default device type
dynet_config.set_gpu()

import dynet as dy
from optparse import OptionParser
parser = OptionParser()
parser.add_option("--dynet-gpu", action="store_true")
# create a class encapsulating the network
class OurNetwork(object):
    # The init method adds parameters to the parameter collection.
    def __init__(self, pc):
        self.pW = pc.add_parameters((10,30))
        self.pB = pc.add_parameters(10)
        self.lookup = pc.add_lookup_parameters((500,10))

    # the __call__ method applies the network to an input
    def __call__(self, inputs):
        lookup = self.lookup
        emb_vectors = [lookup[i] for i in inputs]
        net_input = dy.concatenate(emb_vectors)
        net_output = dy.softmax((self.pW * net_input) + self.pB)
        return net_output
    def create_network_return_loss(self, inputs, expected_output):
        dy.renew_cg()
        out = self(inputs)
        loss = -dy.log(dy.pick(out, expected_output))
        return loss
    def create_network_return_best(self, inputs):
        dy.renew_cg()
        out = self(inputs)
        return np.argmax(out.npvalue())

dy.init()
dy.renew_cg()
# create parameter collection
m = dy.Model()

## create network
network = OurNetwork(m)

## create trainer
trainer = dy.SimpleSGDTrainer(m)

## train network
for epoch in range(50):
    for inp,lbl in (([1,2,3],1), ([3,2,4],2)):
        loss = network.create_network_return_loss(inp, lbl)
        loss.value()
        loss.backward()
        trainer.update()
        print(loss.value()) # need to run loss.value() for the forward prop

print('Predicted smallest element among {} is {}:'.format([1,2,3], network.create_network_return_best([1,2,3])))

cydur:
I've got a similar  problem using C++ with CUDA V 10.2.
In my case the forward step works fine and I get a loss value which looks OK. But in the backward step there is an Exception telling:
 ** On entry to SGEMM  parameter number 8 had an illegal value
[...]

PR #1666, opened 2023-10-24
autobatching bug fixed.

cydur:
This bugfix addresses the problem I already found several years ago. See my comment to Issue #1582.
I don't think it really was the same problem as the original error described in this issue because my problem was in the backward step and now I think #1582 could have been a problem in the forward step.

In any case this change fixes my problem. And the problem was (autobatching = 1):
- A auto-batched "+" node was the first argument of an also auto-batched "*" node.
In this case the backward step in exec.cc called node->backward() with the batch tensor as last argument instead of the correct sub-tensor of this batch.

### DIFF
--- a/dynet/exec.cc
+++ b/dynet/exec.cc
@@ -566,8 +566,11 @@ const Tensor& BatchedExecutionEngine::incremental_forward_no_update(
           *(active_un_end++) = j;
         }
       }
-      for (size_t j = 0; j < (size_t)sigmap.size(); ++j)
-        prof2avg[j] /= prof2cnt[j];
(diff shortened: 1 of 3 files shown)


cydur:
Well, I don't think my change has anything to do with the above 
"ERROR conda.core.link:_execute(502): An error occurred while installing package 'defaults::charset-normalizer-2.0.4-pyhd3eb1b0_0'.
CondaError: Cannot link a source that does not exist. C:\Miniconda36-x64\Scripts\conda.exe"

Seems to be a problem in the build environment...

PR #1668, opened 2023-11-29
compile on gcc

jmikedupont2:
First version compiling with C++20


cydur:
Hi Mike. There seems to be no reaction on pull requests on clab/dynet. Same problem with my bugfix. 
There is an error in the first line of dynet.cc in your patch. Please have a look.
As soon as you fix this I would be glad to merge your changes into my own branch cydur/dynet.
Thanks, Cydur

